PR #13, opened 2015-07-07
Fix string length in pack_string()

alkor:
Specification says that strings are encoded as sequences of UTF-8
characters preceded by length in bytes. But the length was calculated
as a number of UTF-8 characters. So packing of strings with
non-ASCII characters was broken.

### DIFF
--- a/opcua/uatypes.py
+++ b/opcua/uatypes.py
@@ -155,6 +155,7 @@ def pack_string(string):
         return struct.pack("<i", -1)
     if not isinstance(string, bytes):
         string = string.encode()
+        length = len(string)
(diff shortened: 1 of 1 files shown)


oroulet:
Thanks, I will merge it as soon as I get good internet access

oroulet:
I tried to write a test for this but I am unsure how to use it. Where did you try to write noon ASCII characters?. Can you write example code that works?

alkor:
@oroulet I've created a test as a separate pull request #15

oroulet:
I saw it, thanks. I will merge asap, I am currently traveling

PR #27, opened 2015-08-28
adding timeout in binary client to prevent the client to get stuck

kashp:


opcua/binary_client.py
```diff
@@ -325,7 +326,7 @@ def publish(self, acks=None):
 
     def _call_publish_callback(self, future):
         self.logger.info("call_publish_callback")
-        response = ua.PublishResponse.from_binary(future.result())
```
oroulet: Please remove this one. There is no network operation here so need for it.
then write
git commit --amend
git push

then I will merge the patch. Thank you!

oroulet:
commited and fixed small error

PR #33, opened 2015-09-27
Add Good, Uncertain and Bad status codes

alkor:


oroulet:
Status_code.py is autogenerated! Maybe hardcode those values in generate_statuscode.py?

oroulet:
oups wrong analysis :-) the csv file is downloaded from opcua foundation. so maybe hardcode the values in generate_statuscode.py or add a new csv??? any better idea?

alkor:
Oops, my bad. I've moved them to a separate file.

oroulet:
merged. You commited quite a lot of small good fixes. Can I add you as
maintainer/mener of freeopcua project? It is better to be many as
maintainer, in case people disappear. the rule is simple, do not push
directly, but create pull request. but I push directly to the repositories,
I know I am almost alone to work on.

alkor:
Yes, this would be great! I'm interested in the development of freeopcua 
project. I'm currently making a client that connects to an OPC UA server 
that uses .Net library. So it's likely that I will push a lot of small 
patches in the near future.

28.09.2015 12:56, ORD пишет:

Issue #34, opened 2015-09-30
Calling Methods from UaExpert


destogl:
To be able to debug/test my application base on this stack I would like to call Methods from UaExpert. Usually, UaExpert generates from based on input and output parameters where the values can be entered. With the python-opcua stack this is not the case. In the figure below you can see on the left side multiply method from example server and on the right call of (completely different) method on UA sdk server.
 
Last two days I am search for solution, I tried add names and descriptions (also play with references) but nothing helped. Somebody has idea?

![uaexpert](https://cloud.githubusercontent.com/assets/1918204/10191813/ae9bc296-6777-11e5-82d0-6a904c4764ca.png)

oroulet:
interresting.... I thought it worked...I see the method GetMonitoredItems
in ObjectTypes/BaseObjectTypes/ServerType also has the same issue. I think
there is not other way than starting wireshark and comparing the server
answer for a browse request on the method and ta read of all attributes on
InputArguments and the method. Must be a difference somewhere......I
thought maybe because Name value of arg was missing. But it is not...

destogl:
So I found the problem (hopefully the right one :) )

It happens by searching element by node ([_find_element_in_node function in address_space.py file](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/address_space.py#L158)). The reason is that in *QualifiedName*  sent from UaExpert *Namespace Index==0* (TargetName in data structure see figure) and the [checking](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/address_space.py#L158) to be 2.

![input_param ua_exp](https://cloud.githubusercontent.com/assets/1918204/10198980/d07f9d5c-679e-11e5-9e3f-666d7e636e60.png)

For now I will just check the BrowseName. I have also following question: Should the TargetName inherit the NamespaceIndex?

destogl:
Solved in #35

oroulet:
I do not understand, I do not see any translateBrowsePathToNodeId request in wireshark before calling a method from uaexpert

oroulet:
I think I found a much more probable reason for the error.
ArrayDimension is 0 in freeopcua server for method while ArrayDimentsino seems to be the number of arguments in prosys server. Looks like ArrayDimensino must be set up when Value is an extensionObject of a given size. Maybe it should also be set for any array... something to look at

destogl:
[...]

PR #35, opened 2015-09-30
Showing paramters in UaExpert

destogl:
Debug of TranslateBrowsePathsToNodeIds service. Still should be checked if it's implemented according to specification. Details in #34 

Current creation of new Arguments in the example-server is not so nice. We should think about a way to avid method *_guess_uatype* and enter type directly.

Also node.create_method should be extended so user can define a description.

Also mymethod is crashing the UaExpert... Don't merge for now.

### DIFF
--- a/example-server.py
+++ b/example-server.py
@@ -12,7 +12,7 @@ except ImportError:
         shell.interact()
 
 
-from opcua import ua, uamethod, Server, Event, ObjectIds
+from opcua import ua, uamethod, Server, Event, ObjectIds, node
 
 
 class SubHandler(object):
@@ -78,7 +78,27 @@ if __name__ == "__main__":
     myarrayvar = myobj.add_variable(idx, "myStronglytTypedVariable", ua.Variant([], ua.VariantType.UInt32))
     myprop = myobj.add_property(idx, "myproperty", "I am a property")
     mymethod = myobj.add_method(idx, "mymethod", func, [ua.VariantType.Int64], [ua.VariantType.Boolean])
-    multiply_node = myobj.add_method(idx, "multiply", multiply, [ua.VariantType.Int64, ua.VariantType.Int64], [ua.VariantType.Int64])
+
+    inargx = ua.Argument()
+    inargx.Name = "x"
+    inargx.DataType = node._guess_uatype(ua.Variant(None, ua.VariantType.Int64))
+    inargx.ValueRank = -1
+    inargx.ArrayDimensions = []
+    inargx.Description = ua.LocalizedText("First number x")
+    inargy = ua.Argument()
+    inargy.Name = "y"
+    inargy.DataType = node._guess_uatype(ua.Variant(None, ua.VariantType.Int64))
+    inargy.ValueRank = -1
+    inargy.ArrayDimensions = []
+    inargy.Description = ua.LocalizedText("Second number y")
+    outarg = ua.Argument()
+    outarg.Name = "Result"
+    outarg.DataType = node._guess_uatype(ua.Variant(None, ua.VariantType.Int64))
+    outarg.ValueRank = -1
+    outarg.ArrayDimensions = []
+    outarg.Description = ua.LocalizedText("Multiplication result")
+
+    multiply_node = myobj.add_method(idx, "multiply", multiply, [inargx, inargy], [outarg])
 
     # import some nodes from xml
     server.import_xml("custom_nodes.xml")
--- a/example-server.py
+++ b/example-server.py
@@ -31,7 +31,10 @@ class SubHandler(object):
 # method to be exposed through server
 
 def func(parent, variant):
-    return [variant.Value * 2]
+    ret = False
+    if variant.Value % 2 == 0:
+        ret = True
+    return [ua.Variant(ret, ua.VariantType.Boolean)]
 
 
 # method to be exposed through server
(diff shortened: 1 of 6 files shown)


destogl:
The UaExpert is crashing because ValueRank paramters is per default set to 0 ([check here](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/uaprotocol_auto.py#L784) and should be -1 if the data is scalar.

ValueRank is described in Part 3, Page 61, Table 23. Can we add this somewhere to be autogenerated as constant dictionary? Also automatic checking of this should be implemented.

opcua/address_space.py
```diff
@@ -155,7 +156,7 @@ def _find_element_in_node(self, el, nodeid):
         nodedata = self._aspace[nodeid]
         for ref in nodedata.references:
             # FIXME: here we should check other arguments!!
-            if ref.BrowseName == el.TargetName:
+            if ref.BrowseName.Name == el.TargetName.Name:
                 return ref.NodeId
```
oroulet: that cannot be correct, they must be in same namespace

opcua/address_space.py
```diff
@@ -139,6 +139,7 @@ def _translate_browsepath_to_nodeid(self, path):
             res.StatusCode = ua.StatusCode(ua.StatusCodes.BadNodeIdInvalid)
             return res
         current = path.StartingNode
+        # FIXME: Should be more according to the Specs: Part 4 Page 44
```
oroulet: what is missing? I do not remember there is any shortcut here

destogl:
Maybe is nothing missing, but it was kind a strange so I wanted to check it again.

I send you tomorrow all logs, so you can see it.

oroulet:
The UaExpert is crashing because ValueRank paramters is per default set to
0 (check here
<https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/uaprotocol_auto.py#L784>
and should be -1 if the data is scalar.

OK we need to fix this one

oroulet:
You can override default value at the end of uaprotocol_hand.py

opcua/address_space.py
destogl: see comment in #34

opcua/address_space.py
destogl: Just checked it. It is good. Yesterday seamed strange... sorry

Even though you are not checking inverse references and subtypes... I don't know if this is important.

destogl:
ValueRank is now corrected. I decided to define it at -2 per default because of description (Part 3, Page 61, Table 23):

```
Indicates whether the dataType is an array and how many dimensions the array has.
It may have the following values:
n > 1: the dataType is an array with the specified number of dimensions.
OneDimension (1): The dataType is an array with one dimension.
OneOrMoreDimensions (0): The dataType is an array with one or more dimensions.
Scalar (−1): The dataType is not an array.
Any (−2): The dataType can be a scalar or an array with any number of dimensions.
ScalarOrOneDimension (−3): The dataType can be a scalar or a one dimensional array.
NOTE All DataTypes are considered to be scalar, even if they have array-like semantics like ByteString and String.
```

oroulet:
Is uaexpert happy now?

oroulet:
Yes -2 looks like a good default. Just hope most clients support it....

oroulet:
I add one more look. I could see the translatebrowsepath request asking for
a child with browsename 0:InputArguments. I then checked the prosys opc-ua
server (the prosys guys are very goos at following the opc-ua spec) and
there I could see that the browsename of inputarguments seems to always be
in namespace 0. event of method has another namespace. In the spec I read
"""
Servers may often choose to use the same namespace for the NodeId and the
BrowseName. However, if they want to provide a standard Property, its
BrowseName shall have the namespace of the standards body although the
namespace of the NodeId reflects something else, for example the local
server.
"""
that seems to confirm the InputArgument should use namespace 0.....and that
translatebrowsepath should compare namespace.
You need to modify the lines:
if inputs: create_property(method, qname.NamespaceIndex, "InputArguments",
[_vtype_to_argument(vtype) for vtype in inputs])
 if outputs: create_property(method, qname.NamespaceIndex,
"OutputArguments", [_vtype_to_argument(vtype) for vtype in outputs])

and write something like
ceate_property(method, GenerateNodeId(), QualifiedName(0,
"InputArguments"), [_vtype_to_argument(vtype) for vtype in inputs])
that shpuld make it

destogl:
Now should be OK.
Should the FIXME from [here](https://github.com/FreeOpcUa/python-opcua/pull/35/files#diff-9bce3cb93ffe2484707e6e3b20c4881dL157) be removed?

I also corrected example-server mymethod didn't work.

example-server.py
```diff
@@ -74,11 +77,32 @@ def multiply(parent, x, y):
...
+    inargy.ArrayDimensions = []
+    inargy.Description = ua.LocalizedText("Second number y")
+    outarg = ua.Argument()
+    outarg.Name = "Result"
+    outarg.DataType = node._guess_uatype(ua.Variant(None, ua.VariantType.Int64))
+    outarg.ValueRank = -1
+    outarg.ArrayDimensions = []
+    outarg.Description = ua.LocalizedText("Multiplication result")
+
+    multiply_node = myobj.add_method(idx, "multiply", multiply, [inargx, inargy], [outarg])
```
oroulet: I got several complains about the example file being too big, so I do not want to merge such a huge thing. Can you put it in another file? like for example-server-methods.py? Then we just leave the simplest method example in example-server.py?

example-server.py
```diff
@@ -12,7 +12,7 @@ def embed():
         shell.interact()
 
 
-from opcua import ua, uamethod, Server, Event, ObjectIds
+from opcua import ua, uamethod, Server, Event, ObjectIds, node
```
oroulet: what is "node" with a small n? don't you get an error here?

destogl: It is the node.py file

PR #36, opened 2015-10-01
Added decorator for class members

destogl:
Added new decorator to automatically convert from and to variants for method which are class members since other is not working with them.

### DIFF
--- a/opcua/__init__.py
+++ b/opcua/__init__.py
@@ -29,18 +29,14 @@ def uamethod(func):
     arguments and output to and from variants
     """
     def wrapper(parent, *args):
-        result = func(parent, *[arg.Value for arg in args])
-        return to_variant(result)
-    return wrapper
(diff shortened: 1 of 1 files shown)


oroulet:
Can't we merge both to have only one decorator?

destogl:
Yes maybe, but it can be tricky, since if there is member class you always get "self" as the first argument... and in other case you get parent...

oroulet:
then we should rename them
uamethod should in fact be uafunction and the other uamethod.
Can you do it?

oroulet:
not sure now. method is the name used by opcua

oroulet:
I vote for rename anyway

destogl:
Just wait a sec, I try to hack it into one and than we decide...

destogl:
Solved, I donesn't look bad :)

so the name can stay...

Issue #38, opened 2015-10-02
Booleans converted to ints


drakes00:
Hello,

As in python ```isinstance(True,int) == True```, all boolean values passed to a Variant will be converted to Int64 when going through Variant._guess_type at line 741 of python-opcua/opcua/uatypes.py.

Is this wanted to avoid the use of booleans ?

Best,
Maxime

destogl:
I would say it is error, I just also has it. That's why I had to make it "dirty" [here](https://github.com/FreeOpcUa/python-opcua/pull/35/files#diff-9a04e0625aca432894645155aefd62f6R37). Let me just take a look...

destogl:
#39 should solve your problem

drakes00:
Thanks a lot :-)

destogl:
@oroulet Can I merge?

PR #39, opened 2015-10-02
Correctly guess booleans

destogl:
this should solve #38.

Ints are tested and still converted correctly.


opcua/uatypes.py
```diff
@@ -736,6 +736,9 @@ def __eq__(self, other):
     def _guess_type(self, val):
         if val is None:
             return VariantType.Null
+        elif isinstance(val, bool):
+            # TDA, added this because it was missing and causes exceptions when 'bool' type is used
```
oroulet: since you are there, can you remove that comment line....

PR #45, opened 2015-10-23
Small fixes

alkor:
Fix some typos in messages. Set zero timestamp in BrowseParameters.View, because .Net UA servers show a strange behaviour when this timestamp is a current time.

### DIFF
--- a/opcua/uatypes.py
+++ b/opcua/uatypes.py
@@ -869,7 +869,7 @@ class DataValue(FrozenClass):
         return obj
 
     def __str__(self):
-        s = 'DataValue(Value:{})'.format(self.Value)
+        s = 'DataValue(Value:{}'.format(self.Value)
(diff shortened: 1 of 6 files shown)


opcua/uatypes.py
```diff
@@ -869,7 +869,7 @@ def from_binary(data):
         return obj
 
     def __str__(self):
-        s = 'DataValue(Value:{})'.format(self.Value)
+        s = 'DataValue(Value:{}'.format(self.Value)
```
oroulet: all good apart from this one. merging anyway. hopefully one of use will remember to fix it back

alkor: No, this bracket is not needed here. Just look at the next lines, the closing bracket is added at the end.

oroulet: you are right. this was not visible in diff...

PR #54, opened 2015-11-09
add script for generating debian package

cedricboudinet:
Hi here is a script for generating a debian package file
Cedric

### DIFF
--- a/make_deb.sh
+++ b/make_deb.sh
@@ -1,0 +1,8 @@
+#! /bin/sh
+#
+# This script generates a .deb file for generating a debian package of python-opcua
+# You need to install python-stdeb to use it.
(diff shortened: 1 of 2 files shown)


oroulet:
Thanks. One comment though, this seems to be hardcoded for python2. shouldn't it be a way to says that there is no dependencies for python > 3.4?

cedricboudinet:
There are no dependencies if you package for python 3. (The keyword Depends is used to specify dependencies for v2 and Depends3 is used to specify dependencies for v3)
The package is generated for the default python. If python is linked or aliased to python3 the generated file is named python3-freeopcua_0.9.12-1_all.deb.

oroulet:
Sounds good!

PR #58, opened 2015-11-30
More tools

alkor:
New `uahistoryread` tool. Now it only supports reading raw values, other modes will be added in future versions.

Also, fixed some typos.

### DIFF
--- a/opcua/client.py
+++ b/opcua/client.py
@@ -307,6 +307,3 @@ class Client(object):
     def get_namespace_index(self, uri):
         uries = self.get_namespace_array()
         return uries.index(uri)
-
-    def history_read(self, params):
-        return self.bclient.history_read(params)
--- a/opcua/client.py
+++ b/opcua/client.py
@@ -307,3 +307,6 @@ class Client(object):
     def get_namespace_index(self, uri):
         uries = self.get_namespace_array()
         return uries.index(uri)
+
+    def history_read(self, params):
+        return self.bclient.history_read(params)
--- a/opcua/status_code.py
+++ b/opcua/status_code.py
@@ -663,4 +663,4 @@ def get_name_and_doc(val):
     elif val == 0x80B70000:
         return 'BadMaxConnectionsReached', 'The operation could not be finished because all available connections are in use.'
     else:
-        return 'UknownUaError', 'Unknown StatusCode value: {}'.format(val)
+        return 'UnknownUaError', 'Unknown StatusCode value: {}'.format(val)
--- a/opcua/tools.py
+++ b/opcua/tools.py
@@ -400,28 +400,10 @@ def uadiscover():
 
     sys.exit(0)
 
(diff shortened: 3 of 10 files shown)


opcua/status_code.py
```diff
@@ -663,4 +663,4 @@ def get_name_and_doc(val):
     elif val == 0x80B70000:
         return 'BadMaxConnectionsReached', 'The operation could not be finished because all available connections are in use.'
     else:
-        return 'UknownUaError', 'Unknown StatusCode value: {}'.format(val)
+        return 'UnknownUaError', 'Unknown StatusCode value: {}'.format(val)
```
oroulet: as the first line states this file is autogenerated :-), you need to fix generate_statuscodes.py under schemas

alkor: Yes, I fixed this in both places :-)

opcua/client.py
```diff
@@ -307,3 +307,6 @@ def get_namespace_array(self):
     def get_namespace_index(self, uri):
         uries = self.get_namespace_array()
         return uries.index(uri)
+
+    def history_read(self, params):
+        return self.bclient.history_read(params)
```
oroulet: the Client class is supposed to be a very high level class and expose BinaryClient for users who want more control, so there is no need to expose the same method in Client and BinaryClient. (btw maybe we should rename bclient to binary_client to make it more discoverable)
I do not remember well the history API but I think that the history_read or better read_history method should be moved to Node class. Maybe just move read_raw_history to Node and call it read_history, if you think it makes sense

opcua/tools.py
```diff
@@ -398,4 +400,69 @@ def uadiscover():
 
     sys.exit(0)
 
+def read_raw_history(client, node_id, startTime, endTime):
```
oroulet: move to Node class

opcua/status_code.py
oroulet: you allready did it!!! Sorry for the noise

setup.py
```diff
@@ -33,6 +33,7 @@
                         'uabrowse = opcua.tools:uals',
                         'uawrite = opcua.tools:uawrite',
                         'uasubscribe = opcua.tools:uasubscribe',
+                        'uahistoryread = opcua.tools:uahistoryread',
```
oroulet: great! nice to finally get som history supprt

opcua/client.py
oroulet: Then you can add a simple history_read(node, start_time, end_time) to Client if you think it is usefull

alkor:
@oroulet Hmm, yes, I'll move history functions to Node class.

opcua/client.py
oroulet: maybe method in Client should allow to read several nodes at once.... you decide ....

alkor:
@oroulet I moved these methods to Node and removed history_read from Client.

PR #59, opened 2015-12-01
Clean up ExtensionObject handling

alkor:
OK, this patch is really huge. The main idea here is to hide ExtensionObject conversions from user of this library.

All ExtensionObject members are now automatically converted to objects of correct type. Also, their default value is now None.

For example, now you can write:
```
if isinstance(params.UserIdentityToken, ua.UserNameIdentityToken):
    print(params.UserIdentityToken.UserName)
```
instead of explicitly converting object:
```
id_token = ua.downcast_extobject(params.UserIdentityToken)
if id_token.TypeId == ua.FourByteNodeId(ua.ObjectIds.UserNameIdentityToken_Encoding_DefaultBinary):
    print(id_token.UserName)
```

I removed downcast_extobject() and ExtensionObject class, they are no longer needed.

### DIFF
--- a/opcua/uaprotocol.py
+++ b/opcua/uaprotocol.py
@@ -4,13 +4,3 @@ from opcua.status_code import StatusCodes
 from opcua.uaprotocol_auto import *
 from opcua.uaprotocol_hand import *
 
-
-# FIXME: this is really crappy, should thing about a better implementation
-# maybe never inherit extensionobject and parse only body....
-def downcast_extobject(item):
-    if item.TypeId.Identifier == 0:
-        return item
-    objectidname = ObjectIdsInv[item.TypeId.Identifier]
-    classname = objectidname.split("_")[0]
-    cmd = "{}.from_binary(utils.Buffer(item.to_binary()))".format(classname)
-    return eval(cmd)
--- a/schemas/generate_protocol_python.py
+++ b/schemas/generate_protocol_python.py
@@ -31,6 +31,22 @@ class CodeGenerator(object):
                 continue
             self.generate_struct_code(struct)
 
+        self.iidx = 0
+        self.write("")
+        self.write("ExtensionClasses = {")
+        for struct in self.model.structs:
+            if struct.name in IgnoredStructs:
+                continue
+            if struct.name.endswith("Node") or struct.name.endswith("NodeId"):
+                continue
+            if "ExtensionObject" in struct.parents:
+                self.write("    ObjectIds.{0}_Encoding_DefaultBinary : {0},".format(struct.name))
+        self.write("}")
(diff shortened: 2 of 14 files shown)


opcua/uaprotocol.py
```diff
@@ -4,13 +4,3 @@
 from opcua.uaprotocol_auto import *
 from opcua.uaprotocol_hand import *
 
-
-# FIXME: this is really crappy, should thing about a better implementation
-# maybe never inherit extensionobject and parse only body....
-def downcast_extobject(item):
```
oroulet: removing this method is welcome.... just need to understand how you made it....

tests/tests.py
```diff
@@ -19,6 +19,8 @@
 from opcua import Event
 from opcua import ObjectIds
 from opcua import AttributeIds
+from opcua.uaprotocol_auto import extensionobject_from_binary
+from opcua.uaprotocol_auto import extensionobject_to_binary
```
oroulet: this is a bit strange why not "from opcua.uaprotocol import extensionobject_from_binary"?

schemas/generate_protocol_python.py
```diff
@@ -31,6 +31,22 @@ def run(self):
                 continue
             self.generate_struct_code(struct)
 
+        self.iidx = 0
+        self.write("")
+        self.write("ExtensionClasses = {")
```
alkor: I added a generated dict { typeid -> class } here...

schemas/uaprotocol_auto_add.py
```diff
@@ -0,0 +1,33 @@
+def extensionobject_from_binary(data):
```
alkor: ... and then created these two functions (they go to the end of uaprotocol_auto.py)...

schemas/uaprotocol_auto_add.py
```diff
@@ -0,0 +1,33 @@
...
+    Convert binary-coded ExtensionObject to a Python object.
+    Returns an object, or None if TypeId is zero
+    """
+    TypeId = NodeId.from_binary(data)
+    Encoding = unpack_uatype('UInt8', data)
+    if Encoding & (1 << 0):
+        Body = unpack_uatype('ByteString', data)
+    if TypeId.Identifier == 0:
+        return None
+    klass = ExtensionClasses[TypeId.Identifier]
```
alkor: Finding class by TypeId

oroulet:
Well done! Looks like you managed to remove one of the ugliest thing in python-opcua.....

tests/tests.py
alkor: OK, I'll fix this

PR #62, opened 2015-12-07
Adds pycrypto requirement and corrects server example in README

kevincolyar:


oroulet:
Thanks, I forgot those

PR #63, opened 2015-12-08
Fix conversion from FILETIME to datetime

alkor:
Previously, datetime.utcfromtimestamp() was used,
which is limited to 1970-2038 range on 32-bit POSIX systems.
Adding timedelta to datetime(1601, 1, 1) works better.


oroulet:
yes much better!

PR #64, opened 2015-12-09
XML importer updates

zerox1212:
Extended the xml parser to cover the basic python data types.
Fixed a bug when the xml parser tried to import a null string.
Fixed a bug where the xml importer wouldn't assign ua node data type because the imported value was a zero.


oroulet:
Looks good

PR #67, opened 2015-12-10
Refactor command-line tools

alkor:
Remove duplicate code in uatools.

Also, add default scheme opc.tcp:// if URL provided in arguments has no scheme.


oroulet:
well done as always :-)

PR #68, opened 2015-12-10
Choose PolicyId from list of policies

alkor:
In ActivateSession parameters, PolicyId must match the Id of needed
UserTokenPolicy of endpoint. Some servers ignore this Id, but some
(like OPC Foundation's .Net UA library) return an error if PolicyId is invalid.


oroulet:
I trust you

PR #72, opened 2015-12-15
guid support corrected

jadro:


oroulet:
the code looks ok, but you have a typo that breaks the build https://travis-ci.org/FreeOpcUa/python-opcua/builds/96972169
Can you also use rebase to remove the unncessary  merge commit? or simply clone again and copy your changed file over

oroulet:
merged

PR #74, opened 2015-12-17
uadiscover was broken after renaming of functions

alkor:


oroulet:
sorry I forgot to update those

PR #76, opened 2015-12-19
Performance enhancement

huazh:
Hi,
This is a performance enhancement patch based on client read/write/subscribe profile. It speed up about 3~5 times depends on test case. I think it should speed up server application too.
The changes are:
1, utils.Buffer don't truncate the saved data on each read(), just remember the reading position.
2, remove class FrozenClass, using `__slots__` to restrict attributes access. Even though it is arguable using __slots__ to restrict access, but performance gain is huge here.
3, rewrite auto generated class in uaprotocol_auto.py, make __init__ of them accept a data argument. if data is not None, it will initialize attribute from it as binary Buffer. It remove duplicated attribute initialization.
4, rewrite pack/unpack code. inline direct pack/unpack function in Type object to/from_binary
5, remove duplicated to_binary call in BinaryClient.send_request() as to_binary on big object is a heavy operation.


oroulet:
Thank for an interresting patch. Please next time send improvments one by one and not as one big patch, then it is much easier to discuss changes. Some changes are very good like like improvement in Buffer class but others will need discussion. This __slots__ stuff is really verbose, OK for autogenerated classes but a mess for hand written classes. I am looking at the rest....

oroulet:
and and please fix tests

huazh:
Sorry for the big patch.
It was quick hack to test performance gain in beginning. Then fixed the bugs found in the progress. So it looks like I can not simply split the patch one by one. But if it really need, I can give it a try.
the test should be fixed in the new commit

huazh:
The use of `__slots__`  have 2 problems:
1, make code verbose
2, every subclass must define `__slots__`, even no new attribute is defined
But it give a big difference in the performance. Sample test code as below.
If the performance is attractive, hand written types must use the __slots__ implementation as the hand written types ( e.g. NodeId/DataValue/Variant) are the most used types.

```python
class FrozenClass(object):

    """
    make it impossible to add members to a class.
    This is a hack since I found out that most bugs are due to misspelling a variable in protocol
    """
    __isfrozen = False

    def __setattr__(self, key, value):
        if self.__isfrozen and not hasattr(self, key):
            raise Exception("Error adding member '{}' to class '{}', class is frozen, members are {}".format(key, self.__class__.__name__, self.__dict__.keys()))
        object.__setattr__(self, key, value)

    def _freeze(self):
        self.__isfrozen = True


class A(FrozenClass):
    def __init__(self):
        self.a = 0
        self._freeze()


class B(object):
    __slots__ = ['a']

    def __init__(self):
        self.a = 0

from timeit import timeit

a = A()
b = B()


def f(o):
    for _ in range(100000):
        o.a = 1

print(timeit("f(a)", "from __main__ import f, a, b", number=100))
print(timeit("f(b)", "from __main__ import f, a, b", number=100))
```
On python 2.7
```
$ python ../sp.py
7.31924845319
0.617287770077
```

On Python 3.4
```
$ python ../sp.py  
7.431930915545557  
0.7190417722480396 
```

oroulet:
just for fun:
pypy t.py 
0.00878620147705
0.0912971496582

oroulet:
what about this then:
class FrozenClass(object):

    """
    make it impossible to add members to a class.
    This is a hack since I found out that most bugs are due to misspelling a variable in protocol
    """
    #__isfrozen = False

    #def __setattr__(self, key, value):
        #if self.__isfrozen and not hasattr(self, key):
            ##raise Exception("Error adding member '{}' to class '{}', class is frozen, members are {}".format(key, self.__class__.__name__, self.__dict__.keys()))
        #object.__setattr__(self, key, value)

    def _freeze(self):
        #self.__isfrozen = True
        self.__slots__ = self.__dict__.keys()


class A(FrozenClass):
    def __init__(self):
        self.a = 0
        self._freeze()


class B(object):
    __slots__ = ['a']

    def __init__(self):
        self.a = 0
        #self._freeze()

    def _freeze(self):
        pass

from timeit import timeit

a = A()
b = B()


def f(o):
    for _ in range(100000):
        o.a = 1

print(timeit("f(a)", "from __main__ import f, a, b", number=100))
print(timeit("f(b)", "from __main__ import f, a, b", number=100))

oroulet:
the code formating is not good :-) but byt setting __slots__ dynamically we avoid verbose and tedious listing of methods. Can this work?

huazh:
`__slots__` must be class attribute and statically defined.
The intended usage of `__slots__` is to reduce object memory usage. If a class have `__slots__`, `__dict__` will disappear

oroulet:
I agree about everything but the '__slots__' stuff. The FrozenClass stuff was an unintrusive way to avoid mistake. having to manually define __slots__ is a NoGo. Can you create a patch without it? (you probably can easily create it by using git revert for the specific __slots__ commit then do a reset to master then a new commit which will then contain everything. You can then create a new pull specifically for the __slots__ stuff we can discuss further

huazh:
ok.
But it will be a job for tomorrow. I'm living in GMT+8.

oroulet:
great!

huazh:
another simple option for the slot problem would be change the checking to noop dynamicly while in production env. Such as an env pyopcua_no_typo_check

oroulet:
> another simple option for the slot problem would be change the checking to noop dynamicly while in
would be great but how?

huazh:
using environment variable to decide the impl of FrozenClass. like:
```python
if "PYOPCUA_NO_TYPO_CHECK" in os.environ:
    FrozenClass = object
else:
    FrozenClass = _FrozenClass
```
sample code here: https://github.com/huazh/python-opcua/commit/4c041a14dfe808225a0c11b670b4c79e8477706e

oroulet:
your last suggestion looks OK for me. it even simplifies FrozenClass

huazh:
You could merge the commit. It's test on Python2/3.

oroulet:
better if you create a pull request then the automated tests will run

PR #77, opened 2015-12-20
performance enhancement 2

huazh:
the same as https://github.com/FreeOpcUa/python-opcua/pull/76, but without `__slot__`


opcua/binary_client.py
```diff
@@ -11,6 +11,15 @@
 import opcua.utils as utils
 
 
+class FakeRequest(object):
+    # HACK to remove duplicate call Request.to_binary
```
oroulet: Good idea but please call it CachedRequest or similar and remove comment

oroulet:
otherwise all good I will merge it

PR #79, opened 2015-12-20
Split messages into MessageChunks

alkor:
Client and server should respect BufferSize values and
should not send chunks bigger than peer's BufferSize.
Also, added base class for SecurityPolicy. Only None policy (no signing, no encryption) is implemented now, but I already have usable client-side security policies (tested on OPC foundation .net server), and I will push them after some more testing.

### DIFF
--- a/opcua/binary_client.py
+++ b/opcua/binary_client.py
@@ -24,7 +24,7 @@ class UASocketClient(object):
     handle socket connection and send ua messages
     timeout is the timeout used while waiting for an ua answer from server
     """
-    def __init__(self, timeout=1):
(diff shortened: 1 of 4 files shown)


oroulet:
own class for Policy and support for msg chunk is good. well done. Tests are broken but it is probably not your fault, seems to be some hanging somewhere with pypy randomly....

oroulet:
I did some simple work with cryptography in another pull request. you can have a look if you did something similiar. cryptography modules seems very good. I prefer to use it than re-implementing everything

alkor:
Yes, I used cryptography too. I can rewrite my code to use your functions.

oroulet:
You really do not have to reuse my functions! Do what you think is best!

PR #81, opened 2015-12-20
Security policies implementation

alkor:
Basic128Rsa15 and Basic256 are implemented.
Client now successfully connects to OPC foundation .net server with signing and encryption.

Dependency on cryptography module is completely optional: you
can use python-opcua without cryptography if you only use #None policy.

This pull request depends on #79.

I hope I didn't break anything :-)

### DIFF
--- a/opcua/binary_client.py
+++ b/opcua/binary_client.py
@@ -10,15 +10,6 @@ from concurrent.futures import Future
 import opcua.uaprotocol as ua
 import opcua.utils as utils
 
-
-class CachedRequest(object):
-    def __init__(self, binary):
-        self.binary = binary
-
-    def to_binary(self):
-        return self.binary
-
-
 class UASocketClient(object):
     """
     handle socket connection and send ua messages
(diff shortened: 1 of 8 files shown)


opcua/binary_client.py
```diff
@@ -10,21 +10,12 @@
 import opcua.uaprotocol as ua
 import opcua.utils as utils
 
-
-class CachedRequest(object):
```
oroulet: Is this normal? It look slike you reverted some stuff I merged this morning.....

oroulet:
This is really great. I will wait to check nothing is broken but we can probably do a 1.0 release now...

opcua/binary_client.py
alkor: Yes, this class was only needed for wrapping binary data and passing it to a function that built requests from objects. Now another function is used, it expects just binary data.

alkor:
I think we need server-side security support before 1.0?

oroulet:
yes probably. but you need most of the work so it should not be very hard on server side.
Looking at the code there are some rough edges.. for example the load_certificate methods save certificate to self.client_certificate while it is the one from security_policy which is used.. maybe it is better that you fix this since you know that code better than I do

oroulet:
then we need to use that code in server and write tests.....

oroulet:
and the entire uacrypto.py file should be removed

oroulet:
or maybe move your entire crypto thing to a uacrypto.py. crypto is complicated we probably need to keep it separated

alkor:
Yes, I'ii try to clean up the code.

oroulet:
I am also wondering about the interface of Client class. Maybe add a
set_security_policy method an the high level 'connect' method could also
accept a security policy arg or just a string 'basic256' for example?

oroulet:
Oh and if you could run pylint/autopep8 --max-line-length to your code it
would be great, there a lot of small wrong pep8 things (or use an editor
with integrated check)

oroulet:
and I get that erro toor:
SyntaxError: Non-ASCII character '\xe2' in file
python-opcua/opcua/security_policies.py on line 254, but no encoding
declared; see http://python.org/dev/peps/pep-0263/ for details ERROR

PR #82, opened 2015-12-20
Set default security policy

alkor:
I've broken command-line tools in previous commit.


oroulet:
we should add tests for command lines..... I am not sure how...

PR #85, opened 2015-12-21
add shadow copy to Buffer

huazh:
change Buffer.copy() to return a shadow copy of Buffer.
add Buffer.skip() to skip length of bytes

### DIFF
--- a/opcua/uaprotocol_auto.py
+++ b/opcua/uaprotocol_auto.py
@@ -14372,12 +14372,22 @@ def extensionobject_from_binary(data):
     """
     TypeId = NodeId.from_binary(data)
     Encoding = ord(data.read(1))
+    body = None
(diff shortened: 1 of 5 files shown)


opcua/uaprotocol_auto.py
```diff
@@ -14372,12 +14372,22 @@ def extensionobject_from_binary(data):
     """
     TypeId = NodeId.from_binary(data)
     Encoding = ord(data.read(1))
+    body = None
```
oroulet: did you modify that file by hand? it is auto generated

huazh: no, uaprotocol_auto_add.py is modified

oroulet:
you are right, got so many patches that I get confused

PR #86, opened 2015-12-21
fix test failed in python2 (windows)

huazh:
1, TestCmdLines should call default python version, not specified python3
2, security_policies.py may failed in python2 as it contains non-ascii char:
```
test_uaread (__main__.TestCmdLines) ... Traceback (most recent call last):
  File "tools/uaread", line 7, in <module>
    from opcua.tools import uaread
  File "D:\subversion\pyopc\pyopcua\tools\..\opcua\tools.py", line 22, in <module>
    from opcua import security_policies
  File "D:\subversion\pyopc\pyopcua\tools\..\opcua\security_policies.py", line 253
SyntaxError: Non-ASCII character '\xe2' in file D:\subversion\pyopc\pyopcua\tools\..\opcua\security_policies.py on line 254, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```
I add file encoding to all python sources to comply PEP 0263


alkor:
About non-ascii characters: security_policies doesn't really need them. I've copy-pasted some text from specs to docstrings, and this text contains unicode dashes. I'll fix this.

But it's a good idea to add coding: utf-8 to all files.

oroulet:
> But it's a good idea to add coding: utf-8 to all files.

Is it? The big advantage of not having encoding specified is that we ensure that all files only contain ascii. Or is now a best practice to have encoding in every python source files?

The python3 stuff was a hack until @alko fixes the encoding of his files :-) Can you send a pull request for that @alkor ?

alkor:
Oh, sorry, I was wrong, PEP-8 explicitly says that you should not add an
encoding declaration for ascii-only files.
22 дек. 2015 г. 9:55 пользователь "ORD" <notifications@github.com> написал:

huazh:
I always declare file encoding as my native language is Chinese. I need to comment in Chinese in my own python script.
But it's a just a style choice, I respect whichever choice

oroulet:
I merged python version stuff. We not do not add encoding header for now

PR #87, opened 2015-12-22
PEP-8 cleanup

alkor:


alkor:
This also removes non-ascii chars mentioned in #86

oroulet:
Are you also looking at removing uacrypto.py? Or should we keep it? We
should also not keep certificates at two places in client..I suppose I need
to take care of server implementation or do you have any plans?

alkor:
I think I'll move cryptographic functions to uacrypto.

About certificates: server_certificate is duplicated, I'll remove it (and there must be checks that server uses the same certificate in all responses). There a two different client_certificate and private_key pairs: one is used for creating SecureChannel and another can be used to identify user in ActivateSession, so there's no duplication.

If you want to implement server-side security, please write. I can do this, but it will take some time (maybe a day).

oroulet:
>>here a two different client_certificate and private_key pairs: one is used for creating SecureChannel and another can be used to identify user in ActivateSession
Aren't a client supossed to use the same certificate for both? I am not a security expert. just wondering...If a certificate is supposed to only be used for identification then maybe we should not store it but only pass it as argument to activate_session?

server-side security is not very high in my todo list. Currently I have random disconnection issues in a larger systems which I need to address and this is going to take me much longer than a day...

alkor:
OK, then I'll make server-side security.

PR #92, opened 2015-12-28
XML importer typo fix

zerox1212:
Had a typo in my earlier XML importer update that I forgot to fix. Importing null strings now works as expected.


oroulet:
would be great if you could write a test for this importer. modify the existing xml test file under tests, set values of different types and test

Issue #42, opened 2015-10-06
Implement support for password in client


mertz3hack:
Looking for help understanding a connection issue.
I am trying to connect to a machine hosted opc ua server using
    <code>client = Client("opc.tcp://user:password@192.168.15.30:4880/Arburg")</code>

I am getting the following error and am not sure if this is something I am doing or if this is not supported yet.

<code>Traceback (most recent call last):
  File "example-client.py", line 41, in <module>
    client.connect()
  File "C:\opcua\opcua\client.py", line 111, in connect
    self.activate_session()
  File "C:\opcua\opcua\client.py", line 199, in activate_session
    params.UserIdentityToken.Password = bytes(self.server_url.password)
TypeError: string argument without an encoding
WARNING:opcua.binary_client:Received an error message type
WARNING:opcua.binary_client:MessageAbort(error:StatusCode(BadInvalidArgument), reason:)
</code>

oroulet:
Yes, there is an issue with the password implementation. I do not know how to encode it. it is probably just a matter of reading the doc , though.... so help welcome here

mertz3hack:
Should this issue be combined into the other issue?

oroulet:
yes there is already an issue on this https://github.com/FreeOpcUa/python-opcua/issues/23 . you are welcome to try to solve this :-)
look at line 199 in client.py. We need to encrypt password and set correct encryption in line under.

To find out what should be there you can read the opc-ua specification (But I could not find anything) and/or start wireshark, set filter type to opcua,  and connect with a client supporting encryption (uaexpert for example). then do the same with python-opcua and look at the difference in the ActivateSession message

oroulet:
btw what server are you using?

mertz3hack:
I am using an Arburg Injection Molding Machine that hosts its own opc-ua server

oroulet:
That's interesting!. Do you manage to connect if you disable password?

oroulet:
yesterday I registered a session between a client and a prosys server:
server return 2 supported PolicyId: username_basic256 and username_basic128
and security policy URI is set to http://opcfoundation.org/UA/SecurityPolicy#Basic256
when sending password in activateSession it refers to following EncryptionAlgorythm
http://www.w3.org/2001/04/xmlenc#rsa-oaep
I have no idea where the key for the encryption is taken from ...

mertz3hack:
I am not able to disable the password
I will try and watch a session to see how it compares

[...]

PR #94, opened 2015-12-30
Clean up client's security policy API

alkor:
Clean up client's security policy API and move all crypto functions to uacrypto.

Added set_security() and set_security_string() methods to Client.
Removed duplicate server_certificate field, renamed
client_certificate and private_key to user_certificate and user_private_key.


alkor:
Also this fixes a typo mentioned in #42, but I doubt that passwords work now.
I have a working server-side security support, I'll push it soon.

oroulet:
Looks good! What do you think is missing for password to work?

PR #95, opened 2015-12-31
Use enums whereever possible

oroulet:
This improve readability a lot!!  but the second commit seems to impact performance dramatically. Do you have any idea on  how we can fix this @huazh ?


huazh:
I think the 3rd commit fixed the problem well enough.
Enum initialization is cpu consuming. But after that, it will be OK.

PR #96, opened 2016-01-01
Support client-side encrypted passwords

alkor:
Checked on OPC Foundation .net UA server

### DIFF
--- a/opcua/client.py
+++ b/opcua/client.py
@@ -357,7 +357,11 @@ class Client(object):
             params.UserIdentityToken.UserName = username 
             if self.server_url.password:
                 pubkey = uacrypto.x509_from_der(self.security_policy.server_certificate).public_key()
-                data = uacrypto.encrypt_basic256(pubkey, bytes(password, "utf8"))
+                # see specs part 4, 7.36.3: if the token is encrypted, password
+                # shall be converted to UTF-8 and serialized with server nonce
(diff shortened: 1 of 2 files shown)


opcua/client.py
```diff
@@ -357,7 +357,11 @@ def activate_session(self, username=None, password=None, certificate=None):
             params.UserIdentityToken.UserName = username 
             if self.server_url.password:
                 pubkey = uacrypto.x509_from_der(self.security_policy.server_certificate).public_key()
-                data = uacrypto.encrypt_basic256(pubkey, bytes(password, "utf8"))
+                # see specs part 4, 7.36.3: if the token is encrypted, password
+                # shall be converted to UTF-8 and serialized with server nonce
+                etoken = ua.pack_bytes(bytes(password, "utf8") + self._server_nonce)
+                #data = uacrypto.encrypt_basic256(pubkey, etoken)
```
alkor: I'm not 100% sure here, but encrypt_basic256() doesn't work for me. It raises UNSUPPORTED_HASH exception:
```
cryptography.exceptions.UnsupportedAlgorithm: This backend supports only SHA1 inside MGF1 when using OAEP
```

oroulet: encrypt_basic256 was not tested. so you are probably correct. I will merge and do a few tests when I get two minutes

PR #101, opened 2016-01-03
partial subscription race condition fix

huazh:
fix two subscription race conditions.
see commit comments for details


oroulet:
the fact that the tests pass cleanly while we were having issue does support your changes :-)

huazh:
There is still race condition in create/delete monitor items. But it required to change the API semantics to fix. The simple solution would be forbid to call notification callback while in create/delete monitor items and remember the notification. After create/delete monitor items, check unprocessing notification and finish it in main thread.

oroulet:
Can you explain the races with a few sentences. I will try to have a look

huazh:
It is the same problem which the first commit try to fix. Notification callback is called in recving thread (as client) or internal server thread (as server). It may be called before Subscription finishes registering the handle for the items in main thread.

oroulet:
But then the symptom is that a warning is sent saying 'received
notification for unknown handle' and a notification is lost. Or is there
another effect?

huazh:
Yes, this is the symptom. The notification is lost.

oroulet:
OK. I knew about that and chose a simple but deficient solution. I will
have a look at your proposition.... Feel free to find an even better one :-)

huazh:
Another solution is move the Subscription item handle register code to recving thread or internal server thread as a callback. But this solution require dramatically in codes.

huazh:
A question:
Why InternalSubscription have a variable _startup set to True which make server return a zero length notification to the client immediately after subscription creation? It may be no monitored items, even no publish request to use. Is this a spec requirement?

oroulet:
> Why InternalSubscription have a variable _startup 

Yes a notification must be sent right after creation even, if nothing no event has been queued. We need a publish request for that but client are supposed to send a publish request immediatly after creation of subscription.

oroulet:
> There is still race condition in create/delete monitor items.

Where is the race condition that is left? your fix looks good I merge it

PR #108, opened 2016-01-08
Improved loading performance when using cached address spaces

stanti:
Hey,

I've tried to run the python-opcua server on a Raspberry-Pi.
Unfortunately starting the server and especially the construction of the address spaces takes quite a while (~57secs), so I started to play with the already built in caching of the address spaces.

To my surprise the loading performance was much worse when loading the the address space from a pickle file (~157secs).

However, simply working with the C-Implementation of the pickle mechanism and using the latest protocol version is delivering better results (cPython: ~33 secs / cPython + HIGHEST_PROTOCOL: ~22 secs).

I didn't do any exact performance tests so you shouldn't take those values too seriously :-)
However, since other persons report similar benchmark results (e.g. http://www.benfrederickson.com/dont-pickle-your-data/), I guess it might also be a performance improvement when running on other platforms...

### DIFF
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -1,7 +1,10 @@
 from threading import RLock
 import logging
-import pickle
 from datetime import datetime
+try:
+    import cPickle as pickle
+except:
+    import pickle
 
 from opcua import ua
 from opcua.server.users import User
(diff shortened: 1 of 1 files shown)


oroulet:
Since you are looking at performance, how does pypy works on rasperry? And how fast is it to load address space from xml instead of code or pickel?

stanti:
The results above were created with python 2.7.
Python3 delivered the fasted import (13secs from pickle / 70 secs from generated code).
Pypy (2.7 compatible) was slower than cpython (93 secs from pickle / 80 secs from generated code).

I never checked the xml import since it didn't finish without errors and I didn't expect it to be faster than pickle/generated code.

PR #110, opened 2016-01-13
Added delete_nodes method to client

zerox1212:
Can someone test the delete_nodes method? Not a big change, but I didn't have time to see if it works.

Also added server_shutdown() call to binary_server similar to binary_client. It seems that calling shutdown() only stops server responses (i.e. the forever loop). It doesn't actually close the socket.


oroulet:
I merged https://github.com/zerox1212/python-opcua/commit/4126a0284a866ba25dcf45b6c5863c2e652ffee9 the last patch is a patch to a file which is not used. the first commit is only noise

PR #114, opened 2016-01-16
Server side deleting of nodes.

destogl:
It is almost finished. There is not clear to me how to notify clients when subscribed node is deleted. As far as I understood, I have to use [trigger_statuschange function](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/server/internal_subscription.py#L185) but don't know how to access it from node... Any help welcome!

### DIFF
--- a/opcua/client/binary_client.py
+++ b/opcua/client/binary_client.py
@@ -423,13 +423,13 @@ class BinaryClient(object):
         response = ua.AddNodesResponse.from_binary(data)
         response.ResponseHeader.ServiceResult.check()
         return response.Results
-        
+
     def delete_nodes(self, nodestodelete):
         self.logger.info("delete_nodes")
         request = ua.DeleteNodesRequest()
         request.Parameters.NodesToDelete = nodestodelete
         data = self._uasocket.send_request(request)
-        response = ua.AddNodesResponse.from_binary(data)
+        response = ua.DeleteNodesResponse.from_binary(data)
         response.ResponseHeader.ServiceResult.check()
         return response.Results
 
--- a/opcua/common/node.py
+++ b/opcua/common/node.py
@@ -254,9 +254,6 @@ class Node(object):
         # FIXME: seems this method may return several nodes
         return Node(self.server, result.Targets[0].TargetId)
 
-    def delete_child(self, path):
-        self.delete_node(self.get_child(path).nodeid)
-
     def read_raw_history(self, starttime=None, endtime=None, numvalues=0, returnbounds=True):
         """
         Read raw history of a node
@@ -317,10 +314,3 @@ class Node(object):
     def call_method(*args, **kwargs):
         from opcua.common import methods
         return methods.call_method(*args, **kwargs)
-
-    def delete_node(self, nodeid, deleterefs=True):
-        deletenode = ua.DeleteNodesItem()
-        deletenode.NodeId = nodeid
-        deletenode.DeleteTargetReferences = deleterefs
-        results = self.server.delete_nodes([deletenode])
-        results[0].check()
--- a/opcua/common/node.py
+++ b/opcua/common/node.py
@@ -254,6 +254,9 @@ class Node(object):
         # FIXME: seems this method may return several nodes
         return Node(self.server, result.Targets[0].TargetId)
 
+    def delete_child(self, path):
+        self.delete_node(self.get_child(path).nodeid)
+
     def read_raw_history(self, starttime=None, endtime=None, numvalues=0, returnbounds=True):
         """
         Read raw history of a node
@@ -288,7 +291,7 @@ class Node(object):
 
     # Hack for convenience methods
     # local import is ugly but necessary for python2 support
-    # feel fri to propose something better but I want to split all those 
+    # feel fri to propose something better but I want to split all those
     # create methods fro Node
 
     def add_folder(*args, **kwargs):
@@ -315,3 +318,9 @@ class Node(object):
         from opcua.common import methods
         return methods.call_method(*args, **kwargs)
 
+    def delete_node(self, nodeid, deleterefs=True):
+        deletenode = ua.DeleteNodesItem()
+        deletenode.NodeId = nodeid
+        deletenode.DeleteTargetReferences = deleterefs
+        results = self.server.delete_nodes([deletenode])
+        results[0].check()
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -254,7 +254,6 @@ class NodeManagementService(object):
                     if rdesc.NodeId == item.NodeId:
                         self._aspace[elem].references.remove(rdesc)
 
-        print self._aspace[item.NodeId]
         del self._aspace[item.NodeId]
 
         return ua.StatusCode()
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -181,7 +181,7 @@ class NodeManagementService(object):
         result = ua.AddNodesResult()
 
         if item.RequestedNewNodeId in self._aspace:
-            self.logger.warning("AddNodeItem: node already exists")
+            self.logger.warning("AddNodesItem: node already exists")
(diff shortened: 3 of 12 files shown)


opcua/client/binary_client.py
```diff
@@ -423,13 +423,13 @@ def add_nodes(self, nodestoadd):
...
         return response.Results
-        
+
     def delete_nodes(self, nodestodelete):
         self.logger.info("delete_nodes")
         request = ua.DeleteNodesRequest()
         request.Parameters.NodesToDelete = nodestodelete
         data = self._uasocket.send_request(request)
-        response = ua.AddNodesResponse.from_binary(data)
+        response = ua.DeleteNodesResponse.from_binary(data)
```
oroulet: yes that was a mistake :-)

opcua/common/node.py
```diff
@@ -315,3 +318,9 @@ def call_method(*args, **kwargs):
         from opcua.common import methods
         return methods.call_method(*args, **kwargs)
 
+    def delete_node(self, nodeid, deleterefs=True):
+        deletenode = ua.DeleteNodesItem()
+        deletenode.NodeId = nodeid
+        deletenode.DeleteTargetReferences = deleterefs
+        results = self.server.delete_nodes([deletenode])
+        results[0].check()
```
oroulet: To fit the current API I would rather have a method delete_nodeS([node]). Then it does not really has its place on Node object. he add_xxnodexx methods exists because adding a node should always be done in the context of a parent node . It tried to move all these add_folder/variable methods to functions but it is not really discoverable. Maybe make it a function and  expose these functions on Server objects?
    class Server:
        def create_variable(self, *args):
              create_variable(*args)
        def delete_nodes(self, [nodes])
              delete_nodes([nodes])

            


 Tand because it makes chaining methods easy. But we do not have the same need for delete_nodes

At the node level I would rather see a delete_node(node) method . Not deleting refs here is asking for trouble. the remove delete_child method. USers can just do mys

opcua/server/address_space.py
```diff
@@ -260,6 +285,36 @@ def _add_reference(self, addref, user):
...
+
+        for rdesc in self._aspace[item.TargetNodeId].references:
+            if rdesc.NodeId is item.SourceNodeId:
+                if rdesc.RefrenceTypeId != item.RefrenceTypeId:
+                    return ua.StatusCode(ua.StatusCode.BadReferenceTypeInvalid)
+                if rdesc.IsForward == item.IsForward or item.DeleteBidirectional:
+                    self._aspace[item.SourceNodeId].references.remove(rdesc)
+
+        return ua.StatusCode()
+
```
oroulet: We will need tests to check that works :-)

oroulet:
Well done work. Looks like you even had to modify "generate_protocol" script and that one is not really documented...

> There is not clear to me how to notify clients when subscribed node is deleted. 

I think the best is to add an argument do datachange_callbacks for example " error=None". Then we addressSpace deletes a node, it check if there are callbackas on that node and call them all with error=StatusCode.BadNodeIdUnknown 

Then let the subscrition enqque the error. I am not exactly sure where the error status code should be set in notfication.... maybe on the DataValue of MonitoredItemNotification? do you have any other proposition?

oroulet:
the internal_subscription should then probably remove it monitored_item for that node, it is dead anyway

opcua/common/node.py
destogl: This is not so clear. Because there is already this function bit receives DeleteNodeItem object. Should I do checking if there is DeleteNodeItem object or NodeId? IMHO it would be nice to have some function which accepts NodeIds so it will be easier for user to use it... Maybe we can call it `delete_nodeIds` than we get following code:

`class Server:
    ...
    def delete_nodeIds(self, nodeIds, deleterefs=True):
        deletenodes = []
        for nodeId in nodeIds:
            deletenode = ua.DeleteNodesItem()
            deletenode.NodeId = nodeid
            deletenode.DeleteTargetReferences = deleterefs
            deletenodes.append(deletenode)
       
        results = self.server.delete_nodes([deletenodes])
        for result in results:
            result.check()
            

    def delete_nodes(self, deletenodes):
         #this is already implemented
`

Th

opcua/server/address_space.py
destogl: This is something I am not sure myself if I implemented it good. I am not sure to understand completely the references by OPC UA.... I implemented it because kind a "stayed in a way" :)

destogl:
> Well done work. Looks like you even had to modify "generate_protocol" script and that one is not really documented...

Thanks, but that was the easier part :)

It should be set `StatusCode.BadNodeIdUnknown ` but this is also no a problem. Problem is to me how to check if a node has monitored items, or better say how to check if there is any callback... after that is everything easy...

The question is: Is MonitoredItem object (internal_subscription.py) accessible from AddressSpace object  (address_space.py)

opcua/common/node.py
oroulet: The idea here was that Node is high level and only uses Node objects, then  session or bclient object is available to give more control to user and accept DeleteNodeItem as argument
Node.delete_nodes([nodes])
server.iserver.isession.delete_nodes(params) ( or client.bclient.delete_nodes(params) on client side)
I should really document that somewhere

opcua/server/address_space.py
```diff
@@ -233,6 +233,31 @@ def _add_node(self, item, user):
...
+            self.logger.warning("DeleteNodesItem: node does not exists")
+            return ua.StatusCode(ua.StatusCodes.BadNodeIdUnknown)
+
+        if item.DeleteTargetReferences:
+            for elem in self._aspace.keys():
+                for rdesc in self._aspace[elem].references:
+                    if rdesc.NodeId == item.NodeId:
+                        self._aspace[elem].references.remove(rdesc)
+
+        del self._aspace[item.NodeId]
```
oroulet: Here instead of deleting it directly you can get the AttributeValueObject

    nodedata = self._aspace[item.NodeId]
    for cb in nodedata.attributes[AttributeId.Value].datachange_callbacks.value():
        try:
              cb(xxxxx, error=ua.StatusCode.UknownNodeId
        except:
             self.logger.exception("Error calling callbackxxxx")
     del self._aspace[item.NodeId]

oroulet: but we should lock here. Not sure why there is no lock. maybe add a delete_node method to AddressSpace class that locks and delete. 

Then you need to modify InternalSuscription to handle that error message and send error notification to clients

destogl: It locks. Because `__delitem__` from AddressSpace class is implicitly called

oroulet: you are right

opcua/common/node.py
destogl: I understand it completely just tought to h

destogl:
@oroulet so now is working properly with your ideas. I find solution very dirty, it would be better to call directly `trigger_statuschange` in `MonitoredItemService` object. becaue standard says:

> When any of the Nodes deleted by an invocation of this Service is being monitored, then a
Notification containing the status code Bad_NodeIdUnknown is sent to the monitoring Client
indicating that the Node has been deleted.

This means als that any monitored item (also event monitoring) should trigger this. For now it works only with subscriptions.

oroulet:
> f you merge it, it would be important to write this into documentation.

I think the current solution is very correct, InternalSubscription is in charge of monitoredItems so it is its responsability to notfify clients. It is true that we do not have currently a way to notify client if a node generating events is deleted. But there is nothing in the way to add a mechanism for it. For example add an attribute EventSubscriber to AttributeValue struct and a method add_event_subscriber(xxx) to AddressSpace that InternalSubscription can call every time a new event subscription is asked by a client. Would be great if you can do it while you are at it

But what we are missing here is tests. delete a node, try to read value after, try to access children, etc..., check the parent has no link to it anymore etc....

destogl:
Ok than the next would be to look at the the tests.

Issue #125, opened 2016-01-25
Allow users to change the password encryption algorithm


sscherfke:
Currently, the high-level client `opcua.Client` uses the algorithm *rsa-oaep* to encrypt session passwords. This is hardcoded in lines [376](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/client/client.py#L367) and [370](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/client/client.py#L370).

However, some servers seem to not suport it but rather require, for example, *rsa-1_5*. In order to support this I have to change `opcua/client/client.py:367+` as follows:

```python
            # data = uacrypto.encrypt_rsa_oaep(pubkey, etoken)
            data = uacrypto.encrypt_rsa15(pubkey, etoken)
            params.UserIdentityToken.Password = data
        params.UserIdentityToken.PolicyId = self.server_policy_id(ua.UserTokenType.UserName, b"username_basic256")
        # params.UserIdentityToken.EncryptionAlgorithm = 'http://www.w3.org/2001/04/xmlenc#rsa-oaep'
        params.UserIdentityToken.EncryptionAlgorithm = 'http://www.w3.org/2001/04/xmlenc#rsa-1_5'
```
A method like `Client.set_pwd_encryption_algorithm(algorithm_string)`, which takes strings like `'rsa-oaep'`, `'rsa-1_5'`, or any other valid algorithm names, would solve this issue.

alkor:
If I understand specs correctly, client must read SecurityPolicyURI from the endpoint description, and then encrypt password with respective asymmetric algorithm. If the endpoint uses Basic128Rsa15 policy, then encrypt with rsa-1_5. If Basic256, then rsa-oaep.

I'll try to implement and test this thing.

oroulet:
Sounds logic

destogl:
Is this solved by #128?

oroulet:
I thjink so but I also remember that someone said there were still an
issue... do not remember where I read it

sscherfke:
Works for me. Thx :)

destogl:
@sscherfke please close the issue than :)

PR #128, opened 2016-02-01
Use correct password encryption algorithm in client

alkor:
Client analyzes SecurityPolicyURI from the endpoint description, and then encrypts password with respective asymmetric algorithm.

This should close issue #125.

Tested on .Net UA server.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -335,6 +335,20 @@ class Client(object):
                 return policy.PolicyId
         return default
 
+    def server_policy_uri(self, token_type):
+        """
+        Find SecurityPolicyUri of server's UserTokenPolicy by token_type.
+        If SecurityPolicyUri is empty, use default SecurityPolicyUri
+        of the endpoint
+        """
+        for policy in self._policy_ids:
+            if policy.TokenType == token_type:
+                if policy.SecurityPolicyUri:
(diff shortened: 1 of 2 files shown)


opcua/client/client.py
```diff
@@ -335,6 +335,22 @@ def server_policy_id(self, token_type, default):
...
+        result = ''
+        for policy in self._policy_ids:
+            if policy.TokenType == token_type:
+                result = policy.SecurityPolicyUri
+                break
+        if result:
+            return result
+        else:
+            return self.security_policy.URI
+
```
oroulet: why not:
for xx in xx:
   if b:
      return
return default

opcua/client/client.py
```diff
@@ -335,6 +335,20 @@ def server_policy_id(self, token_type, default):
...
+        If SecurityPolicyUri is empty, use default SecurityPolicyUri
+        of the endpoint
+        """
+        for policy in self._policy_ids:
+            if policy.TokenType == token_type:
+                if policy.SecurityPolicyUri:
+                    return policy.SecurityPolicyUri
+                else:   # empty URI means "use this endpoint's policy URI"
+                    return self.security_policy.URI
+        return self.security_policy.URI
```
alkor: Yes, this function was ugly. Is it better this way?

PR #130, opened 2016-02-09
Added flexibility by adding functions into server

destogl:
This is to add no input and output arguments if they are not provided. In this way I manage to have the same "interface" for adding methods and other objects into OPC UA server.


oroulet:
Make sense

Issue #60, opened 2015-12-03
Storing OPC UA Server Configuration


zerox1212:
Hello,

I'm working on an analytics program with a "built in" OPC UA server to share the data. The built in server also allows me to adjust variables in the software (via a web based SCADA type GUI).

When I start the server and import all my nodes from XML, everything works good and I can manipulate the variables from the SCADA. However if I close the OPC UA server and start it again, I have to adjust all my variables again, because the XML gets imported again with default values.

What is a good way to store the OPC UA server configuration? I was thinking of "exporting" all the nodes and their values back to XML when the server shuts down. Is this a valid method? Is there some easier way to store my nodes and values?

oroulet:
Interesting question. I always use opc-ua to represent data from an underlying system, thus I do not want to store anything. At start I query the underlying system to be sure to publish up-to date information. But I can imagine that for a very slow underlying system, it might be interesting to save state. InternalServer (available as self.iserver on Server class has a method dump_address_space and load_address_space. they may solve your issue directly. But they dump the entire address space. you may want to only dump selected nodes....
I do not know your application but I think I would prefer to store data internally in my application and then simply write a method to set value of opc-ua node at startup... Where is the drawback of this?

zerox1212:
With your idea you wouldn't you have to manage your nodes in two locations? Your application has to know about the nodes to set the values, and your OPC UA server has to have the nodes existing. 

Or is your application always creating the OPC UA nodes every time it starts? Or if the OPC UA server restarts it also has to check for the nodes and then create them again?

I wanted to add/remove nodes (I've been calling them "tags") by only editing the XML file. When I finish my application I don't want to have to edit python code to add a new tag.

Currently my application works like this:
Server starts
Server creates all nodes from XML files
Server code creates python objects (class) with the same structure as OPC UA object
All the python objects which mirror the OPC UA objects are stored in a "tags" dict
Application code interacts with the python objects internally as it's much simpler than using all the built in python OPC UA methods to do what I need to

I have no clue if this is a bad way to do it, I'm not much of a programmer. Ultimately I just wanted to add or remove tags/nodes by just editing the XML file. I suppose my python object class which mirrors the OPC UA object node could save the data, but I don't know how to do that in python yet.

zerox1212:
Maybe if I show you what I made it will be helpful. I'm open to any advice as I'm new to Python and OPC UA.

My OPC UA object structure: Every "tag" has these attributes:
[...]

PR #132, opened 2016-02-11
Asyncio Transport Closed Explicitly, Fixes for Dumping Address Space

zerox1212:
Fixes for issues discussed at the bottom of https://github.com/FreeOpcUa/python-opcua/issues/60 . Please review changes in binary_server_asyncio.py closely as this is the best solution I was able to come up with. Especially the while loop in stop(), this is the only way I could make sure the transport was really closed before the server is closed. Without this there seems to be a race condition.

I also added the address space dump methods back into the high level server because it seems like it could be a nice feature for people that want to back up their address space easily.

TLDR:
can't dump address space without cleaning up client connections

### DIFF
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -428,8 +428,11 @@ class AddressSpace(object):
         """
         dump address space as binary to file
         """
-        with open(path, 'wb') as f:
-            pickle.dump(self._nodes, f, pickle.HIGHEST_PROTOCOL)
+        try:
+            with open(path, 'wb') as f:
+                pickle.dump(self._nodes, f, pickle.HIGHEST_PROTOCOL)
+        except TypeError:
+            self.logger.info("Failed to dump address space")
 
     def load(self, path):
         """
--- a/opcua/server/binary_server_asyncio.py
+++ b/opcua/server/binary_server_asyncio.py
@@ -25,6 +25,7 @@ class BinaryServer(object):
         self.loop = internal_server.loop
         self._server = None
         self._policies = []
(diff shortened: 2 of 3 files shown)


opcua/server/address_space.py
```diff
@@ -428,8 +428,11 @@ def dump(self, path):
         """
         dump address space as binary to file
         """
-        with open(path, 'wb') as f:
-            pickle.dump(self._nodes, f, pickle.HIGHEST_PROTOCOL)
+        try:
+            with open(path, 'wb') as f:
+                pickle.dump(self._nodes, f, pickle.HIGHEST_PROTOCOL)
+        except TypeError:
+            self.logger.info("Failed to dump address space")
```
oroulet: do not catch things here, it shoulb be catch at higher level

opcua/server/binary_server_asyncio.py
```diff
@@ -25,6 +25,7 @@ def __init__(self, internal_server, hostname, port):
         self.loop = internal_server.loop
         self._server = None
         self._policies = []
+        self.shutdown = [False]  # needs to be immutable list for referencing reasons
```
oroulet: no way :-) This is clearly impossible that this is required

oroulet:
seems to break tests

zerox1212:
Well I tried. I'm still new at programming. I will cancel this pull request and fix address space dumping only in my local code.

However, I still think transport needs to be closed explicitly when OPC UA server shuts down.

Thanks for checking it.

oroulet:
No problem :-) you're attacking the worst thing in programming: multi threading.... I have been doing that kind of things in 15 years and make mistakes all the time...

PR #133, opened 2016-02-12
Events debugging

destogl:
This PR is with some corrections of events implementation. But for a lot of stuff I am not sure what to do. So please do not merge before we answer following questions (changes done according to comments):

1. Do you now in which part of specification are described Event parameters which are send when events happens? I checked in more detail parts 3 and 4 without any success.
2. What should be format of EventId variable? In one implementation of server based on UA SDK is has following format: "len=16, 0xac90cce224ca5d4ca7cfef1a4ba43029".
3. Should Event ID always be the same for the Event? In UA SDK implementation I always get new ones.
4. What does mean all this time-variables? I guess (according to the UA SDK implementation) Time should be time of event not the time of event creation.
5. Is SourceNode variable anywhere else used? Should it be type Node or NodeId?
6. Shouldn't the parent object change its 'EventNotifier' attribute there are events defined for it? If it's not defined than you can not observe events with UAExpert.


oroulet:
As far as I can see all your changes are correct. I could not find the doc but I am quite sure that EventId is an uuid that should change at every fire and the length is up to the server.

    What does mean all this time-variables?
No idea either. I thought it was up to the server. Maybe they should be set to None at Event instanciation and set it to current time when triggering if user has not set it.

But I see we have an issue here. We use the Event object both as an event factory/generator and a event object. It is the source of several errors here.

Can you split this class in 2? One to generate events and one to store event information from a specific event and is returned from SubscriptionHandler?They can inherit the same base class if it reduces code

    Is SourceNode variable anywhere else used? Should it be type Node or NodeId?
Tja, as you want...leave it to NodeId

    Shouldn't the parent object change its 'EventNotifier' attribute there are events defined for it? 
Yes. Server node has it ON by default

destogl:
Yesterday I started with this changes but I didn't manage to finish it. Currently I have to do other suff, so this changes will take some time. I hope in the next 10-15 days it should be finished...

destogl:
@oroulet here is finally my proposal. It is not perfect. I tried to enable users of API to define they own events inheriting BaseEvent object.

Few ideas:
1. Generation of Event classes from XML (not so easy but should be possible)
2. Writeing of missing OPC UA types to close the gap and enable easier developing. This types should probably be just wrappers. E.g. UTCTime (wrapper for datetime), ByteString (wrapper for Varinat ByteString), TimeZoneDataType (wrapper for tzdata)

Does anyone else need this types, or think this would be usefull?

opcua/server/server.py
```diff
@@ -318,12 +318,12 @@ def get_namespace_index(self, uri):
         uries = self.get_namespace_array()
         return uries.index(uri)
 
-    def get_event_object(self, etype=ua.ObjectIds.BaseEventType, source=ua.ObjectIds.Server):
+    def get_event_object(self, event=ua.BaseEvent()):
```
oroulet: rename to get_event_generator.

oroulet:
I think the main idea looks OK. But as you see it breaks tests somewhere... I am also wondering what happens when a user want to create custom events.... I will have a look when I get 2 minutes

destogl:
I tested on my computer and tests are working. Or should I merge it with current master to be easier to check?

oroulet:
sorry I am a bit late on that one, I am wondering how the API should be

destogl:
What do you mean? Regarding custom Events? or something else?

With this implementation triggering of custom Events shouldn't be a problem it is just where are you defining them... but I would concentrate on this when we have this Use-Case since IMHO the `uatypes.py` file could be also a bit changed to define all OPC-UA Types as base types which are then using internally Python types.... But this is different discussion. I just wanted to stress this question is quite complicated and has bigger implications.... (The small notice on types in the few posts before...)

destogl:
I would like to close this PR. should I port it to the newest master?

oroulet:
Yes, please port to master. I will try to have a look again, sorry I had to
prioritize other things

oroulet:
Hi,
I am having some issues with the API here.
* Now we have an EventGenerator class but it does nothing, everything is in the trigger method. We could as well move the method to the Server class
* We harcode the event in your BaseEvent class. We can autogenerate the standard events, but the entire point of events is to have custom events. I think it would be better to have an Event class that  takes an event type node as argument and copies all childs to itself. Or maybe have it more functional: A function generate_event_object(event_type) that takes an event type as argument and returns an event object.  Users can then set member values to whatever they want and call trigger method on Server object....

destogl:
> Now we have an EventGenerator class but it does nothing, everything is in the trigger method. We could as well move the method to the Server class

yes we could do that. I don't see any problems in the near future...

> We harcode the event in your BaseEvent class. We can autogenerate the standard events, but the entire point of events is to have custom events. I think it would be better to have an Event class that takes an event type node as argument and copies all childs to itself. Or maybe have it more functional: A function generate_event_object(event_type) that takes an event type as argument and returns an event object. Users can then set member values to whatever they want and call trigger method on Server object....

The BaseEvent is not hardcoded, since all the custom events has to be descendants of BaseEvent so it make sense to set them in the Generator-class, or as you proposed in the generate-event function...

oroulet:
> The BaseEvent is not hardcoded, since all the custom events has to be descendants of BaseEvent so it make sense to set them in the Generator-class, or as you proposed in the generate-event function...

sounds good... and we need to add a way to generate events based on a type node

destogl:
@oroulet: sorry I am currently busy with other stuff, I will try to finish it until the end of the next week...

oroulet:
no problem, this is open-source...thanks for contributing when you have time..

destogl:
I just pushed the new version...

So I implemented similar as we wrote before... The different is that I need Event object and not EventType in `get_event_object` function. The reason for this is that we have do not have mapping from ObjectId to Types... I added following issue #159  with ideas regarding this

opcua/server/server.py
oroulet: I still think it should be called get_event_generator

oroulet: and that it should take etype=ua.ObjectIds.BaseEventType as argument. As we discussed earlier, a events are something dynamic that can be generated on the fly from the standard namespace or whatever homemade structure

tests/tests_common.py
```diff
@@ -263,54 +262,6 @@ def test_subscribe_events_to_wrong_node(self):
...
-        self.assertEqual(ev.Message.Text, msg)
-        #self.assertEqual(msclt.ev.Time, tid)
-        self.assertEqual(ev.Severity, 500)
-        self.assertEqual(ev.SourceNode, self.opc.get_server_node().nodeid)
-
-        # time.sleep(0.1)
-        sub.unsubscribe(handle)
-        sub.delete()
-
-    def test_events(self):
```
oroulet: why did you move this to server tests? events are supposed to be received on client side. We should test it if we can...

opcua/ua/uatypes.py
```diff
@@ -1176,4 +1174,52 @@ def __str__(self):
...
 
+# TODO: This should be autogeneratd form XML description of EventTypes
+class BaseEvent(FrozenClass):
+
+    '''
+    BaseEvent implements BaseEventType from which inherit all other events and it is used per default.
+
+    '''
 
+    def __init__(self, sourcenode=NodeId(ObjectIds.Server), message=None, severity=1):
```
oroulet: I agree BaseEvent is a good name but I think it should take a  type as argument so we can set all attributes dynamically.

opcua/server/server.py
destogl: I agree, but this is not implementable without much more effort, since we have somehow come to attributes of the event.... In this way is also dynamic... just quicker solution... I could try other way around but it will take some time...

destogl: And also I don't think that name `get_event_generator` is better than `get_event_object`. Probably the most correct would be `generate_event` but anyway I will rename it to `get_event_generator`

opcua/ua/uatypes.py
destogl: Than BaseEvent is not any more good name :) Since BaseEvent is event type.

OK, I think I understand in which direction you would like to go... to have custom events even there is no that class existing... I am really not fun of that approach since everything will be just more messy, but no problem I can adapt :)

tests/tests_common.py
destogl: I remove it because you have here generation of events and this can only server do... So the test with the client is not passing since there is no `get_event_object` function for client and it should not exist. Before we had EventGenerator so that passed.

for really test client to receive event, we should have client and server running at the same time, which I am not sure is possible here... If you have any ides you are welcome :)

destogl:
Huh... Now I see that your first implementation was pretty good, just filed were set wrongly... Unfortunately I understood it wrongly.

So for now I wrote quickly just one proposal and we can go into two directions from here:

1. Extend your (first) implementation to support also Event objects if we want to have them
2. Finish my proposal which is actually just moving some stuff into `server.py` from `event.py`

Which one do you prefer?

destogl:
new clean begin in #161 so closing...

PR #134, opened 2016-02-13
fix race condition in subscription

huazh:
Insert monitored item into map to avoid notification arrive before result return. This is the way how official .net opcua library handle async nature of the subscription.

### DIFF
--- a/opcua/common/subscription.py
+++ b/opcua/common/subscription.py
@@ -292,7 +292,6 @@ class Subscription(object):
                     del self._monitoreditems_map[mi.RequestedParameters.ClientHandle]
                     mids.append(result.StatusCode)
                     continue
-                data = SubscriptionItemData()
                 data = self._monitoreditems_map[mi.RequestedParameters.ClientHandle]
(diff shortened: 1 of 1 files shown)


oroulet:
Does it help? If we do not have server handle, we cannot process notification anyway. Or do I miss domething? BTW tests seem broken.

oroulet:
Maybe it helps because we are even faster but there is still a race

huazh:
sever handle is not need for notification process.
I don't know why it can not pass test. It seems a random timeout.

ORD <notifications@github.com>于2016年2月13日 周六 17:58写道：

huazh:
Can you give me a hint? I found no race warning in my test.

Zhang, Hua <zhanghua@sjtubbs.org>于2016年2月13日 周六 18:02写道：

oroulet:
the test failling is just a timeout because travis was so slow. I should increase it.

Patch looks good! the crazy thing is that this was the original implementation but probably had another bug at that time

oroulet:
changes look good. Well done! I made a new release

PR #140, opened 2016-03-01
Server history support for DataValues (Not events)

oroulet:
low priority, do not know when this will be finished. Feel free to comment


zerox1212:
What do you consider finished? Save the data to SQlite instead of python dicts? I might be able to help with this implementation some if I know what direction you want to go with it.

oroulet:
Yes help to implement other backend, like sqlite is welcome. But wait a bit, I need to change storage interface. Although you can start looking at the python-sqlite interface and thing how to save DataValue and events in sqlite database. We can store binaries but we should expose timestamps as they will be used for time queries. But maybe we should store json or even better ua-xml...

oroulet:
this works with uaexpert and also contains some parsing fixes, thinking to merge soon although no support for event history yet

PR #143, opened 2016-03-08
Handle Asyncio Transports Explicitly

zerox1212:
Please check if this is an acceptable solution for handling the transport objects. I tested it on my local system and it seems to work (with a single client). However it should be tested more. Especially because I am not sure if calling:
```python
self.iserver.asyncio_transports.remove(self.transport)
```
in connection_lost() is really going to remove the correct object.


oroulet:
it looks OK. Does it solve your issue?

zerox1212:
It has solved my issue. Now I can call server.stop() and the client connections get cleaned up. After that you can dump address space to a file without issue.

I only have one client to connect to my OPC UA server though, so I don't know if there is any issue with multiple clients (i.e. multiple transport objects in the list). I think it should be OK, but I have no way to test it.

Issue #148, opened 2016-03-21
Adding new nodes from the clients


destogl:
Hello all,

I was just trying to add some nodes from client, after they are some added from server and there is error that client is asking for already used NodeId. So I check the specs and found this:

> Client requested expanded NodeId of the Node to add. The serverIndex in the
expanded NodeId shall be 0.
If the Server cannot use this NodeId, it rejects this Node and returns the
appropriate error code.
If the Client does not want to request a NodeId, then it sets the value of this
parameter to the null expanded NodeId.
If the Node to add is a ReferenceType Node, its NodeId should be a numeric id.
See Part 3 for a description of ReferenceType NodeIds.

S I wanted to implement second part where client doesn't want to ask NodeId, but I am not sure that I fully understand the meaning of 

> sets the value of this parameter to the null expanded NodeId

I would appreciate your opinion on it.

Thanks!

destogl:
I found it in Part 3 page 59.

Now implementation :)

destogl:
Solved in #149

PR #149, opened 2016-03-22
Implemented support for null nodes and generation of NodeIds only by …

destogl:
…the AddressSpace in the server.

This solves the issue #148 and I would say is the most proper why to generate new NodeIds because until know we had NodeIds generated in the Client and Server separately which would sooner or later make problems.

### DIFF
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -180,34 +180,46 @@ class NodeManagementService(object):
     def _add_node(self, item, user):
         result = ua.AddNodesResult()
 
-        if item.RequestedNewNodeId in self._aspace:
+        # Check if the null NodeId is given
+        if item.RequestedNewNodeId.is_null():
+            self.logger.debug("RequestedNewNodeId is null")
+            if item.BrowseName.NamespaceIndex:
+                idx = item.BrowseName.NamespaceIndex
+            else:
+                idx = item.ParentNodeId.NamespaceIndex
+            nodedata = NodeData(self._aspace.generate_nodeid(idx))
+            item.BrowseName = ua.QualifiedName(item.BrowseName.Name, nodedata.nodeid.NamespaceIndex)
+        else:
+            nodedata = NodeData(item.RequestedNewNodeId)
+
+        if nodedata.nodeid in self._aspace:
             self.logger.warning("AddNodesItem: node already exists")
             result.StatusCode = ua.StatusCode(ua.StatusCodes.BadNodeIdExists)
             return result
-        nodedata = NodeData(item.RequestedNewNodeId)
+
+        if item.ParentNodeId.is_null():
+            #self.logger.warning("add_node: creating node %s without parent", nodedata.nodeid)
+            pass
+        elif item.ParentNodeId not in self._aspace:
+            #self.logger.warning("add_node: while adding node %s, requested parent node %s does not exists", nodedata.nodeid, item.ParentNodeId)
+            result.StatusCode = ua.StatusCode(ua.StatusCodes.BadParentNodeIdInvalid)
+            return result
+
+        if not user == User.Admin:
+            result.StatusCode = ua.StatusCode(ua.StatusCodes.BadUserAccessDenied)
+            return result
+
         # add common attrs
-        nodedata.attributes[ua.AttributeIds.NodeId] = AttributeValue(ua.DataValue(ua.Variant(item.RequestedNewNodeId, ua.VariantType.NodeId)))
+        nodedata.attributes[ua.AttributeIds.NodeId] = AttributeValue(ua.DataValue(ua.Variant(nodedata.nodeid, ua.VariantType.NodeId)))
         nodedata.attributes[ua.AttributeIds.BrowseName] = AttributeValue(ua.DataValue(ua.Variant(item.BrowseName, ua.VariantType.QualifiedName)))
         nodedata.attributes[ua.AttributeIds.NodeClass] = AttributeValue(ua.DataValue(ua.Variant(item.NodeClass, ua.VariantType.Int32)))
         # add requested attrs
         self._add_nodeattributes(item.NodeAttributes, nodedata)
 
-        # add parent
-        if item.ParentNodeId == ua.NodeId():
-            #self.logger.warning("add_node: creating node %s without parent", item.RequestedNewNodeId)
-            pass
-        elif item.ParentNodeId not in self._aspace:
-            #self.logger.warning("add_node: while adding node %s, requested parent node %s does not exists", item.RequestedNewNodeId, item.ParentNodeId)
-            result.StatusCode = ua.StatusCode(ua.StatusCodes.BadParentNodeIdInvalid)
-            return result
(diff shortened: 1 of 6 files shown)


opcua/server/address_space.py
```diff
@@ -180,34 +180,46 @@ def add_nodes(self, addnodeitems, user=User.Admin):
     def _add_node(self, item, user):
         result = ua.AddNodesResult()
 
-        if item.RequestedNewNodeId in self._aspace:
+        # Check if the null NodeId is given
+        if item.RequestedNewNodeId.is_null():
+            self.logger.debug("RequestedNewNodeId is null")
+            if item.BrowseName.NamespaceIndex:
```
oroulet: should we use the index of the browse name? that seems broken...

opcua/ua/uatypes.py
```diff
@@ -409,6 +412,21 @@ def __ne__(self, other):
     def __hash__(self):
         return hash(self.__key())
 
+    def is_null(self):
+        ret = True
+        if self.NamespaceIndex != 0:
```
oroulet: where is the definition of a null node? I cannot find it. I cannot understand that a client can request a nodeid without specifying an namespace index. maybe NodeId(0, 3) if also null? that would solve our issue...

opcua/ua/uatypes.py
```diff
@@ -341,6 +343,7 @@ def __ne__(self, other):
         return not self.__eq__(other)
 
 
+# FIXME: This is not according to standard, check Part 3, Page 59
```
oroulet: what is wrong? cannot find anything

oroulet:
look s all good, just wondering about the default namespaceindex stuff, it should not be necessary, namespace index must be specified one way or another...
ExpandedNodeId has a namespaceUri maybe this one must be specificied? then it is possible to find index using NamespaceIndex table in standard address space

opcua/ua/uatypes.py
destogl: If you check specs Part 3, page ~~59~~ 60 (Table 19) there is definition of null node. You can initialise null NodeId just using `NodeId()` which was already implemented :)

Regarding the NamespaceIndex citation:

> A null NodeId always has a NamespaceIndex equal to 0.

I implemented request for NamespaceIndex through BrowseName which is of type `QualifiedName` and there is NamespaceIndex inside. So the high-level interface to request the NamespaceIndex stayed the same... Check the implementation in manage_node.py of `_parse_add_args`.

opcua/ua/uatypes.py
destogl: NodeIdTypes are IdentifierTypes if I understood correctly, there are defined in Part 3, Page 59, Table 18 so they are:

* NUMERIC_0
* STRING_1
* GUID_2
* OPAQUE_3

Sorry if I got it wrong.

opcua/server/address_space.py
destogl: this works with tests and works with my server/client applications....

Using the NamespaceIndex of the BrowserName was the only solution to get custom NamespaceIndex to the server since the standard definition... see comment down

destogl:
I didn't find anything in the specification regarding this... This was also strange for me, but than I implemented that you can send custom NamespaceIndex by BrowseName (type `QualifiedName`) and if not than the Node get NamespaceIndex of the parent. This seams the most sensible solution for me...

opcua/ua/uatypes.py
oroulet: this seems to be higher level, NodeIdType is about encoding. please remove the FIXME line

opcua/server/address_space.py
oroulet: OK. null NodeId has 0 as idx. so I propose to use the namespace of the parent, not the namespace of the browsename.
And if parent is Null, return BadParentNodeIDInvalid. We should not allow creation of node without valid parent. Then I am fine with PR

opcua/ua/uatypes.py
destogl: OK, sorry

opcua/server/address_space.py
destogl: But if we do that, than a lot of tests will be broken, and you can not add child in the new NameSpace.

Also we can not return hard `BadParentNodeIDInvalid` if the parent is null, I tested it already, but there is a lot of server nodes which does not have parent specified so you can not start server any more...

I agree we should not allow it from the Client, but at this level we have no idea if the server or client are creating nodes. On the other hand this could be solved using `username` variable, where we can define e.g. ServerAdmin. I would propose to merge this and open new Issue regarding this and I am willing to propose me as responsible for issue and than I create proposal for this in the next weeks....

oroulet: This is wrong all nodes should have a a parent, but OK, write a FIXME there
and leave it as is. But can you use the idx of parent instead of browse
name if nodeid is null? and remove the _default_namespace variable, we
should not have such a thing. If no parent and and null nodeid, we refuse
to create...

destogl: OK, but I would like just to remark something...

This will imply bigger changes since in the test it is written that you create NodeId with the NamespaceIndex 4, but since the parent is `Objects` folder you will add it into NamespaceIndex 0 (which is by the way forbidden by the standard). So if we always use the NamespaceIndex of the parent there is no way to have the Nodes in other NamespaceIndex than 0, and you can not specify other NamespaceIndex for the child node (which should be possible)....

Maybe I understood you wrong, please correct me...

oroulet: You are right!!!!!  We need a way to specify namespace.... Must be a
solution.... But using namespace of browse name is wrong... Any suggestion?
Going skiing now :-)

destogl: Have a fun :)

After the analysis of AddNode Service I found BrowseName the most propriate, why that would be wrong? On the end of the day you are mostly searching for nodes through BrowseName which is contained of NamespaceIndex and Name...

oroulet: Now I am back,....what about extending the specification a bit and
creating  new nodeid every time the
identifier is 0 or ""? We should not anyway create such a a NodeID and it
allows us to specify a namespace while letting the server create the nodeid.
We could add a method has_null_identifier() to NodeId. Does it sound
reasonable? We also need to documente this behaviour in add_node method of
InternalServer

destogl: I it not bad, but still what is a difference if you transport the data in the BrowseName? Both solution are extension of specs :)

Sorry that I am stubborn, but I kind a like my solution :)

oroulet: OK merge it. We'll look at it later

oroulet: I think I found an case where it makes trouble ( in addition to my
subjective aversion to it :-) )
When instantiating a node and subnode, it seems to be common to create the
new node tree with browsename in the same namespace as the original
nodes,while node ID are created in new namespace. Especially this is
required for method arguments otherwise they are broken ( must be in idx
0). But we still want the server to create nodeids, in our custom namespace
and preserve original browsename namespace.

PR #150, opened 2016-03-27
use null identifier to trigger generation of nodeid 

oroulet:
add tests

### DIFF
--- a/opcua/common/manage_nodes.py
+++ b/opcua/common/manage_nodes.py
@@ -7,23 +7,27 @@ from opcua.common import node
 
 def _parse_add_args(*args):
     try:
+        if isinstance(args[0], int):
+            nodeid = ua.NodeId(0, int(args[0]))
+            qname = ua.QualifiedName(args[1], int(args[0]))
+            return nodeid, qname
         if isinstance(args[0], ua.NodeId):
             nodeid = args[0]
-            qname = args[1]
         elif isinstance(args[0], str):
             nodeid = ua.NodeId.from_string(args[0])
-            qname = ua.QualifiedName.from_string(args[1])
-        elif isinstance(args[0], int):
-            nodeid = ua.NodeId(0, int(args[0]))
-            qname = ua.QualifiedName(args[1], args[0])
         else:
             raise RuntimeError()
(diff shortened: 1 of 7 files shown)


opcua/common/manage_nodes.py
```diff
@@ -6,22 +6,22 @@
...
-        nodeid = ua.NodeId.from_string(args[0])
-
-    if isinstance(args[1], ua.QualifiedName):
-        qname = args[1]
-    elif isinstance(args[1], str):
-        if isinstance(args[0], int):
+    try:
+        if isinstance(args[0], ua.NodeId):
+            nodeid = args[0]
+            qname = args[1]
```
destogl: IMO we should check if there is ua.QualifiedName object in args[1]

opcua/server/address_space.py
```diff
@@ -198,10 +195,12 @@ def _add_node(self, item, user):
...
 
         if item.ParentNodeId.is_null():
-            #self.logger.warning("add_node: creating node %s without parent", nodedata.nodeid)
+            # self.logger.warning("add_node: creating node %s without parent", nodedata.nodeid)
+            # We should return Error here, but the standard namespace seems to define many nodes
+            # without parents, so ignore...
             pass
         elif item.ParentNodeId not in self._aspace:
-            #self.logger.warning("add_node: while adding node %s, requested parent node %s does not exists", nodedata.nodeid, item.ParentNodeId)
+            self.logger.warning("add_node: while adding node %s, requested parent node %s does not exists", nodedata.nodeid, item.ParentNodeId)
```
destogl: I think this generates a lot of output... don't forget to check

destogl:
Thanks for follow-up :)

opcua/common/manage_nodes.py
oroulet: no we do not since it will be catched by the except and an explicative error message will be thrown

destogl: OK, but we have more flexibility, Because NodeId can be String and and qname QualifiedName object, which is now not possible....

oroulet: That's a point...

PR #155, opened 2016-04-08
Add SQLite History Support for Values

zerox1212:
Tested this with the minimal_server.py example and it works.

Because SQLite objects have to be used on the same thread there are a few weird things that I would like feedback on. 
1. How to decide when to make a new conn object accounting for thread that is making SQL transactions?
2. How to close the SQL connections when they might have been created on other threads?

Please check it and let me know if something should be changed.

### DIFF
--- a/opcua/server/history.py
+++ b/opcua/server/history.py
@@ -4,6 +4,8 @@ from datetime import datetime
 from opcua import Subscription
 from opcua import ua
 
+import sqlite3
+
 
 class HistoryStorageInterface(object):
 
@@ -61,6 +63,13 @@ class HistoryStorageInterface(object):
         """
         raise NotImplementedError
 
+    def stop(self):
+        """
+        Called when the server shuts down
+        Can be used to close database connections etc.
+        """
+        raise NotImplementedError
+
 
 class HistoryDict(HistoryStorageInterface):
     """
@@ -72,15 +81,14 @@ class HistoryDict(HistoryStorageInterface):
         self._events = {}
 
     def new_historized_node(self, node, period, count=0):
-        self._datachanges[node] = []
-        self._datachanges_period[node] = period, count
-
-    def new_historized_event(self, event, period):
-        self._events = []
+        node_id = node.nodeid
+        self._datachanges[node_id] = []
+        self._datachanges_period[node_id] = period, count
 
     def save_node_value(self, node, datavalue):
-        data = self._datachanges[node]
-        period, count = self._datachanges_period[node]
+        node_id = node.nodeid
+        data = self._datachanges[node_id]
+        period, count = self._datachanges_period[node_id]
         data.append(datavalue)
         now = datetime.now()
         if period:
@@ -90,8 +98,9 @@ class HistoryDict(HistoryStorageInterface):
             data = data[-count:]
 
     def read_node_history(self, node, start, end, nb_values):
+        node_id = node.NodeId
         cont = None
-        if node not in self._datachanges:
+        if node_id not in self._datachanges:
             print("Error attempt to read history for a node which is not historized")
             return [], cont
         else:
@@ -99,7 +108,7 @@ class HistoryDict(HistoryStorageInterface):
                 end = datetime.now() + timedelta(days=1)
             if start is None:
                 start = ua.DateTimeMinValue
-            results = [dv for dv in self._datachanges[node] if start <= dv.ServerTimestamp <= end]
+            results = [dv for dv in self._datachanges[node_id] if start <= dv.ServerTimestamp <= end]
             if nb_values:
                 if start > ua.DateTimeMinValue and len(results) > nb_values:
                     cont = results[nb_values + 1].ServerTimestamp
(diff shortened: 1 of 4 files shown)


oroulet:
>  How to decide when to make a new conn object accounting for thread that is making SQL transactions?

Everything is called by same thread, so no threading issue here

> testing

run all tests: 
     python tests/tests.py

run a specific test (example, could for debugging): 
    python3 tests/tests.py TestServer.test_history_read

> storing variants

The quick and dirty hack is to store everything as binary: Variant.to_binary()
but this will hard to debug
you should anyway have two columns:
*   one for type: save Variant.VarianType.name
*   one for value: there you anyway will have to do some testing for type before read and write

opcua/server/history.py
```diff
@@ -4,6 +4,8 @@
 from opcua import Subscription
 from opcua import ua
 
+import sqlite3
+
```
oroulet: Can you move your sql backend to history_sql.py? so this file does not get too long, and we make iteven more  clear people can write their own.

opcua/server/history.py
```diff
@@ -61,6 +63,13 @@ def read_event_history(self, start, end, evfilter):
         """
         raise NotImplementedError
 
+    def stop(self):
+        """
+        Called when the server shuts down
+        Can be used to close database connections etc.
+        """
+        raise NotImplementedError
+
```
oroulet: OK

oroulet:
Sorry, I was wrong, calls from server side happen in two different threads. So the easiest thing to do is to only have omne connection object and use a Lock from threading.
write: "with self._lock:" at the beginnig og every interface method.

opcua/server/history.py
```diff
@@ -108,20 +117,191 @@ def read_node_history(self, node, start, end, nb_values):
...
+        self._conn_new = None
+        self._c_new = None
+
+        # subscriptions are in another thread so it needs it's own sqlite connection object
+        self._conn_sub = None
+        self._c_sub = None
+
+        # FIXME: no idea what thread the read values happen in, just make a new conn object for now
+        self._conn_read = None
+        self._c_read = None
```
oroulet: no nead for several connections, use a lock

opcua/server/history.py
```diff
@@ -108,20 +117,191 @@ def read_node_history(self, node, start, end, nb_values):
...
+
+        sql_type = self._get_sql_type(node)
+
+        # create a table for the node which will store attributes of the DataValue object
+        try:
+            self._c_new.execute('CREATE TABLE "{tn}" (ServerTimestamp TIMESTAMP,'
+                               ' SourceTimestamp TIMESTAMP,'
+                               ' StatusCode INTEGER,'
+                               ' Value {type},'
+                               ' VariantType INTEGER)'.format(tn=node_id, type=sql_type))
```
oroulet: as writen ealier, use a column for type

opcua/server/history.py
```diff
@@ -108,20 +117,191 @@ def read_node_history(self, node, start, end, nb_values):
...
+        node_type = node.get_data_type()
+        # see object_ids.py
+        if node_type.Identifier in (10, 11,):  # Float, Double
+            return 'REAL'
+        elif node_type.Identifier in (4, 5, 6, 7, 8, 9,):  # Ints
+            return "INT"
+        elif node_type.Identifier in (12,):  # String
+            return "TEXT"
+        else:
+            return "NULL"
```
oroulet: use the VariantType attribute from variant, data_type form Node is redundant and can be wrong

opcua/server/history.py
```diff
@@ -130,7 +310,7 @@ def event_notification(self, event):
 class HistoryManager(object):
     def __init__(self, iserver):
         self.iserver = iserver
-        self.storage = HistoryDict()
+        self.storage = HistorySQLite()  # HistoryDict() HistorySQLite()
```
oroulet: Do not overwrite the default history backend. sqlite will create a a file and we do not want to do that by default.
Create a new server_history_sql.py example

oroulet: and new tests!!!!

opcua/server/history.py
zerox1212: I have never used locks and I don't know how they work. Even with a lock I think the SQL module will complain. Anyways I will try to do this.

opcua/server/history.py
zerox1212: I have a column for value and the variant type.
Do you mean I should store the value as a string, and then have a column for the variant name (such as Double), and then a third column for the variant type integer as described in the OPC UA spec?

opcua/server/history.py
zerox1212: Sorry that was left over from my testing and I didn't know how you wanted to organize other history interfaces.

Sorry again, I'm bad at programming. I will try to figure out how to make some tests for this.

zerox1212:
Thanks for the feedback. Looks like I have made a lot of mistakes. I will try to fix in the next few days.

opcua/server/history.py
oroulet: sorry my fault. you did it correctly. when reading just reconstruct Variant using ua.VariantType(myint)

opcua/server/history.py
zerox1212: I found a solution for this without using locks.

opcua/server/history.py
zerox1212: Can you tell me how to properly use Variant.to_binary() and Variant.from_binary(data)?

When I write binary data to the database it seems to work. However when I read the databack and call Variant.from_binary(data) it crashes on `encoding = ord(data.read(1))`. What is "data" supposed to be?

opcua/server/history.py
zerox1212: How should I do the import if want to use SQL? I can't add the import statement at the top because then HistoryStorageInterface isn't defined with my SQL class is imported?

zerox1212:
I put in as many fixes as I could. I will make a new pull request for you to check.

PR #156, opened 2016-04-11
Updated SQLite Historizing

zerox1212:
I tried to fix everything you told me about in https://github.com/FreeOpcUa/python-opcua/pull/155

I don't know how you want to organize the code. I also don't know how the user should choose which history method, for now I just commented out the SQLite import.

Perhaps you can just take history_sql.py into the main branch and then reorganize the code how you see fit.


opcua/server/history.py
```diff
@@ -90,16 +37,17 @@ def save_node_value(self, node, datavalue):
             data = data[-count:]
 
     def read_node_history(self, node, start, end, nb_values):
+        node_id = node.NodeId
```
destogl: Is this spelled wrongly? Since type `Node` has `nodeid` attribute?

opcua/server/history.py
```diff
@@ -4,83 +4,30 @@
...
         self._datachanges_period = {}
         self._events = {}
 
     def new_historized_node(self, node, period, count=0):
-        self._datachanges[node] = []
-        self._datachanges_period[node] = period, count
-
-    def new_historized_event(self, event, period):
-        self._events = []
+        node_id = node.nodeid
```
destogl: I would remove this line since IMHO is not important it also doesn't really remove complexity

opcua/server/history.py
```diff
@@ -4,83 +4,30 @@
...
+        self._datachanges[node_id] = []
+        self._datachanges_period[node_id] = period, count
 
     def save_node_value(self, node, datavalue):
-        data = self._datachanges[node]
-        period, count = self._datachanges_period[node]
+        node_id = node.nodeid
```
destogl: The same as line 23

destogl:
For me it looks good. I would just put  `history_interface.py`  file in  `history.py` since this would be more Pythonic way. `history_sql.py` should stay since it needs additional module etc...

Anyway, nice work!

oroulet:
> I would just put history_interface.py file in history.py
agree. Can you do this @zerox1212 ?

opcua/server/history.py
oroulet: yes it is very strange.... I will  pull that request and check these things

opcua/server/history.py
oroulet: agree. Can you do this @zerox1212 ?

opcua/server/history.py
zerox1212: It's because in that part you don't get a UA Node. You get a ua.HistoryReadValueId object. In that object node id is in camel case.

Node object spelling:
image

HistoryReadValueId spelling:
image

zerox1212:
I can move the interface back to history.py, however how can I import my `HistorySQlite` class from history_sql.py then?

The only way it works is if `from opcua.server.history_sql import HistorySQLite` is after the class definition of the interface. I thought all import statements should be at the top?

opcua/server/history_sql.py
```diff
@@ -0,0 +1,168 @@
...
+            return 'Q'
+        elif variant_code is 10:  # Float
+            return 'f'
+        elif variant_code is 11:  # Double
+            return 'd'
+        elif variant_code in (12,):  # String
+            return "s"
+        else:
+            # FIXME: Should raise exception here that historizing of node type isn't supported in SQL storage interface
+            return None
```
zerox1212: Do you guys have a recommendation on how to handle the case where the node being historized can't be packed into a binary for SQL?

opcua/server/history.py
destogl: Can you maybe than put different name of the parameter to be clearer

opcua/server/history_sql.py
destogl: You should here check the constants not the hard-coded number, I would write e.g.:

```
...
variant_type is VariantType.Boolean:
     return '?'
...
```

By the way, there is solution to do this with [dictionary](https://bytebaker.com/2008/11/03/switch-case-statement-in-python/) which could be possibly cleaner since you have constants... @oroulet: which way do you prefere?

I have no proposal for your problem sorry....

opcua/server/history.py
oroulet: Why do you pass the entire object if you only use the nodeid?

destogl:
> I can move the interface back to history.py, however how can I import my HistorySQlite class from history_sql.py then?

> The only way it works is if from opcua.server.history_sql import HistorySQLite is after the class definition of the interface. I thought all import statements should be at the top? HistorSQLite can't import the interface class before it's declared.

this is all correct you wrote, but I don't understand your problem, since you just move `HistoryStorageInterface` class and all imports stay the same.... or am I missing something?

opcua/server/history_sql.py
```diff
@@ -0,0 +1,165 @@
...
+        try:
+            _c_new.execute('CREATE TABLE "{tn}" (ServerTimestamp TIMESTAMP,'
+                           ' SourceTimestamp TIMESTAMP,'
+                           ' StatusCode INTEGER,'
+                           ' Value TEXT,'
+                           ' VariantType INTEGER,'
+                           ' ValueBinary BLOB)'.format(tn=node_id))
+
+        except sqlite3.Error as e:
+            print(node_id, 'Historizing SQL Table Creation Error:', e)
```
oroulet: add a logger object: self.logger = logging.getLogger(__name__) (in __init__)
then
self.logger.error("Creating table failed %", ex)

zerox1212:
See my example:
history.py
```python
from opcua.server.history_sql import HistorySQLite

class HistoryStorageInterface(object):
    code....
```

history_sql.py
```python
from opcua.server.history import HistoryStorageInterface

class HistoryStorageSQL(HistoryStorageInterface)
    code...
```

This won't work because HistorySQLite will be imported before the class HistoryStorageInterface is declared.

opcua/server/history_sql.py
zerox1212: Yes I wanted to ask you about this. Should there be a logger per backend? Or should there be a "history" logger that all backends use?

opcua/server/history_sql.py
```diff
@@ -0,0 +1,165 @@
...
+            return "s"
+        else:
+            # FIXME: Should raise exception here that historizing of node type isn't supported in SQL storage interface
+            return None
+
+    def _get_table_name(self, node):
+        if type(node) is Node:
+            return str(node.nodeid.NamespaceIndex) + '_' + str(node.nodeid.Identifier)
+        if type(node) is ua.HistoryReadValueId:
+            return str(node.NodeId.NamespaceIndex) + '_' + str(node.NodeId.Identifier)
```
oroulet: we may have an issue here. Identifier can be very strange things such as guid... but we can skip that issue for now...

opcua/server/history_sql.py
```diff
@@ -0,0 +1,165 @@
...
+
+        value_blob = self._pack_value(datavalue)
+
+        # insert the data change into the database
+        try:
+            _c_sub.execute('INSERT INTO "{tn}" VALUES (?, ?, ?, ?, ?, ?)'.format(tn=node_id), (datavalue.ServerTimestamp,
+                                                                                               datavalue.SourceTimestamp,
+                                                                                               datavalue.StatusCode.value,
+                                                                                               str(datavalue.Value.Value),
+                                                                                               datavalue.Value.VariantType.value,
```
oroulet: what about saving VariantType.name instead of value? I think that it does not impact much performance and make things more readable. We can always change it later. just a suggestion...

opcua/server/history_sql.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def save_event(self, event):
+        raise NotImplementedError
+
+    def read_event_history(self, start, end, evfilter):
+        raise NotImplementedError
+
+    def _pack_value(self, datavalue):
+        variant_code = datavalue.Value.VariantType.value
+        return struct.pack(self._get_pack_type(variant_code), datavalue.Value.Value)
```
oroulet: OK, this is not necessary. If you want to save variant as binary ust called Variant.to_binary() and variant.from_binary(). No need to use pack directly. 

Would be nice to save data as a python type, but we will get issue with extensions objects, etc,,,,

opcua/server/history_sql.py
oroulet: As written in last comment this method is not necessary. remove it and use to_binary()

opcua/server/history_interface.py
```diff
@@ -0,0 +1,62 @@
...
+        raise NotImplementedError
+
+    def save_node_value(self, node, datavalue):
+        """
+        Called when the value of a historized node has changed and should be saved in history
+        Returns None
+        """
+        raise NotImplementedError
+
+    def read_node_history(self, node, start, end, nb_values):
```
oroulet: why did you change passed type from nodeid to node? it does not look like you are using it anywhere

oroulet:
Sorry for all the comments, I know it is hard to understand the entire UA sprotocol and this particular stack

opcua/server/history_sql.py
zerox1212: I will do it if I don't need to convert it to a value manually...

Can I pass a string to VariantType()? I am not sure how Enums work. I assumed I needed to pass a value to it.

opcua/server/history_sql.py
zerox1212: I tried this several times. I couldn't get it to work because `Variant.from_binary()` expects a buffer. SQL returns a byte array.

opcua/server/history_interface.py
zerox1212: Sorry I forgot to change that back. Before I was using the node name for table naming. Now I only use the nodeid so I will change it back.

zerox1212:
I appreciate all the comments. I try my best to help you guys, but it seems like the amount of time you spend looking at my could you could of written it yourself. Sorry for that.

oroulet:
It is faster to comment than to write :-) and then you learn and maybe
write more  code later !

opcua/server/history_sql.py
oroulet: Make your own for your SQL backbend

opcua/server/history_sql.py
oroulet: Wrap your byte array with opcua.utils.Buffer then it will work

opcua/server/history.py
```diff
@@ -4,7 +4,70 @@
...
+    def stop(self):
+        """
+        Called when the server shuts down
+        Can be used to close database connections etc.
+        """
+        raise NotImplementedError
+
 
 # if you want to use an SQL based history uncomment this import and change the storage type of the history manager
 # from opcua.server.history_sql import HistorySQLite
```
zerox1212: @destogl This is why I moved the interface to it's own file. I don't know the correct way to import the HistorySQLite backend, but you can't do it at the top of history.py.

zerox1212:
I did all the corrections you spoke of except using `Variant.to_binary()` and `Variant.from_binary()`. This because I don't know how to get `from_binary() to work as it requires a buffer object. SQL doesn't return that when I SELECT it.

Please check it, thanks.

oroulet:
To use from_binary from a string: xx.from_binary(opcua.utils.Buffer(mystring))

zerox1212:
I will give that a try and test it if it works.

I also saw some people using `sqlite3.Binary(blob)` when inserting into the database so I need to look into what that does.

zerox1212:
I implemented `to_binary()` and `from_binary()` as you mentioned and it is working. Thanks for that.

destogl:
> I appreciate all the comments. I try my best to help you guys because I really like this project, but it seems like the amount of time you spend looking at my code you could of written it yourself. Sorry for that.

Don't worry about that, all of us need feedback about the code, this is what make the code nice. You are welcome to comment my PRs as well :)

> See my example: ...

Thanks for example now I understand the problem, sorry I missed that. But I am not sure that this is the best way for doing that, so to say to import `HistoryStorageSQL` from `history.py`. I would propose the following:

* The `history_sql.py ` should be imported from server implementation and not from `history.py` file. So we have always dict history as default and if user wants to use sql than set it by himself so in the example file `server-minimal-history-sql` i would write following:

```
from opcua.server.history_sql import HistorySQLite
...
server = Server()
server.history_manager.set_storage(HistorySQLite())
...
```

@oroulet: should we have option to enable/disable history? I can imagine use cases where user doesn't want to use history...

zerox1212:
Having a set_storage() method works for me. If we could clean up these last points and commit then I can move forward and take a look at historizing events.

Thanks everyone.

oroulet:
@zerox1212 look at #158 . I moved your work there. and I close this one

PR #158, opened 2016-04-12
SQL History storage

oroulet:
cleanup and merge work of @zerox1212


oroulet:
@zerox1212  I added tests in tests/tests_history.py and of course they catch some errors. 
Can you fix the issues there?
running:

     python tests/tests.py TestHistorySQL.test_history_read

should run as good as 

     python tests/tests.py TestHistory.test_history_read

zerox1212:
I will check the tests.

Also, I don't know if you really need to add `with._lock()` to the methods. I found out that python's SQLite3 implementation should be compiled in thread mode 1(?) This means calls to SQLite from multiple threads get serialized.

If you use a lock, I think that an SQL error could potentially block execution for a long time (there is a timeout parameter).

zerox1212:
The test_history.py is missing from the repository.

oroulet:
Fixed

oroulet:
No_thread_check is a rather suspect option name. I prefer to have a lock.

zerox1212:
It's just a misleading name by the python implementation.
https://www.sqlite.org/threadsafe.html

When I try:
```python
import sqlite3
sqlite3.threadsafety```
It returns mode 1.

From what I read, older versions of SQLite were NOT threadsafe. In SQLite3 they changed this, but the python build of sqlite still warns you about using SQLite from multiple threads in case you use an older version (or perhaps for very old versions of Python).

I don't think a lock is needed personally. SQL shouldn't be editing data that could be selected so I think the risk that SQL screws up and locks up the thread is worse than history returning something weird. However I'm not an expert on this.

**Comment by zerox1212** [comment]
When you moved all the code to the other branch it made my fork out of sync...

I fixed history_sql.py failing tests. You can get the code from here. No point in making a pull request.
https://github.com/zerox1212/python-opcua/blob/master/opcua/server/history_sql.py

If you want to check the commit, it was failing tests because of this:
```python
 if period:
                # after the insert, if a period was specified delete all records older than period
                date_limit = datetime.now() - period
```

I wasn't checking for period to be None.
https://github.com/zerox1212/python-opcua/commit/f67adb53c00a08a59bc86df068cdb810c379d6ca

oroulet:
@zerox1212 

    When you moved all the code to the other branch it made my fork out of sync...

yes, sorry for that

but your fix is not enough, the test still fails, if no period is given by users then max num results should be returned. Do you have the possibility to fix this? and make tests pass? I haven't used sql for a while...

zerox1212:
When I run TestHistorySQL.test_history_read and TestHistory.test_history_read I get the exact same results: 1 Failure.

oroulet:
What error do you get when running history test without sql? I get no error here with  python tests/tests.py TestHistory.test_history_read

Anyway I see in your code that the count parameter is ignored while it should be used to return max 'count' number and return a continuation point if number of result is > count. See the HistoryDict implementation and the explanation in the interface class

zerox1212:
Normal Test

> C:\.....\python-opcua-sql2>python tests/tests.py TestHistory.test_history_read
> cryptography is not installed, use of crypto disabled
> cryptography is not installed, use of crypto disabled
> WARNING: CRYPTO NOT AVAILABLE, CRYPTO TESTS DISABLED!!
> C:\Python34\lib\asyncio\tasks.py:522: DeprecationWarning: asyncio.async() function is deprecated, use ensure_future()
>   DeprecationWarning)
> Listening on localhost:48530
> test_history_read (tests_history.TestHistory) ... FAIL
> Cleanup client connection:  ('127.0.0.1', 56715)
> 
> ======================================================================
> FAIL: test_history_read (tests_history.TestHistory)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "C:\Users\d15138\PycharmProjects\python-opcua-sql2\tests\tests_history.py", line 35, in test_history_read
>     self.assertEqual(len(res), 2)
> AssertionError: 0 != 2
> 
> ----------------------------------------------------------------------
> Ran 1 test in 2.598s
> 
> FAILED (failures=1)

SQL Test

> C:\...\python-opcua-sql2>python tests/tests.py TestHistorySQL.test_history_read
> cryptography is not installed, use of crypto disabled
> cryptography is not installed, use of crypto disabled
> WARNING: CRYPTO NOT AVAILABLE, CRYPTO TESTS DISABLED!!
> C:\Python34\lib\asyncio\tasks.py:522: DeprecationWarning: asyncio.async() function is deprecated, use ensure_future()
>   DeprecationWarning)
> Listening on localhost:48530
> test_history_read (tests_history.TestHistorySQL) ... FAIL
> Cleanup client connection:  ('127.0.0.1', 56736)
> 
> ======================================================================
> FAIL: test_history_read (tests_history.TestHistorySQL)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "C:\Users\d15138\PycharmProjects\python-opcua-sql2\tests\tests_history.py", line 36, in test_history_read
>     self.assertEqual(res[-1].Value.Value, vals[-1])
> AssertionError: 1 != 19
> 
> ----------------------------------------------------------------------
> Ran 1 test in 2.608s
> 
> FAILED (failures=1)

zerox1212:
I see that I forgot to make a query for "count" based results. I will add that.

To me returning history by count seems pointless when you can do it by date + limit, but if it's in the spec I will add it.

oroulet:
>> To me returning history by count seems pointless 

Yes the spec has a bunch of strange things, but limiting by count is not that bad an idea since in many situation you may want a number of results and avoid to get thousand of results back

I just checked on travis (The automated test running system linked to github) and the test without SQL do run fine there:

     test_history_read (tests_history.TestHistory) ... ok

must be something special with your PC...

zerox1212:
I will investigate.

zerox1212:
I need to test this. You can take a look if you want.
https://github.com/zerox1212/python-opcua/commit/20635ef561c35197b162c504dc2490d5a21dccce

opcua/server/history.py
```diff
@@ -89,17 +97,17 @@ def save_node_value(self, node, datavalue):
...
         else:
             if end is None:
                 end = datetime.now() + timedelta(days=1)
             if start is None:
                 start = ua.DateTimeMinValue
-            results = [dv for dv in self._datachanges[node] if start <= dv.ServerTimestamp <= end]
+            results = [dv for dv in self._datachanges[node_id] if start <= dv.ServerTimestamp <= end]
             if nb_values:
                 if start > ua.DateTimeMinValue and len(results) > nb_values:
                     cont = results[nb_values + 1].ServerTimestamp
```
zerox1212: This line needs to be fixed. It should be:
```python
cont = results[nb_values + 1].ServerTimestamp
```
Otherwise 1 array member will be skipped due to the list starting at 0.

zerox1212: Results:
image
Cont:
image

zerox1212:
Can you grab the latest history_sql.py? https://github.com/zerox1212/python-opcua/commit/4f3082796765d0c0944de7919c4b922e8267f243

Also, when running tests on SQL. We need to make sure history.db is deleted before running tests. Otherwise it will persist and have old data. This could cause problems with tests (such as if you add 20 datavalues, then test if history_read returns 20 items).

zerox1212:
Edit, don't bother with the above commit. Tests still fail and at the moment I have no idea why. The exact same SQL query works in my SQL testing tool, but in python it returns nothing.

zerox1212:
OK... are you positive that test_history.py is the one you are using? And history.py in this branch is the latest?

1. The first test will never pass with the code I have. 
```python
        # test a read with: no start, end time of now, and number of values 2
        res = var.read_raw_history(None, now, 2)
        self.assertEqual(len(res), 2)
        self.assertEqual(res[-1].Value.Value, vals[-1])
```
Both my SQL and your Dict grab the two OLDEST values in the results. This test expects the opposite.

2. In the test the "end" time seems to cause my SQL to find no data due to differences in timezone.
```python
now = datetime.now()
```
In the real implentation we have this:
```python
if end is None:
                end = datetime.now() + timedelta(days=1)
```

oroulet:
Hi @zerox1212 ,
I split and added more tests and found one bug on HistoryDict. I also replaced all datetime.now with datetime.utcnow. I do not think it will fix your issues, but maybe it will make ti clearer to you where your issue is. run: 

```
python tests/tests.py TestHistory

test_history_read_2_with_end (tests_history.TestHistory) ... ok
test_history_read_5_in_timeframe (tests_history.TestHistory) ... ok
test_history_read_all (tests_history.TestHistory) ... ok
test_history_read_all2 (tests_history.TestHistory) ... ok
test_history_read_last_3 (tests_history.TestHistory) ... ok
test_history_read_none (tests_history.TestHistory) ... ok
test_history_read_one (tests_history.TestHistory) ... ok
```


Then
```
python tests/tests.py TestHistorySQL
test_history_read_2_with_end (tests_history.TestHistorySQL) ... FAIL
test_history_read_5_in_timeframe (tests_history.TestHistorySQL) ... ok
test_history_read_all (tests_history.TestHistorySQL) ... ok
test_history_read_all2 (tests_history.TestHistorySQL) ... FAIL
test_history_read_last_3 (tests_history.TestHistorySQL) ... FAIL
test_history_read_none (tests_history.TestHistorySQL) ... FAIL
test_history_read_one (tests_history.TestHistorySQL) ... FAIL
```

zerox1212:
Using UTC time fixed a few of them.

I see the other problems as well. I didn't realize that if there are no start or end times that your history selects EVERY history item in the dict. Seems like it could be rather inefficient... I will work on fixing SQL right away. Thanks for making tests more clear for me.

oroulet:
>>  I didn't realize that if there are no start or end times that your history selects EVERY history item in the dict

I tried to follow the spec, this is what I understood from spec and why there is the nb_values parameter which limits max number of points to return. Would be nice to use sql as much as possible for this limits and not python, but I let you take care of that

zerox1212:
I just thought if you wanted to select the last X values it would be better to do `reverse()` or something on the list. Since I didn't see this I didn't know that nb_values should select the MOST RECENT values.

zerox1212:
Can you tell me where you find this in the spec?

The reason I ask is this:
If there is a valid time frame (start, end) AND nb_values, you return the oldest values. If ONLY nb_values is supplied, you return the most recent values. This seems contradictory.

zerox1212:
OK... please take a look at my fork to see what you think. https://github.com/zerox1212/python-opcua/commits/master

I changed your tests to be consistent. Anytime no start or end time is supplied results should be returned in REVERSE order (even though this is not in OPC UA spec). If only END time is supplied results should be in reverse order, as described in spec. I also added a test for the case where start > end time, in which spec says return results in reverse order.

Of course this makes your HistoryDict now fail tests. It's up to you what you want to do.

I attached the spec pages I looked at.
image

oroulet:
you are right, trying to pull your changes now

oroulet:
merged your work, still need to go through and fix HistoryDict

oroulet:
now it should be OK! 
Thank you @zerox1212 for the help!
have a look and if I did not miss anything, I will merge

oroulet:
merging anyway, send PR if I missed something

zerox1212:
Looks good. Thanks for the help.

Issue #159, opened 2016-04-19
Autogenerate the UATypes


destogl:
Following the discussion in #133 I saw that we have problem with not to generate UA Types (standard types) as python objects. This can be very useful to achieve good mapping between UA Types and internal objects. So my proposal would be following:

* For every UA Type we should have mapping to Python types/objects (this is already achieved in `uatypes.py`
* Every UA Type could extend Python type/object to have all properties as python type/object, e.g. on `BaseEvent` UA Object

   `BaseEvent` (ua) should inherit `Event` (python) which than has `fill_base_fileds`-mehtod.

In this way `BaseEvent` class can be autogenerated  and the same is true for all other (inheriting) events...

* Similar approach is also interesting for other types/object like base types or Node/NodeId...

This is just my two cents regarding this issue.... hopefully at one point in time I provide also some inputs regarding that.

oroulet:
OK let's try, if you make it we can merge your event stuff as is :-)
But generate the objects in their own files! and maybe only start with events. And I also propose to create a new python file to parse xml and generate types so we do not make the generate_address_space.py too complicated. We need to do something completely different...

oroulet:
OK I wrote some code for it. I will show it soon BUT I am sure this is not enough for events.
Typically a server does not generate BaseEvents it generate custom events from its custom datamodell. You want to let user to create new events in python something like

```
basevent = server,get_node(ua.ObjectIds.BaseEventType)
myeventtype = basevent.add_object("MyNewEvent")
myeventtype.add_property("MySuperProperty", 9,8)
myeventtype.add_property("MySuperOtherProperty", 9,8)

# now use my event
generator = server.get_event_generator(myeventtype)
generator.MySuperProperty= 4
generator.trigger()
```

And on client side
```
customeventype = client.get_node(NodeIDOfEvent)
sub = client.create_subscription()
sub.subscribe_events(customeventtype)
```

destogl:
I will just write one small proposal...

For custom events I agree with you, lets do this. Even though I am more fun of writing new event class with all properties.... But actually we can enable booth of this things... Just give me a second :)

PR #161, opened 2016-04-20
Events implementation proposal

destogl:
Follwing the discussion in #133 and #159 I created new proposal, which is mix of original implementation and objects based implementation. @oroulet please give your opinion on following since I am starting to doubt on my object based idea (I generates additional code which is maybe not important at all. (This is just proposal so the test will probably not pass)

1. Object based proposal:
Object based proposal provides to freeze standard event types and better control over automatic filling of fields. We could implement an OPCUA-Object which takes care that every one of the UA objects (standard or custom) are provided into address space (something for #159).

@oroulet: Is this important in your opinion, does object based proposal provides some added value?

2. Generator-only based proposal:
The events and objects are generated from addresspace. The fields are automatically filed based on ObjectIds. We have Less code and very high flexibility. Users can potentially change standard event types unintentionally.

@oroulet: This is your original implementation

I will change this proposal according to the choice we make and provide first working version (at least for BaseEventType)

### DIFF
--- a/opcua/common/event.py
+++ b/opcua/common/event.py
@@ -22,36 +22,32 @@ class EventGenerator(object):
         etype: The event type, either an objectId, a NodeId or a Node object
     """
 
-    def __init__(self, isession, etype=ua.ObjectIds.BaseEventType, source=ua.ObjectIds.Server):
+    def __init__(self, isession, etype=ua.BaseEvent(), source=ua.ObjectIds.Server):
         self.isession = isession
-        self.node = None
-        self.event = etype
+        self.event = None
+        node = None
 
         if isinstance(etype, ua.BaseEvent):
-            pass
+            self.event = etype
         elif isinstance(etype, Node):
-            self.node = etype
+            node = etype
         elif isinstance(etype, ua.NodeId):
-            self.node = Node(self.isession, etype)
+            node = Node(self.isession, etype)
         else:
-            self.node = Node(self.isession, ua.NodeId(etype))
+            node = Node(self.isession, ua.NodeId(etype))
 
-        if self.node:
-            event = CustomEvent()
-            references = node.get_children_descriptions(refs=ua.ObjectIds.HasProperty)
-            for desc in references:
-                node = Node(self.isession, desc.NodeId)
-                setattr(self, desc.BrowseName.Name, node.get_value())
+        if node:
+            self.event = get_event_from_node(node)
 
         if isinstance(source, Node):
             pass
-        elif isinstance(source, NodeId):
-            source = ua.Node(isession, source)
+        elif isinstance(source, ua.NodeId):
+            source = Node(isession, source)
         else:
             source = Node(isession, ua.NodeId(source))
 
         if self.event.SourceNode.Identifier:
-            source = Node(self.iserver.isession, self.event.SourceNode)
+            source = Node(self.isession, self.event.SourceNode)
 
         self.event.SourceName = source.get_display_name().Text
         source.set_attribute(ua.AttributeIds.EventNotifier, ua.DataValue(ua.Variant(1, ua.VariantType.Byte)))
@@ -69,9 +65,9 @@ class EventGenerator(object):
             self.event.Time = time
         else:
             self.event.Time = datetime.utcnow()
-        self.event.ReciveTime = datetime.utcnow()
+        self.event.RecieveTime = datetime.utcnow()
         #FIXME: LocalTime is wrong but currently know better. For description s. Part 5 page 18
-        self.event.LocaleTime = datetime.utcnow()
+        self.event.LocalTime = datetime.utcnow()
         if message:
             self.Message = ua.LocalizedText(message)
         elif not self.event.Message:
@@ -79,15 +75,36 @@ class EventGenerator(object):
         self.isession.subscription_service.trigger_event(self.event)
 
 
+def get_event_from_node(node):
+    event = None
+    if node.nodeid.Identifier in ua.uaevents_auto.IMPLEMNTED_EVENTS.keys():
+        event = ua.uaevents_auto.IMPLEMNTED_EVENTS[node.nodeid.Identifier]()
+
+    else:
+        pass
+        #node.get
+
+        #class CustomEvent():
+
+            #pass
+    #references = node.get_children_descriptions(refs=ua.ObjectIds.HasProperty)
+    #for desc in references:
+        #child = Node(self.isession, desc.NodeId)
+        #setattr(self.event, desc.BrowseName.Name, child.get_value())
+
+    return event
+
+
 class CustomEvent(ua.BaseEvent):
 
-    def __init__(self, etype=ua.ObjectIds.BaseEventType, sourcenode=ua.NodeId(ua.ObjectIds.Server), message=None, severity=1):
-        super(ua.BaseEvent, self).__init__(sourcenode, message, severity, True)
+    def __init__(self, etype=ua.BaseEvent, sourcenode=ua.NodeId(ua.ObjectIds.Server), message=None, severity=1):
+        super(CustomEvent, self).__init__(sourcenode, message, severity, True)
         #TODO: Add fileds
 
     #TODO: Extend to show all fields of CustomEvent
     def __str__(self):
-        s = 'CustomEvent(EventId:{}'.format(self.EventId)
+        s = 'CustomEvent('
+        s += 'EventId:{}'.format(self.EventId)
         s += ', EventType:{}'.format(self.EventType)
         s += ', SourceNode:{}'.format(self.SourceNode)
         s += ', SourceName:{}'.format(self.SourceName)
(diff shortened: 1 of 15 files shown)


oroulet:
This looks OK for me.

Concerning the UA object generation, it might be interesting. Maybe we should first merge this PR then try to generate UA objects and see how far we get?

destogl:
Thanks for the feedback!

destogl:
I need to write few more tests and example, but all in all this should be it.

This depends on #165 which is already merged (thanks!)

opcua/common/event.py
```diff
@@ -22,53 +22,140 @@ class Event(object):
...
+        return None
+    if parents[0].nodeid.Identifier in ua.uaevents_auto.IMPLEMNTED_EVENTS.keys():
+        return node.nodeid.Identifier, ua.uaevents_auto.IMPLEMNTED_EVENTS[parents[0].nodeid.Identifier]
+    else:
+        _find_parent_eventtype(parents[0])
+
+
+
+
+class CustomEvent(ua.BaseEvent):
```
oroulet: why 2 definition of CustomEvent?

opcua/server/address_space.py
```diff
@@ -226,8 +232,13 @@ def _add_node(self, item, user):
...
 
-        # now add our node to db
-        self._aspace[nodedata.nodeid] = nodedata
+            addref = ua.AddReferencesItem()
+            addref.ReferenceTypeId = item.ReferenceTypeId
+            addref.SourceNodeId = nodedata.nodeid
+            addref.TargetNodeId = item.ParentNodeId
+            addref.TargetNodeClass = self._aspace[item.ParentNodeId].attributes[ua.AttributeIds.NodeClass].value.Value.Value
+            addref.IsForward = False
+            self._add_reference(addref, user)
```
oroulet: this is already merged. The diff is outdated. Maybe it get fixed if you rebase over master?

opcua/common/event.py
destogl: Uhhh, the second should not exist any more.... This is not completely cleaned version... I will clean it with other stuff, sorry :/

opcua/server/address_space.py
destogl: Yes I know I will do this...

destogl:
I rebased it. Now I have problems in test with history. Could you please look at it did I introduced it or some other commit?

destogl:
Oh? It works? Coll, at my computer I had problems with History :/

zerox1212:
I will try to test this myself. It looks like a big change, we should probably up the version for this.

As soon as this is tested and committed I will move forward with events history (and I might need some help because it looks like rebuilding an event from history will be different than before).

opcua/common/event.py
```diff
@@ -22,53 +22,113 @@ class Event(object):
...
+        self.event.LocalTime = datetime.utcnow()
+        if message:
+            self.Message = ua.LocalizedText(message)
+        elif not self.event.Message:
+            self.event.Message = ua.LocalizedText(self.source.get_browse_name().Text)
+        self.isession.subscription_service.trigger_event(self.event)
+
+
+def get_event_from_node(node):
+    if node.nodeid.Identifier in ua.uaevents_auto.IMPLEMNTED_EVENTS.keys():
```
oroulet: spelling error here

opcua/common/event.py
```diff
@@ -22,53 +22,113 @@ class Event(object):
...
+        return ua.uaevents_auto.IMPLEMNTED_EVENTS[node.nodeid.Identifier]()
+    else:
+        parent_identifier, parent_eventtype = _find_parent_eventtype(node)
+        if not parent_eventtype:
+            return None
+
+        class CustomEvent(parent_eventtype):
```
oroulet: We do not need to have this as an internal class. We can as well call setattr after creating the object, or give everything necessary as arg to __init__ og CustomEvent

destogl: This is creating an event from EventType in the namespace. So I added internal class, so that user can not do what ever he wants, but he needs first define EventType object in the namespace and than can create CustomEvent object...

If user add attributes after than is hard to track compatibility witht he EventType in the namespace, and theoretically clients can have problem than...

But if you think, this is not important I can change it...

But actually the main reason to do it as internal class, is that you can define arbitrary parent class, so in this way your object can have as parent any EventType, not just BaseEventType.

tests/tests_common.py
```diff
@@ -289,53 +288,29 @@ def test_subscribe_events_to_wrong_node(self):
...
+        #ev = msclt.future.result()
+        #self.assertIsNot(ev, None)  # we did not receive event
+        #self.assertEqual(ev.SourceNode, self.opc.get_server_node().nodeid)
+        #self.assertEqual(ev.Message.Text, msg)
+        ##self.assertEqual(msclt.ev.Time, tid)
+        #self.assertEqual(ev.Severity, 500)
+
+        ## time.sleep(0.1)
+        #sub.unsubscribe(handle)
+        #sub.delete()
```
oroulet: one way or another yo uneed to reenable client side tests. the server object is always available as sefl.srv to setup and generate events

destogl: Yes I will do it. I am currently adding more tests and debugging. This was just one intermediate commit. My computer was strange today, so I push it as back up, and you can see the progress :)

destogl:
So this would be it. I added also tests for most of the Use cases and updated examples.

opcua/common/subscription.py
```diff
@@ -207,11 +202,11 @@ def _get_node(self, nodeid):
     def _get_filter_from_event_type(self, eventtype):
         eventtype = self._get_node(eventtype)
```
destogl: Off-topic: This `_get_node` method is very useful and in a lot of cases we implement the same part of the code. Maybe we should think about creating one `common/tools.py` where we will put all this methods useful from different places.

zerox1212:
I grabbed your event proposal branch and took a look. Here are a few things I noticed:
1. Declaring a custom event (MyEvent in the example) sets the property type, but the actual event instance ends up as NoneType.
```python
myevgen = server.get_event_generator(etype, myobj)
```
I'm not sure if this is intended, but I was expecting to see the properties of myevgen.event match the custom type.

2. Should MyObj have a reference to the event generator? I haven't read the spec on this, but I was a bit surprised I didn't see anything here:
image

3. This is probably my fault, but I can't figure out how to view the values of the event properties in my test client (UA Expert). Have you seen the fired event show the custom fields an a client other that python opcua?

Other than that it looks good. Events seem a bit cumbersome to use, but at this point I can't think of a better way.

destogl:
> Declaring a custom event (MyEvent in the example) sets the property type, but the actual event instance ends up as NoneType. I'm not sure if this is intended, but I was expecting to see the properties of myevgen.event match the custom type i passed it in the call myevgen = server.get_event_generator(etype, myobj).

I don't see it? Or did I understand it wrongly?

bildschirmfoto12

> Should MyObj have a reference to the event generator? I haven't read the spec on this, but I was a bit surprised I didn't see anything here (such as a reference to type MyEvent?)

This could be. I will check it. But the most important part is `EventNotifier` Attribute. It seams it should be done. I will do it :)

> This is probably my fault, but I can't figure out how to view the values of the event properties in my test client (UA Expert). Have you seen the fired event show the custom fields an a client other than python opcua? How can I test this?

I also didn't see them. So I don't konw is this is possible.

opcua/common/subscription.py
oroulet: common/utils.py is the place!

oroulet:
> 
> 
>     This is probably my fault, but I can't figure out how to view the values of the event properties in my test client (UA Expert). Have you seen the fired event show the custom fields an a client other than python opcua? How can I test this?
> 
> I also didn't see them. So I don't konw is this is possible.

I think it is not. limitation of uaexpert. But I have used the feature with another sdk some years ago

oroulet:
> etype = server.create_custom_event_type(2, 'MyEvent', ua.ObjectIds.BaseEventType, [('MyNumericProperty', ua.VariantType.Float), ('MyStringProperty', ua.VariantType.String)])

subjectively I do not really like this method. I think it more understandable to let users create event type by hand:

`parent = server.get_node(ua.ObjectIds.BaseEventType)
custom = parent.add_object(2, "mycustomevent")
custom.add_property(2, "myproperty", 88)

gen = server.get_event_generator(custom)
`

but this PR has been opened far too long so I am merging it :+1:

destogl:
Thanks for merging. I also found some small error, I will add them into new PR.

Regarding your proposal: It is also possible to use is as you proposed, but than the user should take care about adding `ObjectType` not `Object` as NodeType. It is also possible to do it as you propose, but than you need to add it with `add_subtype` to get all references added correctly...

I you like I can add the correct way into `examples/server-event.py`. Because we still have a lot places where we are not really aligned to standard :/ Even though it doesn't matter since for now it works :)

oroulet:
Yes good idea. You should also start from server-minimal.py. it is much
cleaner

PR #162, opened 2016-04-21
Null string instead of string of 0 length for null QualifiedName and LocalizedText

oroulet:


oroulet:
This is what uaexpert does, so it is probably correct

PR #163, opened 2016-04-22
Fix: Use daemon threads which will stop after other threads finished

plieningerweb:
Why?
Because this way, even if there is a programming error, the python process will finally exit. Otherwise, you have to kill it on purpose. 
Furthermore, one has to think about how to handle the client, keep it running and finally maybe exit it.

Contra?
Session.Close() will not be called, but that is also the case for the forced exit.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -34,6 +34,7 @@ class KeepAlive(Thread):
         if timeout == 0:  # means no timeout bu we do not trust such servers
             timeout = 360000
         self.timeout = timeout
+        self.daemon = True
         self.client = client
         self._dostop = False
         self._cond = Condition()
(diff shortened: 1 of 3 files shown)


oroulet:
Doesn't it encourage bad practices like not calling stop() ?

plieningerweb:
Yes, I think it is true, that it would encourage bad practices to some extend

zerox1212:
In my opinion this should not be in the default library. If the user is having issues which don't call stop() they can change to daemon threads them selves like you have.

oroulet:
We will not merge this one. But maybe you can create a method called set_daemon on client and server class that does this. That would be OK.

oroulet:
@plieningerweb ??

PR #165, opened 2016-04-23
Added support for Inverse references

destogl:
Working on #161 I found out that inverse references are missing (from child to parent). So here is initial implementation. The result is cross-compared with server based on Unified Automation SDK and Textual specification (Part 3) is consulted. Following major changes are done:

* Hierarchical references are automatically added (parent <--> child)
* create_method now returns Node (as all other create methods (variable, object))
* create_folder is now merged with create_object since the difference is in only one variable
* On multiple places are written constants instead of hard-coded numeric values
* Tests for references are written and some others are extended

### DIFF
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -108,7 +108,7 @@ class ViewService(object):
         """
         if ref1.Identifier == ref2.Identifier:
             return True
-        if not subtypes and ref2.Identifier == ua.ObjectIds.HasSubtype:
+        elif not subtypes or ref2.Identifier == ua.ObjectIds.HasSubtype:
             return False
         oktypes = self._get_sub_ref(ref1)
         return ref2 in oktypes
@@ -117,7 +117,7 @@ class ViewService(object):
         res = []
         nodedata = self._aspace[ref]
         for ref in nodedata.references:
-            if ref.ReferenceTypeId.Identifier == ua.ObjectIds.HasSubtype:
+            if ref.ReferenceTypeId.Identifier == ua.ObjectIds.HasSubtype and ref.IsForward:
                 res.append(ref.NodeId)
                 res += self._get_sub_ref(ref.NodeId)
         return res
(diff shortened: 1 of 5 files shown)


opcua/server/address_space.py
```diff
@@ -108,7 +108,8 @@ def _suitable_reftype(self, ref1, ref2, subtypes):
         """
         if ref1.Identifier == ref2.Identifier:
             return True
-        if not subtypes and ref2.Identifier == ua.ObjectIds.HasSubtype:
+        #TODO: Please check the changes here
+        elif not subtypes:
```
destogl: @oroulet Please check this line. i change it to get all test to work. I don't understand it at all, for me the if statement before doesn't make any sense :/

oroulet: did the original implementation break something? This is copied from C++ and I kind of remember we had issues here. But removing the last test as you did seems to make sense... if it does not break tests we can try with your version...

tests/tests_common.py
```diff
@@ -856,14 +882,57 @@ def test_method_array2(self):
...
+        self.assertEqual(v.get_type_definition(), ua.ObjectIds.BaseDataVariableType)
+
+        p = o.add_property(3, 'MyProperty', 2)
+        nodes = o.get_referenced_nodes(refs=ua.ObjectIds.HasProperty, direction=ua.BrowseDirection.Forward, includesubtypes=False)
+        self.assertTrue(p in nodes)
+        nodes = p.get_referenced_nodes(refs=ua.ObjectIds.HasProperty, direction=ua.BrowseDirection.Inverse, includesubtypes=False)
+        self.assertTrue(o in nodes)
+        self.assertEqual(p.get_parent(), o)
+        self.assertEqual(p.get_type_definition(), ua.ObjectIds.PropertyType)
+
```
oroulet: you added tests! good reason to trust you new code :-)

oroulet:
looks all good

tests/tests_common.py
destogl: Yes I decided for the future to try to use Test-driven development, since it is easier to get what you want :)

PR #167, opened 2016-04-25
Turn Server into a context manager

Adirio:
Allow using Server with the with statement:

```
server = Server()
# Set up the server
with server as server:
    # Update the server
```


oroulet:
fine

PR #168, opened 2016-04-26
New-style classes

Adirio:
Classes involved:
- EventResult
- SubscriptionItemData
- DataChangeNotif

Unless you are doing some hack that requires old-style classes, they should be new-style ones to ensure forward compatibility.


destogl:
I agree.

PR #170, opened 2016-04-29
Events fixing

destogl:
Fixing the staff commented in #161:

* Added reference to event type (reference type: GeneratesEvent)
* Added in example @oroulet's way for creating custom event types
* Updated tests
* Correct definition of nodes defining new types


destogl:
That was fast :)

oroulet:
When you have new tests and do not break others, then it is much less dangerous to merge :-)

PR #172, opened 2016-05-01
Implement Event History

zerox1212:
Working event history.
No tests yet, sorry. I want to wait until multiple event types for a single object are supported.

Need to decide best way to discover or give history all possible event fields for an Object. I would prefer to give enable_history the Object node only, then history searches references for event nodes and compiles a list of fields for the SQL columns. However I have yet to find a way to accomplish getting the events via references supplied by object node.


destogl:
Please check my comment in #171

examples/server-history.py
```diff
@@ -34,12 +42,17 @@
 
     # enable history for this particular node, must be called after start since it uses subscription
     server.iserver.enable_history(myvar, period=None, count=100)
+
+    # enable history for myobj events generated by myevgen
+    server.iserver.enable_event_history(myobj, myevgen.event, period=None)
```
destogl: Here you are using event object not eventtype. Mybe is better to use eventtype?

opcua/common/event.py
```diff
@@ -83,8 +83,8 @@ def trigger(self, time=None, message=None):
             self.event.Time = time
         else:
             self.event.Time = datetime.utcnow()
-        self.event.RecieveTime = datetime.utcnow()
-        #FIXME: LocalTime is wrong but currently know better. For description s. Part 5 page 18
+        self.event.ReceiveTime = datetime.utcnow()
```
destogl: Thanks :)

opcua/server/internal_subscription.py
```diff
@@ -91,7 +91,7 @@ def _create_monitored_item(self, params):
 
             if params.ItemToMonitor.AttributeId == ua.AttributeIds.EventNotifier:
                 self.logger.info("request to subscribe to events for node %s and attribute %s", params.ItemToMonitor.NodeId, params.ItemToMonitor.AttributeId)
-                if self.aspace.get_attribute_value(params.ItemToMonitor.NodeId, ua.AttributeIds.EventNotifier).Value.Value != 1:
+                if self.aspace.get_attribute_value(params.ItemToMonitor.NodeId, ua.AttributeIds.EventNotifier).Value.Value & 1 == 0:
```
destogl: This doesn't make any sense... Better check the Variant, see how I did it in tests.

opcua/ua/uaevents_auto.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, sourcenode=None, message=None, severity=1, extended=False):
         self.SourceNode = sourcenode
         self.SourceName = None
         self.Time = None
-        self.RecieveTime = None
+        self.ReceiveTime = None
```
destogl: Thanks :)

opcua/server/internal_subscription.py
zerox1212: My version checks only bit 0, not the entire byte. I thought spec says EventNotifier is only a single bit of the byte?

zerox1212:
@destogl I didn't use JSON because I thought it was going to add a dependency. I guess it doesn't so I could change to that maybe. I knew this way would work and it's more uniform with Variable historizing approach. You guys can tell me what you want because I don't really know what I'm doing...

examples/server-history.py
zerox1212: I will check it, even though I don't think it matters for this implementation as a long as all the fields are there.

oroulet: as written ealier I think we should pass a list which can be either a list of properties to store, a list of events or a list of event type

opcua/server/history.py
```diff
@@ -38,7 +38,7 @@ def read_node_history(self, node_id, start, end, nb_values):
         """
         raise NotImplementedError
 
-    def new_historized_event(self, event, period):
+    def new_historized_event(self, source_id, etype, period):
```
oroulet: I propose 
`def new_historized_event(self, source_id, list_of_properties, period):`
But I have not thought much about it...

examples/server-history.py
oroulet: I just checked the spec. user can specify what eventtype they want to read from. so the argument should be a list of event types, not a list of properties. Then internally we get all properties for every event type

opcua/server/history_sql.py
```diff
@@ -137,19 +140,180 @@ def read_node_history(self, node_id, start, end, nb_values):
...
 
-    def read_event_history(self, start, end, evfilter):
-        raise NotImplementedError
+            if nb_values:
+                limit = nb_values + 1  # add 1 to the number of values for retrieving a continuation point
+            else:
+                limit = -1  # in SQLite a LIMIT of -1 returns all results
+
+            table = self._get_table_name(source_id)
+            clauses = self._get_select_clauses(source_id, evfilter)
```
oroulet: the evfilter also has the paramter typeId. I might be wrong but my understanding is that we should store evtype and check the typeId parameter of evfilter. if evtype == evfilter.typeID OR evtype is subtype of evfilter.typeId then we return value, otherwise we ignore.

But I do not understand why typeId is specified in every select clause, it should be a parameter of evfilter...

zerox1212:
I looked into the JSON storage method. It looks ok but then we need to subclass JSON encoder / decoder because default one doesn't support things like datetime. I'm also not sure if there is a speed impact of selecting specific fields with SQL vs. parsing all fields from JSON for ever result. I will try both ways I guess and you can decide.

oroulet:
I do not think we need json.

opcua/server/history_sql.py
zerox1212: By TypId do you mean the select clause EventType?

This is because a specific event type would be selected via the WhereClause. For this to work EventType field would need to be selected.

zerox1212:
So... I need some clarification after developing this further.

Is the idea to subscribe to an object that generates events? Or are we subscribing to a single event type?

Clients like UaExpert seem to subscribe to the object node, not a single event type, but our implementation leans towards subscribing to a specific event type.

Even more odd is that if you subscribe to events on a node and supply a specific event type, the subscription still gets called when ANY event is triggered, even if it's a type you didn't subscribe to.

When I read the spec it isn't very clear about this. What should we do?

destogl:
@zerox1212 Which part of spec is that? I will also take a look

oroulet:
@zerox1212 @destogl This is what I tried to explain in my last comments. a client subscribe to an object but with a filter (spec part 4 and 11). In our client I abstracted the filter stuff by using an event as argument, I am quite sure this is what 99.99% of the users will want, is some users want something else they can create their monitoredItemRequest by hand. The filter  (part4 7.17.3) has two parts
* the SelectClause which specifies the properties of an event a to send (including a typeId which we might use to filter what event to return)
* The WhereClause which we do not support yet, but maybe the where clause is what we should use to filter per event type. I have not understood it yet

zerox1212:
I think it's fine to just subscribe to all events from one object. I am also fine with historizing all events from an object. I don't really see the point to filter the subscription. Even adding the WhereClause right now seems like a waste of time as I'm guessing most client don't even use it (in my experience).

Besides, any calls to history for past events could be filtered separately by the user for now if they really needed it.

zerox1212:
Do you guy want two separate `enable_history_variable` and `enable_history_events` method to start and stop historizing?

Or do you prefer something like this:
```python
def enable_history(self, target, period=timedelta(days=7), count=0):
        """
        If target is a node, set attribute Historizing of node to True and start storing data for history
        If target is an event, check that the source node supports events, then enable event history read and historize
        """
        target_type = target.get_type_definition()
        # if isinstance(target, Node):
        if target_type == ua.ObjectIds.BaseDataVariableType:
            target.set_attribute(ua.AttributeIds.Historizing, ua.DataValue(True))
            target.set_attr_bit(ua.AttributeIds.AccessLevel, ua.AccessLevel.HistoryRead)
            target.set_attr_bit(ua.AttributeIds.UserAccessLevel, ua.AccessLevel.HistoryRead)
            self.history_manager.historize(target, period, count)

        # elif isinstance(target, ua.BaseEvent):
        elif target_type == ua.ObjectIds.BaseObjectType or target_type == ua.ObjectIds.Server:
            # to historize events of an object, first check if object supports events
            # FIXME: would be better to give an object to enable_history instead of event, then find it's event types
            # FIXME: via references; is it possible to get event types via GeneratesEvent attribute?

            # source_node = Node(self.isession, target.SourceNode)
            source_event_notifier = target.get_attribute(ua.AttributeIds.EventNotifier)
            # FIXME: some how need to only check bit 0 for true
            if source_event_notifier.Value.Value >= 1:
                # if it supports events, turn on bit 2 (enables history read)
                target.set_attr_bit(ua.AttributeIds.EventNotifier, 2)
                # send the event to history manager
                self.history_manager.historize_event(target, period)
        else:
            self.logger.error("Target doesn't support historizing")
```

oroulet:
I vote for two different metodes

zerox1212:
Can you explain how to rebase my fork? You merged changes I don't have so now I'm out of sync again. When I tried `git pull `r` it just says "Current branch master is up to date". Most likely because it's looking at my fork as the remote. Sorry I am new to git and I struggle with it's commands.

oroulet:
I never remember myself. Something like
Git remote add main HTTPS://xxxx addressee to main repository
Git pull -r main master

If not Google it : git pull from another depository

zerox1212:
You were correct about the commands, but it fails to rebase.

Issue #174, opened 2016-05-02
Issue with Subscribing to Multiple Event Types


zerox1212:
As I have been writing history support for events I found this issue.

If I set up two custom event types with event generators on the same object node, both event objects return the same properties.

See screen shots:
![image](https://cloud.githubusercontent.com/assets/13090472/14946025/b7628a14-0fe1-11e6-97e3-18ac42bbe0b8.png)

Event results that show up in SubHandler.event_notification
Object 2001, EventType 2003
![image](https://cloud.githubusercontent.com/assets/13090472/14946035/dafe1c7c-0fe1-11e6-9b49-3e56425a5b48.png)

Object 2001, EventType 2006
![image](https://cloud.githubusercontent.com/assets/13090472/14946037/e8747842-0fe1-11e6-8be3-89eed0293c52.png)

Code (event_types comes from your example in https://github.com/FreeOpcUa/python-opcua/pull/171 )
```python
ev_handlers = []
for event_type in event_types:
    handler = self._sub.subscribe_events(source, event_type)
    ev_handlers.append(handler)
self._handlers[source] = ev_handlers
```

destogl:
This is the problem you mentioned in #172. This depends not on event implementation but on subscription implementation. so we should take deeper look into spec and maybe reimplement subscriptions...

oroulet:
"reimplement" is a big word, but we surely need to fix this. I am guessing
this is due to the WhereClause which we currently ignore.
Btw does subscribing to events with uaexpert work currently? I tried
yesterday with your server-event.py file and I could not see anything in
uaexpert.,,,,

zerox1212:
If you take the code from my PR it works in UaExpert. However I think you need to set the start and end times a day early and a day late. I'm not sure what time zone it uses.

destogl:
For me worked without any issues.

oroulet:
ok I tested again. it works, I was just confused because in the example file you make the Object node generate events. This is unusual, I have never seen a server emit events from another node than Server. Did you? If not maybe we should also generate events from sever node in example.

another thing is that the common/event.py file now only contains server code. maybe you should move it?

also sourceName property seems to be a binary string in uaexpert. Is this normal?

And our answer to a modifyMonitoredItemRequest seems to be broken. Wireshark does not manage to parse it

destogl:
>This is unusual, I have never seen a server emit events from another node than Server. Did you? If not maybe we should also generate events from sever node in example.
[...]

PR #179, opened 2016-05-05
WhereClause support

oroulet:
I finally tok some time to understand WhereClause stuff.
WhereClause is clearly the correct way to filter by event type

Here is a first implementation to setup where clause from client when subscribing to events.
* this is not finished, we should have a recursive option to subscrive_events
* Need parsing of WhereClause on server, but this might be a bit complex...


opcua/common/subscription.py
```diff
@@ -201,22 +201,45 @@ def _get_node(self, nodeid):
...
-            op.TypeDefinitionId = eventtype.nodeid
+            op.TypeDefinitionId = evtype.nodeid
             op.AttributeId = ua.AttributeIds.Value
             op.BrowsePath = [property.get_browse_name()]
-            evfilter.SelectClauses.append(op)
-        return evfilter
+            clauses.append(op)
+        return clauses
+
+    def _where_clause_from_evtype(self, evtype):
```
zerox1212: I think this needs to take a list of event types so that it's possible to subscribe to multiple event types with a single filter and the WhereClause.

oroulet: Agree

opcua/common/subscription.py
```diff
@@ -201,22 +201,54 @@ def _get_node(self, nodeid):
...
+        for subtypeid in [st.nodeid for st in self._get_subtypes(evtype)]:
+            op = ua.LiteralOperand()
+            op.Value = subtypeid
+            el.FilterOperands.append(op)
+        el.FilterOperator = ua.FilterOperator.InList
+
+        cf.Elements.append(el)
+        return cf
+
+    def _get_subtypes(self, parent, nodes=None):
```
destogl: Is getting subtypes needed anywhere else? If yes than you could add it into Node class, since it is on the end node operation

opcua/common/subscription.py
```diff
@@ -201,22 +201,54 @@ def _get_node(self, nodeid):
...
+    def subscribe_events(self, sourcenode=ua.ObjectIds.Server, evtype=ua.ObjectIds.BaseEventType, evfilter=None):
         """
         Subscribe to events from a node. Default node is Server node.
         In most servers the server node is the only one you can subscribe to.
+        if evfilter is provided, evtype is ignored
         Return a handle which can be used to unsubscribe
         """
         sourcenode = self._get_node(sourcenode)
-        evfilter = self._get_filter_from_event_type(evtype)
+        if evfilter is None:
```
destogl: I would use `if not evfilter`

opcua/common/subscription.py
```diff
@@ -319,7 +351,7 @@ def get_event_properties_from_type_node(node):
             break
 
         parents = curr_node.get_referenced_nodes(refs=ua.ObjectIds.HasSubtype, direction=ua.BrowseDirection.Inverse, includesubtypes=False)
-        if len(parents) != 1: # Something went wrong
+        if len(parents) != 1:  # Something went wrong
```
destogl: Thanks :)

opcua/common/subscription.py
oroulet: I think 'is None' is correct here since None is the default value. they are equivalen here, but one could alway think that a user could pass a structture with a __boo__ method return False although they wanted really to pass it as arg: http://stackoverflow.com/questions/7152441/python-if-not-val-vs-if-val-is-none

opcua/common/subscription.py
oroulet: maybe but the Node class has already so many methods...

oroulet:
basic of where clause parsing in place. Not working and no tests yet.

opcua/server/internal_subscription.py
```diff
@@ -140,10 +142,12 @@ def _delete_monitored_items(self, mid):
 
     def datachange_callback(self, handle, value, error=None):
         if error:
-            self.logger.info("subscription %s: datachange callback called with handle '%s' and erorr '%s'", self, handle, error)
+            self.logger.info("subscription %s: datachange callback called with handle '%s' and erorr '%s'",
```
destogl: Is this just cosmetic change or are you limiting line on some specific length?

destogl:
BTW, recently we have a lot of additions/debugging, but we are not really documenting anything. I am aware there is existing website generated from docs, but I didn't contribute anything and as you said before the classes are getting bigger.

Could you give us few pointers regarding documentation (or maybe open a new issue regarding this.)

Thanks! The work seams good.

opcua/server/internal_subscription.py
oroulet: I have never been able to decide anything about line length. I usually do not care but those lines where to long for my screen. I am open to suggestions

oroulet: BTW my editor is set up for line length of 120. It might be an OK default

oroulet:
Yes we have an issue with documentation.  As you said  the documentation is auto generated so if you document your classes and methods it should be OK. In addition we can modify the intro in doc/index.rst maybe point to some good example files there. Or anything you think is missing....
And we have the readme

oroulet:
@destogl SourceName from events were displayed as hex in uaexpert, looking a the createmonitoreditem from uaexpert I saw that uaexpert was expecting a qualifiedname. in your code you sent a utf string from LocalizedName. In this patch I modified your code to send the browsename and uaexpert is happy. What should SourceName be?

opcua/server/internal_subscription.py
zerox1212: If you use PyCharm you get really good PEP8 checking.

zerox1212:
As far as documentation, I wouldn't mind going through a lot of the code and adding doc strings and clarifying existing doc strings. I'm also native English speaker. :)

oroulet:
>> I wouldn't mind going through a lot of the code and adding doc strings and clarifying existing doc strings.

That would be great!

>> I'm also native English speaker. :)

Event better :-) I wrote most things without re-reading myself and I am not a native speaker.

the best place to start would be the README and maybe the page on http://freeopcua.github.io/
That is what most people see first

opcua/server/internal_subscription.py
oroulet: >> If you use PyCharm 

after soon 20 years vim, my mind is hardcoded for it. But vim also checks pep8, syntax, etc so it is just a question on what is most readable...

destogl: But 80 characters length according to PEP8 I find not soo cool. I would than rather agree with 120 or something like that.

destogl:
> @destogl SourceName from events were displayed as hex in uaexpert, looking a the createmonitoreditem from uaexpert I saw that uaexpert was expecting a qualifiedname. in your code you sent a utf string from LocalizedName. In this patch I modified your code to send the browsename and uaexpert is happy. What should SourceName be?

I never had problem with displaying it. According to specs: Part 5, Section 6.4.2., Table 22 SourceName should be type String which are interpreted as BinaryString. I don't now what is exact differents between String and BinaryString

oroulet:
One is encoded, the other not. Today I changed your call to display_name
().Text returning bytestring  browse_name ().Name returning encoded string.
That should make everybody happy (apart those liking localization). The
spec is a bit broken there, source name should be a localized string

destogl:
What you are saying make sense, but than why not have it as LocalisedString if UaExpert is happy?

Because if you leave it this we sill have to change it ones we are start working on Localisation

oroulet:
Uaexpert is happy with anything, but It shows bytestring as bytes and this
looks strange when it should be be a 'name'.  The spec says source name
should be a string, so one way or another we should return a string. I
think the most correct way would be to add a method to_string() to
localized text which uses the client/local locale to return an encoded
string using correct locale, instead of using browse name. But I do not
understand very well this localized string stuff, I do not really care..

opcua/server/internal_subscription.py
oroulet: OK let says we use 120, this seems rather common.

oroulet:
@destogl if you have a server sending different event types, could you try to subscribe using this branch? I tried to follow the spec, but we should test this. few clients seems to send whereclause

destogl:
@oroulet: currently I don't have such a server. But this should be easily testable by subscribing specific type of events on the object sending different event types.

oroulet:
We can test with out own server and I already made a test for it, but we
should test with an external server generating different event types...
Just need to find one...

destogl:
Aha, if it that what you mean, I have external server generating Audit events and custom events. Can you just give me few points what should I exactly test and I will do it!

oroulet:
@destogl . Just pull out the branch and check that you still receive events when subscribing using client.subscribe_events(myev)
The other thing is to check if we correctly filter events using code like the one in test: When we subcsribe to using an event type we should only receive events of that type.


`
def test_several_different_events(self):

         objects = self.srv.get_objects_node()
         o = objects.add_object(3, 'MyObject')

        etype1 = self.srv.create_custom_event_type(2, 'MyEvent1', ua.ObjectIds.BaseEventType, [('PropertyNum', ua.VariantType.Float), ('PropertyString', ua.VariantType.String)])
        evgen1 = self.srv.get_event_generator(etype1, o)

        etype2 = self.srv.create_custom_event_type(2, 'MyEvent2', ua.ObjectIds.BaseEventType, [('PropertyNum', ua.VariantType.Float), ('PropertyString', ua.VariantType.String)])
        evgen2 = self.srv.get_event_generator(etype2, o)

        myhandler = MySubHandler2()
        sub = self.opc.create_subscription(100, myhandler)
        handle = sub.subscribe_events(o, etype1)

        propertynum1 = 1
        propertystring1 = "This is my test 1"
        evgen1.event.PropertyNum = propertynum1
        evgen1.event.PropertyString = propertystring1

        propertynum2 = 2
        propertystring2 = "This is my test 2"
        evgen2.event.PropertyNum = propertynum2
        evgen2.event.PropertyString = propertystring2
        
        for i in range(3):
            evgen1.trigger()
            evgen2.trigger()
        time.sleep(1)

        self.assertEqual(len(myhandler.results), 3)
        ev = myhandler.results[-1]
        self.assertEqual(ev.EventType, etype1.nodeid)

        handle = sub.subscribe_events(o, etype2)
        for i in range(4):
            evgen1.trigger()
            evgen2.trigger()
        time.sleep(1)


        ev1s = [ev for ev in myhandler.results if ev.EventType == etype1.nodeid]
        ev2s = [ev for ev in myhandler.results if ev.EventType == etype2.nodeid]

        self.assertEqual(len(myhandler.results), 11)
        self.assertEqual(len(ev2s), 4)
        self.assertEqual(len(ev1s), 7)

        #self.assertEqual(ev.SourceName, 'MyObject')
        #self.assertEqual(ev.SourceNode, o.nodeid)
        #self.assertEqual(ev.Message.Text, msg)
        #self.assertEqual(ev.Time, tid)
        #self.assertEqual(ev.PropertyNum, propertynum)
        #self.assertEqual(ev.PropertyString, propertystring)

        # time.sleep(0.1)
        sub.unsubscribe(handle)
        sub.delete()

`

zerox1212:
I reviewed the code and it looks good for me. This evening I will grab this branch and try implement history and see if https://github.com/FreeOpcUa/python-opcua/issues/174 is solved.

Seems like we could merge this soon.

oroulet:
Yes, I am just waiting for tests from destogl, then I will merge

oroulet:
one thing missing is subscribing at once to several event types. must add it before merge

destogl:
Hi,

I finally started testing and the line:

`handle = sub.subscribe_events(ua.ObjectIds.Server, ua.ObjectIds.BaseEventType)`

gives me following error: "opcua.common.uaerrors.UaStatusCodeError: The event filter is not valid.(BadEventFilterInvalid)".

So it seams something is wrong but unfortunately I have just binary from the server so can not debug...

oroulet:
ok thanks, I will try again here.

oroulet:
does wireshark shows the event filter well?

destogl:
@oroulet: last days I was sick so was not able to test this deeper, also today is my last day before vacation so I will probably be able to test this in 10 days. Sorry for this but I have currently to many other stuff to do... Hope this is OK...

oroulet:
not perfect but it does not seem to break things so merging to master

zerox1212:
Thanks for working on this and merging.

>one thing missing is subscribing at once to several event types. must add it before merge

Looks like only a single event type is supported at the moment, correct?

oroulet:
You can call method several times to subscribe to several events

zerox1212:
What is the correct way to implement this? Do we want to pass a list of events or do we just want to call the subscription for every event we want?

zerox1212:
How do I build the WhereClause for multiple event types? Do I do multiple filter operands or multiple filter elements?
image

zerox1212:
I'm not sure if this is the correct behavior, but this is what it looks like when I add two event types to the FilterOperands.
Event 1 (not that the event result object has attributes of both events, but the data returned is correct):
image
Event 2 (not that the event result object has attributes of both events, but the data returned is correct):
image

I'm going to make a pull request and move this comment.

PR #180, opened 2016-05-06
Events History with Tests

zerox1212:
This commit is very functional, but WhereClause is not supported yet.
Default history will fail tests because it raises NotImplemented on
event calls.


zerox1212:
Can we please look at reviewing and merging this? I did everything possible to make it complete, including tests.

There are getting to be so many changes that if you change the master and I have to rebase it's going to be very difficult to fix and I might have to stop working...

opcua/common/subscription.py
```diff
@@ -52,10 +52,21 @@ def __str__(self):
         return "EventResult({})".format([str(k) + ":" + str(v) for k, v in self.__dict__.items()])
     __repr__ = __str__
 
+    def get_field_variants(self):
+        field_vars = {}
+        for key, value in vars(self).items():
+            if not key.startswith("__") and key is not "server_handle":
+                if key is "SourceName":
+                    # hack because client is expecting string not byte string?
+                    field_vars[key] = ua.Variant(value.decode(encoding='utf-8'))
```
oroulet: remove the hack. will get fixed somewhere else

opcua/common/subscription.py
```diff
@@ -52,10 +52,21 @@ def __str__(self):
         return "EventResult({})".format([str(k) + ":" + str(v) for k, v in self.__dict__.items()])
     __repr__ = __str__
 
+    def get_field_variants(self):
```
oroulet: can you give that method a more understanble name? since it is now exposed. get_event_properties_as_field_list() or whatever make sense

opcua/server/history.py
```diff
@@ -162,7 +163,7 @@ def _create_subscription(self, handler):
         params.Priority = 0
         return Subscription(self.iserver.isession, params, handler)
 
-    def historize(self, node, period=timedelta(days=7), count=0):
+    def historize_var(self, node, period=timedelta(days=7), count=0):
```
oroulet: histprize_data_change would be a better name since this is what we use at other places

opcua/server/internal_server.py
```diff
@@ -153,16 +153,16 @@ def register_server2(self, params):
     def create_session(self, name, user=User.Anonymous, external=False):
         return InternalSession(self, self.aspace, self.subscription_service, name, user=user, external=external)
 
-    def enable_history(self, node, period=timedelta(days=7), count=0):
+    def enable_history_var(self, node, period=timedelta(days=7), count=0):
```
oroulet: same here: enable_history_data_change

opcua/server/internal_subscription.py
```diff
@@ -91,7 +91,11 @@ def _create_monitored_item(self, params):
 
             if params.ItemToMonitor.AttributeId == ua.AttributeIds.EventNotifier:
                 self.logger.info("request to subscribe to events for node %s and attribute %s", params.ItemToMonitor.NodeId, params.ItemToMonitor.AttributeId)
-                if self.aspace.get_attribute_value(params.ItemToMonitor.NodeId, ua.AttributeIds.EventNotifier).Value.Value != 1:
+                ev_notify_byte = self.aspace.get_attribute_value(params.ItemToMonitor.NodeId, ua.AttributeIds.EventNotifier).Value.Value
+                if ev_notify_byte is not None:
+                    if ev_notify_byte & 1 == 0:
+                        result.StatusCode = ua.StatusCode(ua.StatusCodes.BadServiceUnsupported)
+                else:
```
oroulet: if ev_notify_byte is not None or ev_notify_byte & 1 == 0:
           result.StatusCode = ua.StatusCode(ua.StatusCodes.BadServiceUnsupported)

that is the same and muvh simpler

oroulet:
I am fine with merging this. but we need to fix the tests before merging. I think the simplest thing to do now is to move your new tests to TestHistorySQL class then they will only be executed with sql backend

zerox1212:
I will make the changes you listed above tomorrow. How can I move the event tests to only the SQL class?

zerox1212:
>I am fine with merging this. but we need to fix the tests before merging. I think the simplest thing to do now is to move your new tests to TestHistorySQL class then they will only be executed with sql backend

Would you consider making the ":memory:" SQL history the default and removing the python dict? Seems like it makes more sense to just maintain one implementation. Both are in memory so I don't think there is a performance impact.

opcua/common/subscription.py
zerox1212: This is done.

opcua/common/subscription.py
zerox1212: This is done. It is now named `get_event_props_as_field_dict`.

opcua/server/history.py
zerox1212: This is done.

opcua/server/internal_server.py
zerox1212: This is done.

opcua/server/internal_subscription.py
zerox1212: This is done.

zerox1212:
Cleaned up the implementation as you noted. I also moved event tests to TestHistorySQL class. If you can merge it, I can start adding support for WhereClause.

zerox1212:
I think everything is fixed now. When event subscription supports the WhereClause I can add it to history as well.

The only minor complaint from myself is that the SQL code for historizing data values could be cleaned up to better match the events implementation for inserting/reading data. For now I think it's good enough.

Maybe next we can add alarms. :)

oroulet:
I did not have time to check everything but I merge since it does not seem to break anything. We can fix small things late.

oroulet:
>> Would you consider making the ":memory:" SQL history the default and removing the python dict?

Not sure what to do. My plan was to have the dict stuff as default as it does not require any dependencies but make the examples use the sql implementation AND a file as users of history will want a file.
Maintaining 2 implementation has the advantage to oblige us to thing about code re-use,  users will want to make their own storage class. To share as much code as possible between our implementationsm maybe we should modify th existing interface...

>> One other note, I don't think event history will react very well to adding custom event fields to a node that is already historized. All the custom fields need to be added before enable history is called.

This is fine. but document it!

PR #181, opened 2016-05-11
Perform lazy loading when restoring a cached address space

stanti:
When starting an opcua server, the creation of the address space is
currently a performance bottleneck. The startup process can be accelerated
by loading a pre-generated address space pickle.
However, the startup process still takes ~25 seconds on a raspberry pi
model b (compared to ~125 seconds when generating the address space from code).
Storing the address space in a shelve, where the data for each node is
pickeled individually, allows to further improve the startup performance
since only the nodes that are actually accessed are loaded from
disc (all other nodes are loaded later when they are accessed).
Since the default address space contains thousands of nodes but just a
small amount is actually accessed during startup, the startuptime could
be improved to ~3.5 seconds.

### DIFF
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -1,6 +1,8 @@
 from threading import RLock
 import logging
 from datetime import datetime
+import collections
+import shelve
 try:
     import cPickle as pickle
 except:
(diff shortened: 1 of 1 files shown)


oroulet:
Good idea

oroulet:
Btw could you document how you enabled this on readme? Add a section raspberry. And write how long it takes to start, so people know. Thanks

stanti:
Is there any way we could include this feature into the public API to make it easier to access this functionality?

Right now its a bit complicated to use it since one has to execute the script schemas/generate_address_space.py (which is not part of the setup package) to generate the addressspace first and then change
the source code in order to actually load from disc.

I thought about adding an optional parameter/property/.. somewhere which holds the path to the cache-file. When the parameter is omitted everything is loaded from code like usual. If a user
defines the parameter the addressspace is loaded from disc or the cache-file is created if it does not exist yet. That way the first startup would take longer and all following startups could make use of the
cachefile.
However, since all of the loading is currently executed in the constructor of the server object, the only way I see to implement that without breaking existing code is to add an additional optional parameter to the constructor...

zerox1212:
I use dump address space in my project to save a snapshot of the server whenever it stops. If there is no backup I load all my nodes from an XML instead.

I just wrote a function like this:

```python
    def start(self):

        if self.server is None:
            # server setup code

        # populating our address space
        try:
            self.server.load_aspace("address_space.bin")
            print("loaded the address space backup")
        except:
            print("No address space backup found, importing XML instead")
            # import nodes from xml
            self.server.import_xml("config_OPCUA_v6.xml")
```

Can you do something like this with shelving?

oroulet:
I am fine with an additional argument to constructor which is path to
cached file

oroulet:
merging this, but would nice if someone could expose this as arg to server

stanti:
sorry for replying so late.
i've created another pull-request which introduces the additional constructor argument

@zerox1212 
Your approach should work fine with shelving as well since this change just modifies the way the cached data is stored but not the data itself. However, I never benchmarked the time required for writing the shelve. It might be slower than with the previous implementation which could lead to a longer shutdown...

I'm afraid that the additional constructor parameter can not be avoided since always writing the cache at shutdown will cause problems when nodes are created from code (e.g. execute server.add_variable before startup; this node(s) will be cached on shutdown; server.add_variable will fail on the next startup since the node was imported from cache and already exists).

PR #184, opened 2016-05-13
Modify Monitored Item function and Deadband Monitor in the client side

Johen8:
Hi, I added these functions to the Client side. I don't think that the Modify Monitored Item function are in a good format, but they work fine if we want to change one monitored item.

### DIFF
--- a/opcua/client/ua_client.py
+++ b/opcua/client/ua_client.py
@@ -478,7 +478,7 @@ class UaClient(object):
         response.ResponseHeader.ServiceResult.check()
         return response.Results
 
-    def modify_monitored_item(self, params):
+    def modify_monitored_items(self, params):
         self.logger.info("modify_monitored_items")
         request = ua.ModifyMonitoredItemsRequest()
         request.Parameters = params
--- a/opcua/client/ua_client.py
+++ b/opcua/client/ua_client.py
@@ -478,13 +478,12 @@ class UaClient(object):
         response.ResponseHeader.ServiceResult.check()
         return response.Results
 
-
-def modify_monitored_item_low(self, params):
-    self.logger.info("modify_monitored_items")
-    request = ua.ModifyMonitoredItemsRequest()
-    request.Parameters = params
-    data = self._uasocket.send_request(request)
-    response = ua.ModifyMonitoredItemsResponse.from_binary(data)
-    self.logger.debug(response)
-    response.ResponseHeader.ServiceResult.check()
-    return response.Results
+    def modify_monitored_item(self, params):
+        self.logger.info("modify_monitored_items")
+        request = ua.ModifyMonitoredItemsRequest()
+        request.Parameters = params
+        data = self._uasocket.send_request(request)
+        response = ua.ModifyMonitoredItemsResponse.from_binary(data)
+        self.logger.debug(response)
+        response.ResponseHeader.ServiceResult.check()
+        return response.Results
--- a/opcua/client/ua_client.py
+++ b/opcua/client/ua_client.py
@@ -477,3 +477,14 @@ class UaClient(object):
         self.logger.debug(response)
         response.ResponseHeader.ServiceResult.check()
         return response.Results
+
+
+def modify_monitored_item_low(self, params):
+    self.logger.info("modify_monitored_items")
+    request = ua.ModifyMonitoredItemsRequest()
+    request.Parameters = params
+    data = self._uasocket.send_request(request)
+    response = ua.ModifyMonitoredItemsResponse.from_binary(data)
+    self.logger.debug(response)
+    response.ResponseHeader.ServiceResult.check()
+    return response.Results
--- a/opcua/common/subscription.py
+++ b/opcua/common/subscription.py
@@ -88,7 +88,6 @@ class Subscription(object):
         self.subscription_id = response.SubscriptionId  # move to data class
         self.server.publish()
         self.server.publish()
-        self.filters = dict()
 
     def delete(self):
         """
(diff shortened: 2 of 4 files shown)


opcua/common/subscription.py
```diff
@@ -334,3 +334,50 @@ def get_event_properties_from_type_node(node):
         curr_node = parents[0]
 
     return properties
+
+
+def modify_monitored_item(self, handle, new_samp_time, new_queuesize = 0, mod_filter = None):
```
oroulet: why are these function but with self as argument? should'nt they be methods?

opcua/common/subscription.py
```diff
@@ -334,3 +334,50 @@ def get_event_properties_from_type_node(node):
...
+    req_params.QueueSize = new_queuesize
+    req_params.Filter = mod_filter
+    req_params.SamplingInterval = new_samp_time
+    return req_params
+
+
+def deadband_monitor(self, var, deadband_val, deadbandtype = 1, queuesize = 0, attr=ua.AttributeIds.Value):
+    """
+    Function to create a subscription with a Deadband Value.
+    Default deadband value type is absolute.
```
oroulet: this is function(should be a method) to create a monitored item or to subscribe to data changes with deadband

Johen8: Thanks for the feedback. Will look into it!

Johen8:
Hello @oroulet, what do you think? I also added the deadband filter support on the server side. (The absolute one only, the percentage one is a little more complicated). If you or anyone else have any remarks, please tell me.

opcua/client/ua_client.py
```diff
@@ -477,3 +477,14 @@ def history_read(self, params):
         self.logger.debug(response)
         response.ResponseHeader.ServiceResult.check()
         return response.Results
+
+
+def modify_monitored_item_low(self, params):
```
oroulet: Why  _low ? Low level? All methods have the same level here. Otherwise looks good.

opcua/server/address_space.py
```diff
@@ -515,7 +515,7 @@ def set_attribute_value(self, nodeid, attr, value):
 
         for k, v in cbs:
             try:
-                v(k, value)
+                v(k, value, old)
```
oroulet: I am not very keen to send the last value in call back. As you wrote there are several different kind of filters that can be implemented. I think it should be the role of internal subscription to remember last value if it is necessary......

opcua/client/ua_client.py
Johen8: Yes, that was the point. I was writing everything in the same file and named it like that to differentiate them easier. Will correct it.

opcua/server/address_space.py
Johen8: Yes, I am not a great fan of this too. Especially since it breaks the first callback when we create the subscription. I modified it so that the two arguments are the same, and therefore on a deadband filter we got no callback. I will see what I can make different. Thanks

oroulet: though a bit about is and I rest my case :-) make internal subscription remember the necessary data. and maybe put the code in its own class if you can, internal_subscription is already too complex

Johen8: Hi, sorry for taking that long to respond, I was working on another aspect of the project. I was thinking about this now, and, if I were to create a new class, say, "Monitored Item Values" to store current and old value, I was thinking of adding it to the MonitoredItemData class. Do you think this could be a problem? Is this class defined in the specfication? I took a quick look and didn't find anything.

oroulet:
When did you made the last commit? I just saw it now. You code looks correct. It seems there it a conflict with master. Can you do a git pull -r https://github.com/FreeOpcUa/python-opcua master?

Johen8:
It was this last Friday. I'm using an 'old' version of the repository, will update it. But before I just noticed something and would like your opinion: when sending a modify request to server, what should I send as filter if I don't want to change the filter? Because in the way I did, I was sending a deadband filter with a -1 value to the server and I was going to change it so that when it detected this filter, it would keep the same. But I think this wouldn't comply with the specification. Also, in the way the code is done in the server, if I send a 'None' to the server, it will replace the current filter with nothing.
Should I recreate the filter everytime? Or perhaps we should detect when a 'None' filter is sent in the server so that it know it should keep the old one, but the problem is if I want to remove the filter, what should I send then?

oroulet:
I guess we should store the filter in subscription class when creating the
monitoreditem. store it in a dict with server_handle as key. and store the
filter from server response, not our request... (writting this without
reading the code so take this as a sugestion)

Johen8:
Okay, I think it is fixed now. Added a new field to Subscription to store the filters. I tried the pull, but it was not very succesfull, I got this:
image

oroulet:
then just read the error message and git tells you how to fix it :-)
1-> open the conflict file
2-> fix whatever is marked (probably very little), nmaybe check byt calling git diff
3-> git add subscription.py
4-> git rebase --continue

oroulet:
then git push --force origin master  when you are happy

oroulet:
Sorry I may have suggested something wrong here. There is already an Subscription attribuute called _subscription_data which is a dict of client_handle:SubscriptionItemData(), You should store your filter there instead of in a new dict.

Johen8:
I think you are talking about the '_monitoreditems_map' attribute? If yes, then yes, it actually already contains the filter. The problem is that it is a dict that has as keys the client handles. I'll see what I can change so that it is not necessary to pass that as argument as well.

oroulet:
Maybe it is possible to use only server-handle?

Johen8:
I did it so that when we modify an item, it iterates through the map with the server handle given, to find the corresponding item with the client handle we want to change.

opcua/common/subscription.py
```diff
@@ -328,10 +327,14 @@ def modify_monitored_item(self, handle, new_samp_time, new_queuesize=0, mod_filt
         :param mod_filter_val: New deadband filter value
         :return: Return a Modify Monitored Item Result
         """
+        for monitored_item_index in self._monitoreditems_map:
```
Johen8: Here.

Johen8:
I think it should be ok now. (I hope I didn't break any of your updates).

oroulet:
ok

PR #186, opened 2016-05-16
Lazyloading

stanti:
Add lazy loading feature to public API


oroulet:
even better that what I had envisionnned

stanti:
thx

PR #187, opened 2016-05-18
Correctly handle count parameter @ HistoryDict.save_node_value

stanti:
The current implementation is actually a no-op since the data is truncated
correctly but not assigned back to self._datachanges.
Pop the first value instead since this function is called for every
datachange which means that not more than one value will every be removed
in a single call.

### DIFF
--- a/opcua/server/history.py
+++ b/opcua/server/history.py
@@ -91,7 +91,7 @@ class HistoryDict(HistoryStorageInterface):
         data.append(datavalue)
         now = datetime.utcnow()
         if period:
-            while now - data[0].ServerTimestamp > period:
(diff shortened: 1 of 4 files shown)


oroulet:
I agree that the last line should copy the list, so it looks like an error but your fix breaks tests...so either we have en error in tests or something else is happening in this code.... There is also a sql implementation, would be strange if we had the exact same error in sql...

stanti:
in the testcode a maximal data value count of 10 is set before historizing 20 data values. seems like the following testcases expect 20 data values to be preset - thats why actually enforcing the count-limit seems to be break testcases

oroulet:
I thought I would try to fix this today, but you already did all the work. great!

PR #188, opened 2016-05-23
event history api tweak

oroulet:
take event as arg not evfilter
return EventResult obj and not fields
update command line to read events


oroulet:
@zerox1212 did som changes to the history api. I understand that this is also how you wanted it. 
need to fix tests since API changed

zerox1212:
This mostly looks like a refactoring of event stuff into events.py now that we have WhereClause support.

When I look at the failed tests it's only the default history dict implementation with failures. All the SQL tests are OK.

oroulet:
tests are OK now. API has changed. it now takes an event as arg not an evfilter and returns EventResult, not event_fields

PR #189, opened 2016-05-24
Subscription Supports List of Event Types, History Update

zerox1212:
Creating a subscription now supports a list of event types for creating a single event filter. Need to confirm it's implemented correctly (see other comments). History updated to support historizing multiple event types from a single source node.

Some doc strings also improved.


zerox1212:
I'm not sure if this is the correct behavior, but this is what it looks like when I add two event types to the FilterOperands.
Event 1 (not that the event result object has attributes of both events, but the data returned is correct):
image
Event 2 (not that the event result object has attributes of both events, but the data returned is correct):
image

opcua/common/subscription.py
```diff
@@ -208,39 +208,58 @@ def _get_node(self, nodeid):
             node = Node(self.server, ua.NodeId(nodeid))
         return node
 
-    def _get_filter_from_event_type(self, eventtype):
-        eventtype = self._get_node(eventtype)
+    def _get_filter_from_event_type(self, eventtypes):
+        # FIXME must be a nicer way to make a list out of eventtypes when it's passed
+        # FIXME an integer (such as ua.ObjectIds.BaseEventType)
```
oroulet: I do not think there is a better way. but add support for tuples with 'type(xx) in (list, tuple):'

opcua/common/subscription.py
```diff
@@ -208,39 +208,58 @@ def _get_node(self, nodeid):
...
+        evtypes = []
+        if type(eventtypes) is not list:
+            evtypes.append(eventtypes)
+        else:
+            evtypes = eventtypes
+
+        evtype_node_list = []
+        for evtype in evtypes:
+            evtype_node_list.append(self._get_node(evtype))  # make sure we have a list of Node objects
+
```
oroulet: now I understand:
`
if not type(evtypes) in (list, tuple):
    evtypes = [evtypes]
evtypes = [self._get_node(i) for i in evtypes]
`

opcua/common/subscription.py
```diff
@@ -208,39 +208,58 @@ def _get_node(self, nodeid):
...
+            op = ua.SimpleAttributeOperand()
+            op.TypeDefinitionId = evtype.nodeid
+            op.BrowsePath.append(ua.QualifiedName("EventType", 0))
+            op.AttributeId = ua.AttributeIds.Value
             el.FilterOperands.append(op)
-        el.FilterOperator = ua.FilterOperator.InList
+            for subtypeid in [st.nodeid for st in self._get_subtypes(evtype)]:
+                op = ua.LiteralOperand()
+                op.Value = ua.Variant(subtypeid)
+                el.FilterOperands.append(op)
```
oroulet: I think the result is wrong. it should be something lile
operator.inlist [SimpleAttributeOperand() (for EventType), type1, type2, type3, type4, etc....)
so you need to write something like:

      op = ua.SimpleAttributeOperand()
      op.TypeDefinitionId = evtype.nodeid
      op.BrowsePath.append(ua.QualifiedName("EventType", 0))
      op.AttributeId = ua.AttributeIds.Value
      el.FilterOperands.append(op)
     all_subtypes = []
     for evtype in evtypes:
           all_subtypes += self._get_subtypes(evtype)
     all_subtypes = list(set(all_stubtypes)) # remove all duplicates
     for subtypeid in [st.nodeid for st in self._get_subtypes(evtype)]:
             op = ua.LiteralOperand()
             op.Value = ua.Variant(subtypeid)

opcua/common/subscription.py
```diff
@@ -251,7 +270,7 @@ def _get_subtypes(self, parent, nodes=None):
             self._get_subtypes(child, nodes)
         return nodes
 
-    def subscribe_events(self, sourcenode=ua.ObjectIds.Server, evtype=ua.ObjectIds.BaseEventType, evfilter=None):
+    def subscribe_events(self, sourcenode=ua.ObjectIds.Server, evtypes=ua.ObjectIds.BaseEventType, evfilter=None):
         """
         Subscribe to events from a node. Default node is Server node.
         In most servers the server node is the only one you can subscribe to.
```
oroulet: write in doc that one can give either an eventype or a list of event type

opcua/common/subscription.py
```diff
@@ -208,39 +208,58 @@ def _get_node(self, nodeid):
...
+        selected_paths = []
+        for evtype in evtypes:
+            for prop in get_event_properties_from_type_node(evtype):
+                if prop.get_browse_name() not in selected_paths:
+                    op = ua.SimpleAttributeOperand()
+                    op.TypeDefinitionId = evtype.nodeid
+                    op.AttributeId = ua.AttributeIds.Value
+                    op.BrowsePath = [prop.get_browse_name()]
+                    clauses.append(op)
+                    selected_paths.append(prop.get_browse_name())
```
oroulet: it might be that you need to remove the line with op.TypeDefinitionId = evtype.nodeid. It is not necessary and we might get issue for different events

oroulet:
sorry I broke the merge by merging the other pull request for events...

opcua/common/subscription.py
zerox1212: I need to look at this closer because I don't really understand what you mean. I actually didn't read the spec on WhereClause yet.

opcua/common/subscription.py
zerox1212: Again I don't really understand WhereClause stuff. Should I remove it or not? Your original code had it, so I left it.

zerox1212:
>sorry I broke the merge by merging the other pull request for events...

How can i synchronize? Should I rebase from the master again and update this PR?

oroulet:
We've few underlying issues here. We miss data type in process for example. I will try to have as look tonight. Do not try to merge yet

zerox1212:
If you want we could merge it into a branch on the master repository so it's easier for you to finish up my work.

oroulet:
@zerox1212 Can you grab that branch and try to fix it? https://github.com/FreeOpcUa/python-opcua/pull/190 I could not really understand the sql stuff.
try to run tests and you will see that reading events from sql fails.
I change the storage interface, to move all the field creation outside of storage classes.
When it works, you can also add your new changes to it

oroulet:
now master should be OK. Can you rebase? I think your approach is correct.

opcua/common/subscription.py
oroulet: yes I had it but because I only supported one event. Now you subscribe at several events at once so we should not have it

zerox1212:
I was on vacation. Now that I'm back I'll try to rebase so you can merge.

oroulet:
Great to see you back! You need to rebase and fix the comments then should be ok

zerox1212:
Well I did something wrong with git and totally screwed up my repository. I'm just going to close this and remake it because I don't know how to fix it anymore...

zerox1212:
Ah so many things are different now :(. 

This is taking a long time to implement again.

Making ```get_filter_from_event_type()``` take a Node as a parameter is really a pain. I have to think about how I can make sure that this function gets a list of Nodes now because it's called in two different places (both subscription and node).

oroulet:
sorry for that, I am trying not to touch that event code, but there were so many things irritating me :-(

getting Node objects should be a one line before calling method: maybe modify a bit get_node to make test for arg type

```
    evtypes = [self._get_node(i) for i in evtypes]
    filter = get_filter_from_event_types(evtypes)
```

zerox1212:
We can't do that anymore because you deleted ```_get_node()``` from events class.

oroulet:
Because it is not needed. Just called us.NodeId(whatever)

oroulet:
ua.Node(whatever)

PR #190, opened 2016-05-26
implement dict based event history, cleanup event stuff

oroulet:


opcua/common/events.py
```diff
@@ -19,19 +24,66 @@ def get_event_props_as_fields_dict(self):
...
+    @staticmethod
+    def from_field_dict(fields):
+        """
+        Create an Event object from a dict of name and variants
+        """
+        result = EventResult()
+        for k, v in fields.items():
+            setattr(result, k, v.Value)
+            result.data_types[k] = v.VariantType
+        return result
```
oroulet: @zerox1212 this method should make it easy for you to create an event and return it in read_event_history

opcua/server/history_sql.py
```diff
@@ -150,7 +152,7 @@ def read_node_history(self, node_id, start, end, nb_values):
 
             return results, cont
 
-    def new_historized_event(self, source_id, ev_fields, period):
+    def new_historized_event(self, source_id, ev_fields, period, count):
```
oroulet: coutn was missing here, it is missing implementation

opcua/server/history_sql.py
```diff
@@ -244,24 +246,22 @@ def read_event_history(self, source_id, start, end, nb_values, evfilter):
...
+                    fdict = {}
+                    for i, field in enumerate(row):
                         # if the field is the _Timestamp column store it in a list used for getting the continuation
                         if i == 0:
                             cont_timestamps.append(field)
                         else:
                             if field is not None:
-                                hist_ev_field_list.EventFields.append(ua.Variant.from_binary(Buffer(field)))
+                                fdict[clauses[i+1]] = ua.Variant.from_binary(Buffer(field))
+                                print(clauses[i+1], fdict[clauses[i+1]])
```
oroulet: here we should just fill the dict with name:variant. but I do not know how to get name, or I am missing EventType

opcua/common/events.py
zerox1212: I thought only the subscription was supposed to return EventResult? I think in my original implementation we discussed this and you told me not to create an EventResult on historical event read.

opcua/server/history_sql.py
zerox1212: From what I remember, the spec says events don't have ```count```.

opcua/server/history_sql.py
zerox1212: I can't remember exactly. The EventFields isn't a dict. It's a list and the order of the variants must match the event filter. If the select clause is Node1, Node2, Node3; the variants that get read out of history must be returned in a list in that order.

opcua/server/history_sql.py
oroulet: that would be very strange...

opcua/server/history_sql.py
oroulet: I fixed it

PR #193, opened 2016-05-30
default strings and ByteString to None

oroulet:
Must be tested, might have side effects


nuccy:
Prosys now connects correctly and interacts with python-opcua normally. Thanks.

PR #200, opened 2016-06-04
fix, renaming and restructure of event stuff

oroulet:


opcua/common/events.py
```diff
@@ -162,16 +179,19 @@ def get_event_obj_from_type_node(node):
...
 
                 while curr_node.nodeid.Identifier != parent_identifier:
                     for prop in curr_node.get_properties():
-                        setattr(self, prop.get_browse_name().Name, prop.get_value())
+                        name = prop.get_browse_name().Name
+                        val = prop.get_data_value()
+                        self.add_property(name, val.Value.Value, val.Value.VariantType)
                     parents = curr_node.get_referenced_nodes(refs=ua.ObjectIds.HasSubtype, direction=ua.BrowseDirection.Inverse, includesubtypes=False)
-                    if len(parents) != 1: # Something went wrong
-                        return None
```
oroulet: @destogl why this test?

oroulet:
@zerox1212 more incompatble changes here .... we should merge this first

oroulet:
now working again with prosys client. still need to check history

oroulet:
event history is working, nothing to fix

opcua/common/events.py
destogl: Because to check if we get only one parent. It just to be sure that structure is set-up right.

opcua/server/internal_subscription.py
```diff
@@ -228,26 +229,9 @@ def _trigger_event(self, event, mid):
             return
         fieldlist = ua.EventFieldList()
         fieldlist.ClientHandle = mdata.client_handle
-        fieldlist.EventFields = self._get_event_fields(mdata.filter, event)
+        fieldlist.EventFields = event.to_event_fields(mdata.filter.SelectClauses)
         self.isub.enqueue_event(mid, fieldlist, mdata.queue_size)
 
-    def _get_event_fields(self, evfilter, event):
```
destogl: Where did you move this method?

tests/tests_server.py
```diff
@@ -179,24 +180,23 @@ def test_get_event_from_type_node_BaseEvent(self):
     def test_get_event_from_type_node_Inhereted_AuditEvent(self):
         ev = opcua.common.events.get_event_obj_from_type_node(opcua.Node(self.opc.iserver.isession, ua.NodeId(ua.ObjectIds.AuditEventType)))
         self.assertIsNot(ev, None)  # we did not receive event
-        self.assertIsInstance(ev, ua.BaseEvent)
-        self.assertIsInstance(ev, ua.AuditEvent)
+        self.assertIsInstance(ev, BaseEvent)
+        self.assertIsInstance(ev, AuditEvent)
         self.assertEqual(ev.EventType, ua.NodeId(ua.ObjectIds.AuditEventType))
-        self.assertEqual(ev.Severity, ua.Variant(1, ua.VariantType.UInt16))
```
destogl: I would not change this line, since is important to have right VariantType....

opcua/common/events.py
oroulet: I changed it to exception. Should be ok?

tests/tests_server.py
oroulet: In events some properties were python built-in, other variants, I changed class to only use built-in. Type is stored in self.data_type (yes a bit ugly). We can add test for data type if you think it is important.
Anyway events are defined by constructing event type where data type can be clearly specified

opcua/common/events.py
destogl: Yes!

PR #203, opened 2016-06-07
Subscribe Multiple Events

zerox1212:
This doesn't pass tests and I don't know how to do multiple events in a way @oroulet likes since events got changed.

I don't have much time to work on this for the next months. It was working before all the events changes, but now I don't have time to fix it. Maybe someone else can tell me how to fix it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ dist
 *.swp
 newdocs
 examples/history.db
+*.sql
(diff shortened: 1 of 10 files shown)


zerox1212:
The main problem I have is that Event class can no longer get event type node objects (method is gone and Event doesn't have reference to the server object). Not sure how you want to handle this.

oroulet:
I will have a look when I get two minutes. It looks mostly correct apart from the filters. But it is easy to fix.

zerox1212:
@oroulet has picked up this PR to ensure it fits in with how events have been restructured.

See https://github.com/FreeOpcUa/python-opcua/pull/204

PR #204, opened 2016-06-08
accept subscription to event type list

oroulet:
must be tested with external servers
@zerox1212 I just added the small fixes I suggested earlier and it seems to fix everything. Have a look

### DIFF
--- a/opcua/common/node.py
+++ b/opcua/common/node.py
@@ -404,15 +404,9 @@ class Node(object):
         if not type(evtypes) in (list, tuple):
             evtypes = [evtypes]
 
-        # FIXME not a very nice way to make sure events.get_filter gets a list of nodes...
-        evtype_nodes = []
-        for evtype in evtypes:
-            if not isinstance(evtype, Node):
-                evtype_nodes.append(Node(self.server, ua.NodeId(evtype)))  # make sure we have a list of Node objects
-            else:
-                evtype_nodes.append(evtype)
+        evtypes = [Node(self.server, evtype) for evtype in evtypes]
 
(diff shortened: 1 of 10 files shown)


opcua/common/node.py
```diff
@@ -404,15 +404,9 @@ def read_event_history(self, starttime=None, endtime=None, numvalues=0, evtypes=
...
             evtypes = [evtypes]
 
-        # FIXME not a very nice way to make sure events.get_filter gets a list of nodes...
-        evtype_nodes = []
-        for evtype in evtypes:
-            if not isinstance(evtype, Node):
-                evtype_nodes.append(Node(self.server, ua.NodeId(evtype)))  # make sure we have a list of Node objects
-            else:
-                evtype_nodes.append(evtype)
+        evtypes = [Node(self.server, evtype) for evtype in evtypes]
```
oroulet: python list comprehension is great,  one simple line instead of 6... @zerox1212

zerox1212: Are you sure this will work? What if evtypes get passed a Node object? Some how I was having problems here if a Node object was passed, or maybe it was when only an int node ID was passed...

oroulet: Should work. I changed init of Node a while back to accept crazy things. If you think a test I missing feel free to add it 😃.
I see you still have a comment about things to check in code. Will have a look.

zerox1212: >I see you still have a comment about things to check in code. Will have a look.

What you did in node.py needs to be done in subscription.py. I think that's it. If I get time I'll add a test for custom event properties, but I'll probably do it in a separate PR. Don't have much free time at the moment.

PR #206, opened 2016-06-12
do not stop client for protocol error, but try again

oroulet:


destogl:
This would be probably good way to catch the bug, but for it we need to print stack-trace, but then the problem is that we don't know exactly what is the problem :/

oroulet:
I think I understand why the error on client side happen see bug request. But the one on server overload is unknown. Logger.exception should print stack. We will see

oroulet:
seems to fix a bug in client, merging

PR #207, opened 2016-06-13
Make History Consistent with Events

zerox1212:
Remove some event FIXMEs from subscription and make it match node.py.

Minor changes and cleanup to make history module consistent with events. Mainly that historizing new events now takes a list of event type nodes. The module which implements the history interface is responsible for formatting the events types itself (for example converting event types to a list of fields like in SQL)

I also cleaned up history examples and added a new example for historizing data changes on a variable, since at some point that example code was lost during all the events stuff.

### DIFF
--- a/opcua/server/history_sql.py
+++ b/opcua/server/history_sql.py
@@ -132,9 +132,12 @@ class HistorySQLite(HistoryStorageInterface):
 
             return results, cont
 
-    def new_historized_event(self, source_id, ev_fields, period, count=0):
+    def new_historized_event(self, source_id, evtypes, period, count=0):
         with self._lock:
             _c_new = self._conn.cursor()
+
+            # get all fields for the event type nodes
+            ev_fields = self._get_event_fields(evtypes)
 
             self._datachanges_period[source_id] = period
             self._event_fields[source_id] = ev_fields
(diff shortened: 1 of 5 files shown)


opcua/server/history_sql.py
```diff
@@ -231,6 +234,25 @@ def read_event_history(self, source_id, start, end, nb_values, evfilter):
...
+        """
+        # get all fields from the event types that are to be historized
+        ev_aggregate_fields = []
+        for event_type in evtypes:
+            ev_aggregate_fields.extend((events.get_event_properties_from_type_node(event_type)))
+
+        ev_fields = []
+        for field in set(ev_aggregate_fields):
+            ev_fields.append(field.get_display_name().Text.decode(encoding='utf-8'))
+        return ev_fields
```
oroulet: Use list comprehension here to simplify and shorten code.

And are you sure you do not get duplicates?

oroulet:
All good. Apart from small comment

opcua/server/history_sql.py
zerox1212: I did this a while ago... I think the set ```set(ev_aggregate_fields)``` is what removes any duplicates.

oroulet: Correct!

zerox1212: I need to use .extend() here ```ev_aggregate_fields.extend```. Will changing this to a list  comprehension still work?

destogl:
I would than just define ev_fields as a set and not a list. Than you are sure there is no duplicates

opcua/server/history_sql.py
oroulet: ev_aggregate_fields = [field for fields in events.get_event_properties_from_type_node(event_type) for field in fields]
ev_fields = [field.get_display_name().Text.decode(encoding='utf-8') for field in set(ev_aggregate_fields)]

I admit the first line (nested list comprehension) if not very clear and a simple for loop is easier to understand but the second one is easy.
We can in fact make this a one liner....... ;-)

oroulet:
merging anyway now. Send a new PR for the small fixes

PR #210, opened 2016-06-15
Support providing objecttype for the method add_object of Node

bitkeeper:
The Node.add_object let you only create object of the type BaseObjectType. It would be nice if you also could create other types by providing an objecttype.

Like:
myobj = objects.add_object(idx, "MyCustomObject", myobjectype_nodeid)

The only thing this patch do is setting the correct NodeId in the TypeDefinition.


destogl:
This should be extended adding also object members... Do we have somewhere TODO list?

oroulet:
Then this would become the instanciate method I started. Feel free to look
at it @destogl 😃 instantiate.py under common

PR #213, opened 2016-06-15
xmlimport breaks when UAVariable doesn't have value element

bitkeeper:
The fragment below breaks then XmlImporter:
```xml
<UAVariable NodeId="i=30004" BrowseName="MyXMLVariableWithoutValue" DataType="String">
    <References>
         <Reference ReferenceType="HasTypeDefinition">i=69</Reference>
         <Reference ReferenceType="Organizes" IsForward="false">i=30002</Reference>
    </References>
 </UAVariable>
```

The import doesn't break if an value element is present. According to the OPC UA schema it is allowed to don't have a value element ( see https://opcfoundation.org/UA/schemas/1.02/UANodeSet.xsd).

The error is due a small typo in XmlParser.__init__ :
```python
  self.value = [] # should be None instead
```
The default value of a value is set to [], while at XmlImporter.add_variable it is checked to unequal to None:
```python
  if obj.value is not None:
```
 Set the default value to None fixes the issue.

Test data of the testcase TestServer.test_xml_import is updated to test this situation.

### DIFF
--- a/examples/server-customobject.py
+++ b/examples/server-customobject.py
@@ -42,11 +42,8 @@ if __name__ == "__main__":
     server.start()
 
     try:
-        count = 0
         while True:
             time.sleep(1)
-            count += 0.1
-            myvar.set_value(count)
     finally:
         # close connection, remove subcsriptions, etc
         server.stop()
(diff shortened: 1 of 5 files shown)


zerox1212:
Checking obj.value for none was actually from a different issue, but I think your fix should work.

https://github.com/FreeOpcUa/python-opcua/commit/6fca9e326a9da3c887e5229e88be52feb4f8d362

oroulet:
Why did you close this one?

bitkeeper:
New to git ... made new commits and they showed up is this pull request.
I will make a new pull request.

PR #217, opened 2016-06-19
Add missing Enums, add move some code to ua_utils.py

oroulet:


oroulet:
@destogl @zerox1212 I found out we were missing definition of EventNotifier and had hardcoded bit position at different places. If you know places we should use thes new Enums let me know, I may habe missed places...

PR #222, opened 2016-06-21
JSON Module

zerox1212:
This code implements a simple JSON encoder for both server and client.

This PR is to discuss if this should be part of the library or not.

The reason I made it is so that I can easily take OPC UA data and send it to cloud services like Azure. This could be useful for other users as you can use this simple Python OPC UA library to communicate machine data to a cloud platform in a few lines. The user would only need to implement AMQP, MQTT or HTTP libraries to transmit the JSON data.

In the future of OPC UA we might need a JSON encoder for the pub/sub standard as well. However this isn't part of the spec at all so it's up to people who use this library.

If we do not need this, perhaps it should be opened as a separate project.

What do you all think?


oroulet:
This is an encoder for a few of the ua objects. Not sure where it should be. Maybe it should stay in its own repository, I can create a new repository for it under freeopcua group if you want. 
Supporting json as s communication format will require a different approach, the code should be auto generated from xml spec. This is what we should do for the xml protocol which we do not support yet.

zerox1212:
OK, no problem. I will keep it in my own repository.

Thanks.

PR #224, opened 2016-06-29
Support xml address space import with custom data types

bitkeeper:
Support custom data types with the XmlImporter. The datatype can be base on:

- Type based on string version of nodeid
`<UAVariable DataType="ns=1;i=3004" ParentNodeId="ns=1;i=5002" NodeId="ns=1;i=6006" BrowseName="1:MyVar">`

- Type based on ua ojectids
`<UAVariable DataType="String" ParentNodeId="ns=1;i=5002" NodeId="ns=1;i=6006" BrowseName="1:MyVar">`

- Typed based on alias
`<UAVariable DataType="MyVarType" ParentNodeId="ns=1;i=5002" NodeId="ns=1;i=6006" BrowseName="1:MyVar">`

custom_nodes.xml is updated to let TestServer:test_xml_import verify the change.

### DIFF
--- a/opcua/common/xmlimporter.py
+++ b/opcua/common/xmlimporter.py
@@ -45,28 +45,25 @@ class XmlImporter(object):
         if obj.parent:
             node.ParentNodeId = ua.NodeId.from_string(obj.parent)
         if obj.parentlink:
-            node.ReferenceTypeId = self.to_ref_type(obj.parentlink)
+            node.ReferenceTypeId = self.to_nodeid(obj.parentlink)
         if obj.typedef:
             node.TypeDefinition = ua.NodeId.from_string(obj.typedef)
         return node
 
-    def to_data_type(self, nodeid):
+    def to_nodeid(self, nodeid):
         if not nodeid:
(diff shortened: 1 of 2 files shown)


zerox1212:
Could you also update xmlparser.py fuction ```_parse_value``` to support the UA Modeler prefix "uax:"?

bitkeeper:
@zerox1212 :
Strange because I don't have any issues at all with the uax ns, I have included an example xml for the TestServer.test_xml_import:
[custom_nodes.zip](https://github.com/FreeOpcUa/python-opcua/files/340878/custom_nodes.zip)

Crucial is that the uax names space is declared, see to of the xml document.

But should problems with uax no be a different issue ? Because it has nothing to do with support for custom types ?

oroulet:
I had a short look and it looks like to_ref_type and to_data_type should be merged to a method called _to_nodeid . Currently we have duplicates code and double tests! Can you have a look?

bitkeeper:
to_ref_type and to_data_type are merged into to_nodeid.

PR #231, opened 2016-07-05
Added support for custom datatypes and removed some trailing spaces

destogl:


bitkeeper:
Thanks. 

1 ) I had to make a small change to get it working:
```python
   nodeid, qname = _parse_add_args(*args[:2])
   if len(args) > 3:
       node_class = args[3]
```
to
```python
   nodeid, qname = _parse_add_args(*args[:2])
   if len(args) > 2:
       node_class = args[2]
```
2) For the method create_subtype; By providing a default value( ua.NodeClass.ObjectType)  for nodeclass you can prevent changing the other methods to add the additional argument.

destogl:
You are right! Thanks!

Issue #226, opened 2016-07-01
Question about server event


TheAtlantis:
Hi, 

I would like catch client item subscribe event on my server.
How catch this event ?

Thank you for you help.

Michel

oroulet:
No code for that. By I guess you can add a callback in create subscription
in subscription service class

oroulet:
And send PR

TheAtlantis:
Thank for your fast reply. 
I try this and PR if the solution is fine.

Issue #227, opened 2016-07-01
ServerDiagnostics question


TheAtlantis:
Hi, 

I have set Server/Auditing and Server/ServerDiagnostics/EnabledFlag to 1 and with an external OPC UA Client (Unified Automation UaExpert) I try to subscribe to Server/ServerDiagnostics/CurrentSessionCout for exemple. 
I gets no diagnostic value ( same for others nodes ). 

Do you have an idea to get diagnostics ?

Michel

oroulet:
The server has absolutely no code for server diagnostic. If you know how it
should be used, patch are welcome

TheAtlantis:
Ok, thanks.

zerox1212:
Somewhat related, how can we properly implement ServerCapabilities? We should make an effort here to populate this object with the capabilities of the python OPC UA server.

PR #232, opened 2016-07-05
Issue #226

TheAtlantis:
Add server Event for client Item subscription

Usage : 
`self.server.subscribe_server_event("create_monitored_items", self.create_monitored_items)
self.server.subscribe_server_event("modify_monitored_items", self.modify_monitored_items)
self.server.subscribe_server_event("delete_monitored_items", self.delete_monitored_items)`

### DIFF
--- a/examples/server-callback.py
+++ b/examples/server-callback.py
@@ -1,0 +1,89 @@
+import sys
+sys.path.insert(0, "..")
+import logging
+from datetime import datetime
+
+try:
+    from IPython import embed
+except ImportError:
+    import code
+
+    def embed():
+        vars = globals()
+        vars.update(locals())
+        shell = code.InteractiveConsole(vars)
+        shell.interact()
+
+
+from opcua import ua, uamethod, Server
+from opcua.common.callback import CallbackType
+
+
+
+def create_monitored_items(event, dispatcher):
+    print("Monitored Item")     
+
+    for idx in range(len(event.response_params)) :
+        if (event.response_params[idx].StatusCode.is_good()) :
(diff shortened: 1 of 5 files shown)


destogl:
Can this be somehow merged with current event implementation? At least for the event side. And I think we have already EventDispatcher for the changeValue events, can this be reused?

TheAtlantis:
Maybe but I don't see how to do that. 
The SubscriptionService create a subscribe on internal item ( with StatusCode, PublishingInterval etc.).
I think to reusing EventDispatcher to implement ServerDiagnostics #227  for example. 

If you have a better solution, I'll take it.

TheAtlantis:
Actually, the server class provide subscribe_server_event or unsubscribe_server_event method. 
If you want I can rename this method (and more : EventDispatcher, etc..) with "callback" like this : 
- subscribe_server_callback 
- CallbackDispatcher

destogl:
OK. I was misguided with the name event. Renaming it to callbacks seams good. If you could add some tests for it into `test/test_server.py` would be great :)

oroulet:
@TheAtlantis OK,  rename them as you think is best but avoid using Event terminology. Maybe Signal is another possibility...
Then I do not like using arbitraty strings as event/callback keys, this is very error prone. What about using an enum? other proppositions?

destogl:
Enums are good. Similar to keys in `VariantType` class

TheAtlantis:
I have commit server exemple for callback use and use callback as event name. 
Callback keys are enum now.

destogl:
I would move `callback.py` from `common` into `server`

oroulet:
looks OK. Although as @destogl wrote, maybe this is server only?

PR #233, opened 2016-07-05
Add method for adding custom type objects with or without properties

destogl:


destogl:
After we agree on #235 I will add this functionality for all this methods and then merge

PR #234, opened 2016-07-05
Example custom enumerations

bitkeeper:
If you want to try it out; doesn't work on the current master it requires fixes for

add_subtype - requires extra argument for NodeClass. see PR #231
add_variable - requires extra argument for datatype. see PR #235

### DIFF
--- a/examples/server-enum.py
+++ b/examples/server-enum.py
@@ -76,7 +76,7 @@ if __name__ == "__main__":
     myobj = objects.add_object(nsidx, 'MyObjectWithEnumVar')
 
     # add var with as type the custom enumeration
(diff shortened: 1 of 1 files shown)


bitkeeper:
Looks like the Travis error has nothing to do with this specific PR ?

destogl:
This seams to be Travis error, but we take care about that when are stuff are clear

destogl:
@bitkeeper: Could you check #241 if this is OK for you for adding custom types. It would be good to update your example according to that and than we can start work on instantiate as @oroulet proposed.

bitkeeper:
The requirements for this PR are already merged (PR #231 #235). Is it possible to first merge this PR and that I then make a new PR for an undated example and also include an updated tests/server-customobject.py.

destogl:
OK

bitkeeper:
Thanks !

PR #235, opened 2016-07-05
Add datatype to create_variable

bitkeeper:
Used to make #234 work.

### DIFF
--- a/opcua/common/manage_nodes.py
+++ b/opcua/common/manage_nodes.py
@@ -80,17 +80,14 @@ def create_property(parent, *args):
 def create_variable(parent, *args):
     """
     create a child node variable
-    args are nodeid, browsename, value, [variant type]
-    or idx, name, value, [variant type]
+    args are nodeid, browsename, value, [variant type], [data type]
+    or idx, name, value, [variant type], [data type] 
     """
     nodeid, qname = _parse_add_args(*args[:2])
-    val = _to_variant(*args[2:])
-    if len(args) > 3:
-        datatype = args[3]
-        if  isinstance(datatype, ua.NodeId) == False:
-            raise RuntimeError()
-    else:
-        datatype = None
+    val, datatype = _to_variant_with_datatype(*args[2:])
+    if  datatype and not isinstance(datatype, ua.NodeId):
+        raise RuntimeError()
+    
     return node.Node(parent.server, _create_variable(parent.server, parent.nodeid, nodeid, qname, val, datatype, isproperty=False))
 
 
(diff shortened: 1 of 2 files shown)


opcua/common/manage_nodes.py
```diff
@@ -85,7 +85,13 @@ def create_variable(parent, *args):
     """
     nodeid, qname = _parse_add_args(*args[:2])
     val = _to_variant(*args[2:])
-    return node.Node(parent.server, _create_variable(parent.server, parent.nodeid, nodeid, qname, val, isproperty=False))
+    if len(args) > 3:
```
destogl: This part should be move into `_to_variant` function because it can be also used for adding properties (or this doesn't make sense?)

opcua/common/manage_nodes.py
```diff
@@ -85,7 +85,13 @@ def create_variable(parent, *args):
...
+        datatype = args[3]
+        if  isinstance(datatype, ua.NodeId) == False:
```
destogl: Better write `if not isinstance(datatype, ua.NodeId)`

destogl:
For `_to_variant` function try something like:

```
def _to_variant(val, vtype=None, datatype=None):
    if isinstance(val, ua.Variant):
        if vtype:
            datatype = vtype
        return val, datatype
    else:
        return ua.Variant(val, vtype), datatype
```

opcua/common/manage_nodes.py
```diff
@@ -85,7 +85,13 @@ def create_variable(parent, *args):
     """
```
destogl: Add to comment: "[data type]"

oroulet: I think we should move that code to me method create _ variable _ type

opcua/common/manage_nodes.py
bitkeeper: Ok I will change this

opcua/common/manage_nodes.py
bitkeeper: I don't think that this functionality should be moved to create_variable_type, because this functionality is about creating an variable of a certain type and not about creating a variable type.

This method is used as part of creating variable:
-as part of creating variable for a certain object type 
-as part of creating variable for a object type instance

bitkeeper: @destogl about moving the code _to_variant:

No problem I will make the change, only I find it not very clear code that the datatype is extract in something called _to_variant.

In the current and new version you still can you use a variant as value and also use the datatype, only as long as you set vtype to None. Looks error prone ...

side mark; personally I find the construction with multiple ( use of *args a little bit vague, makes it harder to understand what the exact signature of a method is.

destogl: You can propose a new name for the function :)

destogl: @oroulet 
> I think we should move that code to me method create _ variable _ type

We need this here for instantiate_object function anyway

oroulet: @destogl instantiate is implemented using low level calls since we need
complete flexibility there to copy all attribute. But it is not finished....

destogl: @oroulet OK.

@bitkeeper plase check discussion in #236 regarding the semantics of functions. So it seams good to add this functions with *_write* sufix

PR #237, opened 2016-07-06
fix __exit__ method of client and server class

georgschoelly:
This seems to have been implemented without testing. As far as I can tell this never worked.

Context managers work the same way since Python 2.5 when they were introduced: https://docs.python.org/2.5/lib/typecontextmanager.html

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -102,7 +102,7 @@ class Client(object):
         self.connect()
         return self
 
-    def __exit__(self):
+    def __exit__(self, exc_type, exc_value, traceback):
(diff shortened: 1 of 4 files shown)


destogl:
could you add test for it?

georgschoelly:
I've added some tests, I've tried to only test the basic functionality necessary and not everything that happens upon `client.connect()` and similar. 

However, I'm not sure how useful they really are. This only protects very simple code and if there was a bug in it, it would be easy to detect it.

oroulet:
looks OK for me

PR #241, opened 2016-07-11
Custom types

destogl:
Updated adding of custom types with an command. Now enables also setting of datatype for variables and properties.


oroulet:
As written somewhere else I have issues with the syntax. This syntax does not allow to create new types with more than one level of children. This might be ok for events but not for types. I am not against adding such shortcuts but we need  create _xxx_type methods. Hopefully will send a proposition very soon

destogl:
But are the `_xxx_type` methods solution for that? Than I didn't get it...

This is also wrong PR! Please check #242

Issue #236, opened 2016-07-05
add_node API


oroulet:
Hi,
@destogl @bitkeeper 

Looks like a lot of people are discussing these methods currently and adding extra arguments everywhere. I think we should take a step back and think about it.

Currently we have 4 methods create_object, create_variable, create_property, create_folder and create_method. All of them are available in Node class as add_xxx.  Of this 2 (create_property and create_folder) are just commodity wrappers.
These methods can be used to quickly populate an address space. This is however in many cases not the correct way to do it, one should create new object types and and then instantiate them

Thus I propose to add three new methods create_object_type, create_variable_type and create_data_type, that take all the necessary arguments: NodeClass, ObjectIds, etc...  (Rmq: these methods are not random, they are the three missing types under types in standard address space apart from references)
Then make the instantiate method work, this should reduce the need to add even more arguments to the existing methods and not add more methods to the Server class. Also remove the possibility to add objecttype as arg to create_object, this duplicate instantiate functionality it is confusing. 
This will also remove simplify every method and remove some bugs, it looks for example that we use ObjectTypeAttribute instead of ObjectAttribute in _create_object...

create_object(nodeid, bname)  # if extra nodeid argument is present,  it instantiate the given node
create_variable(nodeid. bname, variant) # if extra nodeid argument is present,  it instantiate the given node
create_object_type(nodeid, bname, objecttype)
create_variable_type(nodeid, bname, datatype) #value is always None
create_reference_type TO_BE_DONE

then some simple wrappers for very common operations:
create_folder(qname, baname)
create_property(qname, bname, variant)

Then some methods for complex types:
create_method(xxxxxx) # existing
create_enum(xxxxxxx)  # TO BE DONE

instantiate(parent_node, node_type, idx)  #recursive instantiate but can be used for example for enum
instantiate_one_node(parent_node, node_type, idx)  # find better name, same as above, not recursive, maybe only internal. 

maybe we can make create_object() and create_variable() call instantiate if objecttype or datatype is provided.....not sure

Populate address space without caring about types (current API)
```python
obj = server.get_object_node()
                                          
myobj = obj_type.create_object(idx, 'MyObject')
myobj.create_variable(0, 'MySimpleVar', 0)
myobj.create_variable(0, 'MySimpleVar', 0, nodeid_of_custom_datatype)  # call instantiate in background
myobj.create_property(0, 'MyComplexVar', 0.2)
```

Example create custom object type, including some vars
```python
[...]

PR #243, opened 2016-07-12
Updated examples to reflect the proposed changes in the API

bitkeeper:
As requested examples updated.

Only I'm in doubt about examples/server-enum.py line 79. After the enum value there is an argument with the value None. This argument can currently be used for specifying the vtype of a variant.
In #236 the argument is gone. Should I keep the argument or remove it,

```python
 79: myenum_var = myobj.create_variable(nsidx, 'MyEnum2Var', MyEnum.ok, None, myenum_type.nodeid)
```


examples/server-customobject.py
```diff
@@ -16,28 +22,44 @@
...
 
+    # Example 1 - create a basic object
+    #-------------------------------------------------------------------------------
+    myobj = objects.create_object(idx, "MyObject",)    
+    #-------------------------------------------------------------------------------
+
+
+    # Example 2 - create a new object type and a instance of the new object type
+    #-------------------------------------------------------------------------------
+    types = server.get_node(ua.ObjectIds.BaseObjectType)
```
destogl: i would write this like
```
server.create_object_type(create_custom_object_type, idx, "MyCustomObject")
```

In this way you "sqash" two lines into one. And this is meant to be used by the users since protects from doing unintended errors.

examples/server-enum.py
```diff
@@ -50,7 +50,7 @@ def enum_to_stringlist(a_enum):
 
     # 1.
     # Create Enum Type
-    myenum_type = enums.add_subtype(nsidx, 'MyEnum', ua.NodeClass.DataType)
+    myenum_type = enums.create_data_type(nsidx, 'MyEnum')
```
destogl: can you also squash line and line 64 into one?

Tip: use `server.create_custom_data_type` (It is possible that method should be extended)

examples/server-customobject.py
bitkeeper: You mean line squash 36+37 ?
Line 37 did contain an error create_object_type should be create_object_type. Already fixed and committed:
```python
    mycustomobj_type = types.create_object_type(idx, "MyCustomObject")
```
I think about replacing line 36 it by the following construction:
```python
    # types = server.get_node(ua.ObjectIds.BaseObjectType) # this only a shortcut for standard types
    object_type_to_derive_from = server.get_root_node().get_child(["0:Types", 
                                                                   "0:ObjectTypes", 
                                                                   "0:BaseObjectType"])
```
This more descriptive, makes it more clear how to extended the code to derive from even your own types.

destogl: I meant:

```
server.create_custom_object_type(idx, "MyCustomObject")
```

oroulet:
@bitkeeper I just merged API changed, Can you have a look at examples/server-instantiate-object.py and update your code? The None parameter is now gone. In add_variable, give keyword arg datatype=xxx and/or varianttype=xxxx. Does it look OK for your use?

bitkeeper:
@oroulet Back from Holidays. I have update the example to the latest state of the API, including in instantiate method. Works great !

oroulet:
Can you try to rebase you PR? it has conflicts. Thanks

PR #244, opened 2016-07-13
restructure use of types

oroulet:
- [x] add create_xxx_type functions (available through Node.add_xx_type)
- [x] finish instantiate function
- [x] remove add_subtype method (it creates a node using ObjectType attrtibutes but set node class to whatever.....)
- [x] use instantiate function inside create_xxx function 
- [x] add tests
- [x] add create_reference_type function
- [x] update all examples


opcua/common/manage_nodes.py
```diff
@@ -85,14 +77,48 @@ def create_variable(parent, *args):
...
-    nodeid, qname = _parse_add_args(*args[:2])
+    nodeid, qname = _parse_nodeid_qname(*args[:2])
     val, datatype = _to_variant_with_datatype(*args[2:])
     if datatype and not isinstance(datatype, ua.NodeId):
         raise RuntimeError()
 
     return node.Node(parent.server, _create_variable(parent.server, parent.nodeid, nodeid, qname, val, datatype, isproperty=False))
 
 
+def create_variable_type(parent, nodeid, bname, datatype):
```
destogl: In this was clients can add types wherever they want. Is this OK solution or should we add some additional checking that the NodeId is descendant of BaseType?

destogl:
You can also create real TODO list here. Check Task Lists [here](https://guides.github.com/features/mastering-markdown/).

To easier instantiate objects and not go through address space to search for all variables and properties the good approach would be to create python objects for every type, as it is done for events in `event_objects.py`.

opcua/common/manage_nodes.py
oroulet: good point

oroulet: but maybe we should let user do what they want, adding types somewhere else in tree might be something some people want and it does not break anything anyway

destogl:
It seams there is error because the reference for the folder is not set correctly

opcua/common/manage_nodes.py
```diff
@@ -115,44 +141,53 @@ def create_method(parent, *args):
...
-        attrs.IsAbstract = True
+    if node.Node(server, parentnodeid).get_type_definition() == ua.ObjectIds.FolderType:
+        addnode.ReferenceTypeId = ua.NodeId(ua.ObjectIds.Organizes)
     else:
-        if node.Node(server, parentnodeid).get_type_definition() == ua.ObjectIds.FolderType:
-            addnode.ReferenceTypeId = ua.NodeId(ua.ObjectIds.Organizes)
-        else:
-            addnode.ReferenceTypeId = ua.NodeId(ua.ObjectIds.HasComponent)
+        addnode.ReferenceTypeId = ua.NodeId(ua.ObjectIds.HasComponent)
+    addnode.NodeClass = ua.NodeClass.Object
```
destogl: this should be under else

opcua/server/address_space.py
```diff
@@ -108,12 +108,13 @@ def _is_suitable_ref(self, desc, ref):
     def _suitable_reftype(self, ref1, ref2, subtypes):
         """
         """
+        if not subtypes and ref2.Identifier == ua.ObjectIds.HasSubtype:
+            return False
         if ref1.Identifier == ref2.Identifier:
             return True
-        #TODO: Please check the changes here
```
destogl: is here error with tests?

Please try by reverting this two lines...

oroulet: not that change is correct. Looking at the test, I think the test was originally wrong.

oroulet: now I moved all chnages linked to this in one commit to make it clearer

destogl: Looks good. But which type is the reference now

oroulet: References are node I'd, this test is a bit wrong we should compare entire
node ID, not only identifier.. Probably a performance optimization, but not
sure we should keep it

destogl: OK, this could be easily fixed... but after you finish this PR

oroulet:
Now I am finished with what I thought was necessary. plannig to merge soon if no big comment arrives

examples/server-example.py
```diff
@@ -74,11 +72,26 @@ def multiply(parent, x, y):
     uri = "http://examples.freeopcua.github.io"
     idx = server.register_namespace(uri)
 
-    # get Objects node, this is where we should put our custom stuff
+
+    # create a new node type we can instantiate in our address space
+    base_type = server.get_root_node().get_child(["0:Types", "0:ObjectTypes", "0:BaseObjectType"])
```
zerox1212: Thoughts on making a shortcut to this? ```server.BASETYPE.add_object_type()```

oroulet: Good idea. Maybe the same as the other shortcuts server.get_ base_type_
node() ?

zerox1212: Either is fine for me. I was just thinking that it might be cleaner to get a few "constants" when the server is created. If you just got a reference to the base type node on init, we do not need more methods in the server class.

However I don't know if that is a proper python convention.

oroulet: Upper case for class members is not part of pep5 but maybe something like:
Server.objects
Server. Server
Server. Basetype

Or:
Server.nodes. Objects
Server.nodes. Server
Or:
Server.shortcuts. Objects
Server.shortcuts. Server

Any better ideas?

oroulet: added implementation for nr:2 . any comments?
this changes a very common API so we should think a bit about it before merging

zerox1212: Looks interesting. I like it that way, but we need to make sure everyone else does.

Did you forget to commit shortcuts.py?

oroulet: Looks like I did :-)
Committing it tomorrow. It is just a class with common nodes as members

destogl: It looks good. It is also applicable if we ever create nodes as python objects, than you would be able to do this on any node/type :)

Great idea guys!

destogl:
Can we somehow also test the examples? It would be nice to automate this. The test would be just if example finished normally. For that one parameter "running_test" could be useful to have

oroulet:
Can require some work to automated examples they are not made to end... but we could tests them with pylint or flake8 to avoid stupide bugs

destogl:
This was the idea, to avoid stupid bugs when changing some stuff in API

oroulet:
@destogl I guess you can simply add a subprocess.call("pylint examples/") in tests but I am wondering how to display good error reports...

destogl:
For me is sufficient to see if one of the examples are not running, than developer itself should look closely on this example.... Fastest solution :)

PR #245, opened 2016-07-13
Add possibility to overide internal server

TheAtlantis:
Add possibility to overide internal server.

### DIFF
--- a/opcua/server/server.py
+++ b/opcua/server/server.py
@@ -69,7 +69,7 @@ class Server(object):
 
     """
 
-    def __init__(self, cacheFile = None):
+    def __init__(self, cacheFile = None, iserver = None):
         self.logger = logging.getLogger(__name__)
         self.endpoint = urlparse("opc.tcp://0.0.0.0:4840/freeopcua/server/")
         self.application_uri = "urn:freeopcua:python:server"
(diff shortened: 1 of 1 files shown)


TheAtlantis:
Exemple : 

`class InternalSessionCR(InternalSession):
    def create_monitored_items(self, params):       
        for item in params.ItemsToCreate :
            FourByteNodeId = item.ItemToMonitor.NodeId            
            nodeId = ua.NodeId(FourByteNodeId.Identifier, FourByteNodeId.NamespaceIndex)
            if (FourByteNodeId not in self.aspace.keys()) :       
                self.iserver.ServerOPCUA.createFromNodeId(FourByteNodeId)

        # Call herited method
        result = super().create_monitored_items(params)     
        return result
    
class InternalServerCR(InternalServer):

    def create_session(self, name, user=User.Anonymous, external=False):
        return InternalSessionCR(self, self.aspace, self.subscription_service, name, user=user, external=external)


iserver = InternalServerCR(None)
iserver.ServerOPCUA = self
# now setup our server
self.server = Server(iserver = iserver)
`
In this exemple, I catch item monitoring request to create item if dosn't exist en address space.
Idea : OPC UA server create node if a client request it.

destogl:
The code seams good. But why is there no Travis action?

oroulet:
@destogl no idea...

PR #249, opened 2016-07-26
replace custom test disabling with unittest.skip

georgschoelly:
Instead of using our own solution for skipping test, we use
unittest.skip. This has the additional benefit that it works
with the automatic test discovery in Python >=3.2.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -3,14 +3,6 @@ import logging
 import sys
 sys.path.insert(0, "..")
 sys.path.insert(0, ".")
-try:
-    from opcua.crypto import uacrypto
-    CRYPTOGRAPHY_AVAILABLE = True
-except ImportError:
-    print("WARNING: CRYPTO NOT AVAILABLE, CRYPTO TESTS DISABLED!!")
-    CRYPTOGRAPHY_AVAILABLE = False
-
-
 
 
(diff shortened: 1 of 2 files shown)


georgschoelly:
I have another small question: the tests throw exceptions, e.g:

```
test_historize_variable (tests_server.TestServer) ...
ERROR:asyncio:Exception in callback InternalSubscription._sub_loop()
    handle: <TimerHandle when=1292932.271 InternalSubscription._sub_loop()>
  File "C:\Program Files (x86)\Python35-32\lib\asyncio\events.py", line 125, in _run
    self._callback(*self._args)
  File ".\opcua\server\internal_subscription.py", line 279, in _sub_loop
    self.publish_results()
  File ".\opcua\server\internal_subscription.py", line 305, in publish_results
    self.callback(result)
  File ".\opcua\common\subscription.py", line 108, in publish_callback
    self._call_event(notif)
  File ".\opcua\common\subscription.py", line 144, in _call_event
    data = self._monitoreditems_map[event.ClientHandle]
KeyError: 201
```

There are a few more (like 4 or so and some printed text).

Is that a problem of my setup or does that happen for everyone?

oroulet:
It is normal to see some crash during tests as some tests provoke crash to ensure we react correctly. But I am unsure about this particular stack....

Issue #250, opened 2016-07-26
[Server] Import XML and generate namespace


maljac:
Hi, 

I am currently working with several companion specifications and trying to load them via `server.import_xml(filename)`.
Since the companion standards use the namespaceId 1 in their spec, the xml import always tries to import exactly in the defined namespace index that is given in the file instead of in the namespace index which must be generated by the server on startup. 

As I read [this](https://opcfoundation.org/forum/opc-ua-companion-standards/namespaces-in-companion-specifications/#p197) it seems that the "correct" way is:

1. Load UA namespace (ns:0)
2. Create custom namespace (ns:1) -> `server.register_namespace(uri)` 
3. Load XML File and translate the nodes given in the xml file to correct namespace (ns:2)
4. Load XML File and translate the nodes given in the xml file to correct namespace (ns:3)

The order of the `Uri` elements in `<NameSpaceUris>` should determine the file local namespace indexes:

```
    <NamespaceUris>
        <Uri>http://somemodel.org</Uri>
        <Uri>http://anothermodel.org/</Uri>
    </NamespaceUris>
```

Then in the XML file, references to `ns=1;i=...` must be translated to the the index of the first uri, i.e. `http://somemodel.org` while nodes reference to `ns=2;i=...` must be translated to the index of the second uri.

The index of the uris must be managed by the server, i.e. the order of loading models determines the namespace index.

Since I am a novice in OPC, I hope I described the scenario correct. Just wondering how other handle this scenario.

oroulet:
I do not know what the correct behaviour is supposed to be but overridding the namespace specified in xml sounds broken. and increasing the default namespace at every import even more broken....

maybe we could offer option to override namespace index when importing, but what do we do if xml specifies several namespaces?

I am expecting that companion specification make XML files defining node TYPES  in namespace 0 (maybe 1??) and then users instanciate these new node types in the namespace they wish. But as I wrote earlier I have not read anything about this so if you find references saying otherwise let me know....



>>  i.e. the order of loading models determines the namespace index.

That soundsd definitely wrong

maljac:
I did a test with UAModeler and created two models:
```
[...]

PR #252, opened 2016-07-27
Feature: Use namespaces in xml import

maljac:
See also #250


maljac:
Close due to wrong rebase

PR #253, opened 2016-07-27
Use namespaces in XML import

maljac:
Resolves issue #250

### DIFF
--- a/opcua/common/xmlparser.py
+++ b/opcua/common/xmlparser.py
@@ -53,7 +53,8 @@ class RefStruct(object):
 
 class XMLParser(object):
 
-    def __init__(self, xmlpath):
+    def __init__(self, xmlpath, server):
+        self.server = server  # POC
         self.logger = logging.getLogger(__name__)
         self._retag = re.compile(r"(\{.*\})(.*)")
         self.path = xmlpath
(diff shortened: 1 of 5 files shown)


opcua/common/xmlparser.py
```diff
@@ -53,7 +53,8 @@ def __init__(self):
 
 class XMLParser(object):
 
-    def __init__(self, xmlpath):
+    def __init__(self, xmlpath, server):
+        self.server = server  # POC
```
maljac: @oroulet Not sure if you are happy with this.

oroulet: Send necessary to implement what you proposed. If it does not break things, should be ok

maljac:
PyPy: 
test_datetime_read (tests_client.TestClient) ... FAIL

All other checks green...

oroulet:
probably because the pypy server ran so slowly that it tok more than 1 second

oroulet:
looks OK but one think I though of: The xml parser is used at startup to load standard address space, this is already very slow. Might this patch slow down startup time? Is so maybe you could disable check if namespace is 0? other ideas?

maljac:
1. The checks only run if there are namespaces defined in the xml (Creating the namespaces and store them in a local dictionary)
2. I havent noticed any notable permformance problems while importing two XML models (~500 lines and ~6000 lines). 

I could add a additional check which would skip the regex check in `_get_node_id` if the namespace dict is empty. This would save the execution time for the pre-compiled regex.

oroulet:
Ok

DrMatthiasArnold:
I'm trying to import the OPC-UA Device companion standard node2.ml utilizing xml_import.
However, besides the DeviceSet object no additional type definitions etc are visible.
Any source code example which sheds some light on the usage of xml_import)

PR #255, opened 2016-07-27
store status code in UaStatusCodeError

georgschoelly:
This is useful to handle certain status codes differently from others, for example I know how to handle `BadConditionBranchAlreadyAcked` but not all the other codes. Example:

```
from opcua.ua.status_codes import StatusCodes

try:
    raise UaStatusCodeError(StatusCodes.BadConditionBranchAlreadyAcked)
except UaStatusCodeError as e:
    if e.code == StatusCodes.BadConditionBranchAlreadyAcked:
        pass
    else:
        raise
```
There are some alternative approaches:

 * pass the full `StatusCode` object instead of only the value
 * make a subclass of `UaStatusCodeError` for each status code, then we could catch them like this:
    
    ```
    try:
        raise BadConditionBranchAlreadyAcked()
    except BadConditionBranchAlreadyAcked:
        pass # we can just ignore this
    except UaStatusCodeError as e:
        logger.warn("Exception found: {}".format(e))
    # crash on everything else
   ```


maljac:
The most Pythonic way would be to have a subclass of `UaStatusCodeError` for each code. Is it possible to generate the exceptions based on a xml / text specification?

georgschoelly:
Should be possible, we can overwrite `__new__` and return a more specific exception.

We could also do this in a later step without breaking backwards compatibility.

oroulet:
Error codes are already auto-generated, so it is easy to make a new exception for every error code. Look at schmeas/generate_statuscode.py. 
make for every error a simple class like 
```
class BadConditionBranchAlreadyAcked(UaStatusCodeError):
   pass
```

I am wondering if it impacts performance, if not it sound very pythonic...

georgschoelly:
I don't see how this would affect the performance much, especially because exceptions are already the exceptional case and should not be raised often.

maljac:
1. I dont see any performance issues
2. I think using the `UaError` prefix for identifying errors that are related to the spec is a good convention
3. According name of exceptions: Even if it does not match the spec 100%, in my opinion it is better to use python conventions. So `BadErrorName` over `Bad_ErrorName`.

Just my opinion.

oroulet:
I am unsure here, but I think I vote for 

```
class BadConditionBranchAlreadyAcked(UaStatusCodeError):
     pass
```

would be nicee with Ua  prefix but those names are already sooooooooooo long. I prefer to change names from spec as little as possible

oroulet:
Current patch looks OK. @georgschoelly  are you alos looking at autogenerating exception per error code?

georgschoelly:
@oroulet: Yes.

georgschoelly:
Implemented here: #275

PR #256, opened 2016-07-28
ignore BadNoSubscription and BadSessionClosed

georgschoelly:
might still change depending on #255

### DIFF
--- a/opcua/client/ua_client.py
+++ b/opcua/client/ua_client.py
@@ -10,7 +10,7 @@ from functools import partial
 
 from opcua import ua
 from opcua.common import utils
-from opcua.common.uaerrors import UaError, BadTimeout
+from opcua.common.uaerrors import UaError, BadTimeout, BadNoSubscription
 
 
 class UASocketClient(object):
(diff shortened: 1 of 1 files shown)


georgschoelly:
I was thinking of adding a test, but I don't see how I would do this. The error is caught inside of a `future` and then printed to the logger, so there never was an exception to catch.

How would I write a proper test for something like this? I could imagine something with a lot of mocks, but I'd be afraid the tests would be very fragile and break often.

oroulet:
Looks like you are looking at something which might be hard to solve. What is your goal exactly? avoiding error messages, when closing subcsription?

georgschoelly:
@oroulet yes, i'm trying to avoid printing the error messages. The reason for that is that I got those error messages and expected that something went wrong, so I had to debug and look it up in the spec only to find out that this is not a problem at all.

If we could catch those "good" errors, we could make using the library a better experience.

oroulet:
What is the status of this one? some of the changes seems to make sense like filtering out the  BadNodSubscription error

georgschoelly:
It is ready for merging. I was thinking about improving `answer_check()`, but that can be done later or even not at all.

Could you have a look at 95bf532ef143b572a539cc5c4e1c743c9d4aced3? I'm pretty sure this is correct and in my testing it also works.

oroulet:
Diff is OK but you need to rebase your PR. Better to let you do it, you are the last one who looked at that code and it looks like you understand how that works

georgschoelly:
Done. The commits are even a bit clearer now.

PR #261, opened 2016-07-30
Instatiate support for more types - FIX

### LINKED ISSUE #260: [Server] Instantiate - Localized Text
Hi, 

I import a XML that defines a custom objectType which has a variable that use a custom data type (e.g. Enum) which has a LocalizedText referenced:

```
MyObjectType -> MyVariable ( Datatype=MyCustomEnum)
```

Now I want to initiate a object of type `MyObjectType` using `node.add_object(idx, name, my_object_type.nodeid)`.

This works, but when I take look with UaExpert, I see that for each node that had a `UAVariable DataType="LocalizedText` node referenced to in the xml, I see two nodes in UaExport.

Example XML:

see #258 

UaExport Screenshot:

screenshot

maljac:
Fixes #260

### DIFF
--- a/opcua/common/instantiate.py
+++ b/opcua/common/instantiate.py
@@ -44,8 +44,8 @@ def _instantiate_node(server, parentid, rdesc, nodeid, bname, recursive=True):
     """
 
     addnode = ua.AddNodesItem()
-    addnode.RequestedNewNodeId = nodeid 
-    addnode.BrowseName = bname 
+    addnode.RequestedNewNodeId = nodeid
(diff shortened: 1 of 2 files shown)


oroulet:
Could you add tests for that kind of things? so we do not break it many times?

maljac:
Yes, I will try in the future. Sorry!

oroulet:
Great! and thanks for your helps. By the way what kind of spec do you write, I am interested for my work. I work a lot with robots cells and CNC. my adress is      o l i v i e r at sintef.no

PR #263, opened 2016-07-31
Improve enums

georgschoelly:
Depends on #262. Only the last two commits are for this pull request.

The changes can be seen here: https://github.com/georgschoelly/python-opcua/compare/fix_event_notifier_bugs...georgschoelly:improve_enums


oroulet:
good cleanup, thanks

PR #267, opened 2016-08-02
[Server] Import XML - Respect depency of nodes (ParentNodeId)

### LINKED ISSUE #258: [Server] Import XML - Dependency Order
Hi, 

the current implementation of the XML Importer / Parser does not take into account the dependencies of the nodes defined in the xml. Currently it tries to create nodes according to the order of the XML elements.

Workaround: Import the files a second time.

Feature proposal: Respect the dependency tree defined in the XML nodes while importing.

maljac:
Fixes #258

### DIFF
--- a/opcua/common/utils.py
+++ b/opcua/common/utils.py
@@ -87,9 +87,10 @@ class Buffer(object):
 
 class SocketWrapper(object):
     """
-    wrapper to make it possible to have same api for 
+    wrapper to make it possible to have same api for
     normal sockets, socket from asyncio, StringIO, etc....
     """
+
     def __init__(self, sock):
         self.socket = sock
 
--- a/opcua/common/xmlimporter.py
+++ b/opcua/common/xmlimporter.py
@@ -20,6 +20,9 @@ class XmlImporter(object):
         self.logger.info("Importing XML file %s", xmlpath)
         self.parser = xmlparser.XMLParser(xmlpath, act_server)
         for node in self.parser:
+            print(node)
+            print(dir(node))
+            print(node.parent)
             if node.nodetype == 'UAObject':
                 self.add_object(node)
             elif node.nodetype == 'UAObjectType':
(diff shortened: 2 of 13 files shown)


maljac:
Currently a UnitTest is missing for this case.

Also, it must be noted, that the sorting stuff currently works only if:

1) Namespaces are defined in the XML
2) NodeIds must be FQ, thus `i=30001` does not work (and it also should not be used I guess). The correct form would be to defined a namespace in the xml and define the nodeid as `ns=1;i=30001`.

oroulet:
what happens if it does not work? does the import aborts?
Have you tried to import the standard namespace? (look at InternalServer code to enable it)

maljac:
1. The import works as before: If the parent node id is not found it is logged and the import proceeds.
2. You mean importing the standard namespace via xml?

zerox1212:
You need to make sure it can still import the standard namespace. I think somewhere in the internal server class there is an option to use the pre-compiled namespace, or build it from scratch. I think this is actually what the XML importer was originally written for.

Maybe @oroulet can explain better.

maljac:
Yeah, I already found the place. Had no time to test. But I dont see why it
should not work anymore
Am 03.08.2016 19:34 schrieb "Andrew" <notifications@github.com>:

maljac:
1. Tested XML import of standard namespace via xml: Works fine.
2. Build fails because Python2.7 build is stalled (test work, but after 10 minutes travis shutdown the test...)
3. Unit test for ordering comming soon

maljac:
You are too fast :-) While writing my unit test i found a bug. I will fix
it this evening.
Am 05.08.2016 18:05 schrieb "ORD" <notifications@github.com>:

oroulet:
yes. I always find bugs when writing tests :-) If you do not want me to
merge, flag it in the title as others do.

Issue #146, opened 2016-03-18
Automatic Recconnect


mertz3hack:

destogl:
There was few time discussion about that and it was decided that the highest-level client implementation should take care about that... This you can do by catching the exceptions.

@oroulet Am I right?

KeepAlive should take care that your connection doesn't time out, so the question is where is really your problem. Maybe the server has short time-out...

oroulet:
Automatic reconnect requires quite a lot of code and internal changes.
Since the client would then also need to remember subscriptions and attempt
to reconnect to session, etc,. So I am unsure how to implement this

Rosepeter:
What about adding some kind of callback-interface to the client that informs the application when e.g. timeout exception in the KeepAlive-thread. At the moment the only way to get informed about this, is using the synchronous interface of Client and this is often is not the best choice (e.g. when calling a method).
When you start to use the underlying UaClient or even the UASocketClient with callbacks it seems not to be possible to get informed about timeouts in the KeepAlive-thread. Of course, you may claim that your on your own, when using that stuff. But it would be very helpful to use Client for setting up the connection etc. and then use these low-level interfaces to establish asynchronous communication with the server.
If anybody is interested in such a callback-interface I could provide a basic-implementation.

Issue #106, opened 2016-01-06
Detecting connection loss


bitmarker:
Hello!
Is it possible to detect if the server has been turned off or the connection between the server and client was lost? The only way I found is to subscribe to the server time and use the subscription as "watchdog signal". Is there any way to detect if the session has ended?

alkor:
UaExpert client periodically reads state of server (node i=2259). Probably that requires less network traffic than subscription.

alkor:
Found in OPC UA specs, part 4, 6.5:

> The preferred mechanism for a Client to monitor the connection status is through the keep-alive of the Subscription. A Client should subscribe for the State Variable in the ServerStatus to detect shutdown or other failure states. If no Subscription is created or the Server does not support Subscriptions, the connection can be monitored by periodically reading the State Variable.

oroulet:
Interesting to see than that kind of thing is in spec. Yes it seems to be
the most reliable way to detect connection issues. Many sdk have this
built-in, and we may implement it but it requires to remember a bunch of
state things ( like running subscriptions). We wait anyway for the port of
client to asyncio....

zerox1212:
Maybe this is related. Is it possible to override the server state to force the client to disconnect on it's end? I'm still looking for ways to force all the clients to disconnect from my server.

ronytesler:
"A Client should subscribe for
the State Variable in the ServerStatus to detect shutdown or other failure
states."
OK, but what should we do with the value we get?!
I got 4 as a value, what does that mean?

AndreasHeine:
https://reference.opcfoundation.org/v104/Core/DataTypes/ServerState/

alexrudd2:
Dead link.  It's now https://reference.opcfoundation.org/Core/Part5/v104/docs/12.6

PR #268, opened 2016-08-03
code improvements

georgschoelly:
I've implemented a class that mirrors the alarms of my opc ua server. This involves keeping open a client subscribed to some events. In case the connection gets interrupted, I can just delete all my state and re-connect.

However, for this to work I need a way to detect a connection interruption. There seem to be some more people interested in this: https://github.com/FreeOpcUa/python-opcua/issues/106 (connection loss), https://github.com/FreeOpcUa/python-opcua/issues/146 (automatic reconnect).

I've started an implementation (untested so far) and I would like some feedback. I realise that the KeepAlive thread is probably going away in a future version of python-opcua when you switch to asyncio. However, I feel we could implement it for 1.0 anyway because it's quite important functionality. I would document the fact that this API might not be around to stay.

What do you think of this? Is this the right way?

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -25,19 +25,26 @@ except ImportError:
 class KeepAlive(Thread):
 
     """
-    Used by Client to keep session opened.
+    Used by Client to keep the session open.
     OPCUA defines timeout both for sessions and secure channel
     """
 
     def __init__(self, client, timeout):
+        """
+        :param session_timeout: Timeout to re-new the session
+            in milliseconds.
+        """
         Thread.__init__(self)
         self.logger = logging.getLogger(__name__)
-        if timeout == 0:  # means no timeout bu we do not trust such servers
-            timeout = 360000
-        self.timeout = timeout
+
         self.client = client
         self._dostop = False
         self._cond = Condition()
+        self.timeout = timeout
+
+        # some server support no timeout, but we do not trust them
+        if self.timeout == 0:
+            self.timeout = 3600000 # 1 hour
 
     def run(self):
         self.logger.debug("starting keepalive thread with period of %s milliseconds", self.timeout)
@@ -74,12 +81,14 @@ class Client(object):
 
     def __init__(self, url, timeout=4):
         """
-        used url argument to connect to server.
-        if you are unsure of url, write at least hostname and port
-        and call get_endpoints
-        timeout is the timeout to get an answer for requests to server
-        public member of this call are available to be set by API users
 
+        :param url: url of the server.
+            if you are unsure of url, write at least hostname
+            and port and call get_endpoints
+
+        :param timeout:
+            Each request sent to the server expects an answer within this
+            time. The timeout is specified in seconds.
         """
         self.logger = logging.getLogger(__name__)
         self.server_url = urlparse(url)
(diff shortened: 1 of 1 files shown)


opcua/client/client.py
```diff
@@ -38,27 +38,39 @@ def __init__(self, client, timeout):
...
         self._dostop = True
         with self._cond:
             self._cond.notify_all()
 
+    def register_interruption_callback(self, callback):
+        self._interruption_callbacks.add(callback)
+
+    def _connection_interrupted(self, exception):
+        # copy the set for thread safety
+        for callback in self._interruption_callbacks.copy():
```
maljac: Isn't a lock a better solution? What if you want that only one thread that uses the client call the callbacks?

georgschoelly: If I understand you correctly, you mean something like:
    
    try:
        self.connection_interrupted = threading.Lock()
        self.connection_interrupted.acquire()
        while True:
            // send keepalive messages
    except Exception as e: // uh oh, connection lost
        self.connection_interrupted.release()

(or alternatively release the lock unconditionally)

and in the main thread you would do:

    client.connect()
    client.keepalive.connection_interrupted.aquire() # wait for connection loss
    # connection lost
    # do something about it

I see two problems: (1) the main thread cannot be used to interact with the opc ua client anymore (2) to fix this, one would have to call `connection_interrupted.aquire()` in a thread, in which case the *after connection loss*-code is called in a not-main thread too. Is there anything we gain by this?

maljac: No, I actually meant using Lock object for registering and calling `self._interruption_callbacks` instead of `self._interruption_callbacks.copy()`.

georgschoelly: I've changed it to use Locks, but I think the functionality stayed the same? I do not understand what you mean by *What if you want that only one thread that uses the client call the callbacks?*, sorry! Could you explain a bit more?

I've read up on the atomicity of Python, the previous code was correct for any Python implementation with a GIL (CPython, PyPy), so this code is more correct in any case.

maljac: Yeah, after your comment I read about `set`,and the atomicity of the operations `pop` and `add` too. As you said: due to the GIL we have the implicit lock. 

Please ignore my first comment...

oroulet:
From what I understand from spec, and my own socket experience, I think you have no other solution than polling to detect connection error.  That is what clients like uaexpert seems to do.
From your main thread read a node attribute at the frequency you need. The polling rate in keep alive thread is probably too slow for most cases, it is just mean to keep connection open.

oroulet:
We could change client to offer that service if others think it is useful. But you will need to increase poll rate, maybe use another thread. We will anyway remove threads when we can make move to asyncio

georgschoelly:
I've given up on finding a simple solution, it's easier to implement it in the client application. A more sophisticated approach would take into account the last received packet from the server and only send keep-alive messages when necessary.

This is probably going to be a lot easier when the library uses asyncio in the future. The relevant spec is in *Part 4, paragraph 6.5 Re-establish connection*.

georgschoelly:
This resulted however in some code / documentation improvements which I think might be worthwhile to merge.

georgschoelly:
Can I somehow retrigger a travis check? I believe the tests failed because of some temporary problem.

zerox1212:
Just add some more doc strings and commit again. :)

oroulet:
Or just ignore. Some tests are time dependent, and fail if the container
runs too slow.

PR #270, opened 2016-08-03
Implement XML exporter

zerox1212:
This is for review. Only UAVariable method is partially done as an example.

Please give feedback on this design before I implement the remaining XML export methods.

FYI, I modeled the exporter after the importer, so a lot of junk from that is commented out. I left it there because the XML exporter basically has to reverse whatever the importer functions do so it makes it easier to write the methods.

### DIFF
--- a/examples/server-xmlexporter.py
+++ b/examples/server-xmlexporter.py
@@ -1,0 +1,49 @@
+import sys
+sys.path.insert(0, "..")
+import time
+from collections import OrderedDict
+
+from opcua import ua, Server, instantiate
+from opcua.common.xmlexporter import XmlExporter
+
+
+if __name__ == "__main__":
+
+    # setup our server
+    server = Server()
+    server.set_endpoint("opc.tcp://0.0.0.0:4840/freeopcua/server/")
+
+    # setup our own namespace, not really necessary but should as spec
+    uri = "http://examples.freeopcua.github.io"
+    idx = server.register_namespace(uri)
+
+    # get Objects node, this is where we should put our nodes
+    objects = server.get_objects_node()
+
+    # populating our address space
+    myobj = objects.add_object(idx, "MyObject")
+    myvar = myobj.add_variable(idx, "MyVariable", 6.7)
+    myvar.set_writable()    # Set MyVariable to be writable by clients
+
+    dev = server.nodes.base_object_type.add_object_type(0, "MyDevice")
+    dev.add_variable(0, "sensor1", 1.0)
+
+    mydevice = instantiate(server.nodes.objects, dev, bname="2:Device0001")
+
+    node_list = [dev, mydevice, myobj, myvar]
+
+    # starting!
+    server.start()
+
+    node_set_attributes = OrderedDict()
+    node_set_attributes['xmlns:xsi'] = '"http://www.w3.org/2001/XMLSchema-instance"'
+    node_set_attributes['xmlns:uax'] = 'http://opcfoundation.org/UA/2008/02/Types.xsd'
+    node_set_attributes['xmlns:s1'] = 'http://yourorganisation.org/dataobject/Types.xsd'
+    node_set_attributes['xmlns:xsd'] = 'http://www.w3.org/2001/XMLSchema'
+    node_set_attributes['xmlns'] = 'http://opcfoundation.org/UA/2011/03/UANodeSet.xsd'
+
+    exporter = XmlExporter(server, node_set_attributes)
+    exporter.build_etree(node_list)
+    exporter.write_xml('ua-export.xml')
+
+    server.stop()
--- a/examples/server-xmlexporter.py
+++ b/examples/server-xmlexporter.py
@@ -35,15 +35,7 @@ if __name__ == "__main__":
     # starting!
     server.start()
 
-    node_set_attributes = OrderedDict()
-    node_set_attributes['xmlns:xsi'] = '"http://www.w3.org/2001/XMLSchema-instance"'
-    node_set_attributes['xmlns:uax'] = 'http://opcfoundation.org/UA/2008/02/Types.xsd'
-    node_set_attributes['xmlns:s1'] = 'http://yourorganisation.org/dataobject/Types.xsd'
-    node_set_attributes['xmlns:xsd'] = 'http://www.w3.org/2001/XMLSchema'
-    node_set_attributes['xmlns'] = 'http://opcfoundation.org/UA/2011/03/UANodeSet.xsd'
-
-    exporter = XmlExporter(server, node_set_attributes)
-    exporter.build_etree(node_list)
+    exporter = XmlExporter(server)
     exporter.write_xml('ua-export.xml')
 
     server.stop()
--- a/examples/server-xmlexporter.py
+++ b/examples/server-xmlexporter.py
@@ -36,6 +36,7 @@ if __name__ == "__main__":
     server.start()
 
     exporter = XmlExporter(server)
+    exporter.build_etree(node_list, ['http://myua.org/test/'])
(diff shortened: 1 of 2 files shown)


maljac:
One question: I found some code that sounds like "dump" etc. Is this related to the xml exporter?

opcua/common/xmlexporter.py
```diff
@@ -0,0 +1,328 @@
...
+        # self.server.add_references(refs)
+
+    @staticmethod
+    def _get_xml_nodeid(obj):
+        """
+        Convert a UA NodeId object to a formatted string for XML
+        :param obj:
+        :return:
+        """
+        return 'ns=' + str(obj.nodeid.NamespaceIndex) + ';' + str(obj.nodeid.Identifier)
```
oroulet: simply write nodeid.to_string()

opcua/common/xmlexporter.py
```diff
@@ -0,0 +1,328 @@
...
+
+    @staticmethod
+    def _get_xml_browsename(obj):
+        bn = obj.get_browse_name()
+        return str(bn.NamespaceIndex) + ':' + bn.Name
```
oroulet: same: bname.to_string()

opcua/common/xmlexporter.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+class XmlExporter(object):
+
+    def __init__(self, server, node_set_attrs):
+        self.logger = logging.getLogger(__name__)
+        self.etree = Et.ElementTree(Et.Element('UANodeSet', node_set_attrs))
+        self.server = server
+
+    def export_xml(self, node_list, xmlpath):
```
oroulet: somewhere we also need to export address space... or maybe we just get the address space if the nodeid identifier is not 0

opcua/common/xmlexporter.py
```diff
@@ -0,0 +1,328 @@
...
+        elif node_class is ua.NodeClass.VariableType:
+            self.add_variable_type(node)
+        elif node_class is ua.NodeClass.RefernceType:
+            self.add_reference(node)
+        elif node_class is ua.NodeClass.DataType:
+            self.add_datatype(node)
+        elif node_class is ua.NodeClass.Method:
+            self.add_method(node)
+        else:
+            self.logger.info("Not implemented node type: %s ", node_class)
```
oroulet: ok.  but do a rename of 'add_xxx" :-)

oroulet:
@zerox1212 the UI is now usable, we can start testing you exporter as soon as you have something

zerox1212:
I won't be able to work on it until later this week. Too many things going on in my regular job.

oroulet:
Looks good :-)

opcua/common/xmlexporter.py
```diff
@@ -310,19 +270,20 @@ def _add_refs(self, obj):
...
+    def _add_ref_sub_els(parent_el, refs):
+        refs_el = Et.SubElement(parent_el, 'References')
 
-    def _get_xml_parent(self, obj):
-        return self._get_xml_nodeid(obj.get_parent())
\ No newline at end of file
+        for ref in refs:
+            ref_name = o_ids.ObjectIdNames[ref.ReferenceTypeId.Identifier]
+            ref_forward = str(ref.IsForward)
+            refx_el = Et.SubElement(refs_el, 'Reference', IsForward=ref_forward, ReferenceType=ref_name)
```
zerox1212: When I look at UAExpert exports it doesn't always include "IsForward". How do I know if this is required or not?

oroulet: I think we will need to check attributes values against the default and only export if they are different.

How to get default is another question:
*Hard coded in exporter
* Modified attrs in ui 
* Create ua.ObjectAttributes object and compare ( I think I vote for that one)

oroulet:
I think we will need to check attributes values against the default and only export if they are different.

How to get default is another question:
* Hard coded in exporter
* Modified attrs in ui 
* Create ua.ObjectAttributes  or ua.VariableTypeAttribute depending on noce class and compare attribute from node in address space and value of attributes in the created object ( I think I vote for that one)

examples/server-xmlexporter.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # starting!
+    server.start()
+
+    node_set_attributes = OrderedDict()
+    node_set_attributes['xmlns:xsi'] = '"http://www.w3.org/2001/XMLSchema-instance"'
+    node_set_attributes['xmlns:uax'] = 'http://opcfoundation.org/UA/2008/02/Types.xsd'
+    node_set_attributes['xmlns:s1'] = 'http://yourorganisation.org/dataobject/Types.xsd'
+    node_set_attributes['xmlns:xsd'] = 'http://www.w3.org/2001/XMLSchema'
+    node_set_attributes['xmlns'] = 'http://opcfoundation.org/UA/2011/03/UANodeSet.xsd'
```
oroulet: Why do we need this, can't you put this inside exporter class?

opcua/common/xmlexporter.py
```diff
@@ -0,0 +1,226 @@
...
+
+        Returns:
+        """
+        # try to write the XML etree to a file
+        self.logger.info('Exporting XML file to %s', xmlpath)
+        try:
+            self.etree.write(xmlpath, short_empty_elements=False)
+        except TypeError as e:  # TODO where to find which exceptions etree.write() raises?
+            self.logger.error("Error writing XML to file: ", e)
+
```
oroulet: I better to raise exception than catching

zerox1212:
I started working on exporting object types and I have run into a problem.

What is the easiest way to make sure all children of an object type get exported? Then all instances of that object also need to be searched for children.

I don't know the proper approach here, Do we search for children on every node in our list? Or do we just go through the entire address space (ignoring standard address space)?

examples/server-xmlexporter.py
zerox1212: It was mainly for testing, plus I think UA Expert lets you add custom schemas?. I can move it inside the class later.

oroulet:
Do not export children. Just references. The GUI takes care of keeping track of children

opcua/common/xmlexporter.py
```diff
@@ -0,0 +1,226 @@
...
+class XmlExporter(object):
+
+    def __init__(self, server, node_set_attrs):
+        self.logger = logging.getLogger(__name__)
+        self.etree = Et.ElementTree(Et.Element('UANodeSet', node_set_attrs))
+        self.server = server
+        self.aliases = {}
+
+    def build_etree(self, node_list):
```
oroulet: Somewhere I would like to be able to add a list of namespace to etree too

examples/server-xmlexporter.py
oroulet: Uamodeler you mean. Yes let's move it inside for now

opcua/common/xmlexporter.py
zerox1212: I have not used the namespace stuff. How can I get a list of the namespaces in the sever?

oroulet: Same here, only add a method add_namespaces(list of string and maybe idx). So the users or GUI take care of consistency

zerox1212:
This is about as far as I can go for now. Not sure when I'll get to exporting methods, and I don't know what references are.

While testing with UA Expert I found it exporting a few things that I didn't understand in certain cases, such as "SymbolicName". Maybe someone with more UA experience can polish up the last few things.

can you merge this and test the basic functionality with modeler?

oroulet:
Looks like a good start! Thanks!

zerox1212:
Let me know when modeler GUI is ready. I would like to test it some.

PR #271, opened 2016-08-05
support multiple depths of inheritance for instantiate

bitkeeper:
If you create object type A and you derive a type B from it, then using the method `instantiate` will only create members of of type. 

This PR will use all parents, optional (reference HasModellingRule) elements are skipped.

CommonTests.test_instantiate_1 is also update to test and demonstrate the change.

### DIFF
--- a/opcua/common/instantiate.py
+++ b/opcua/common/instantiate.py
@@ -5,7 +5,6 @@ Instantiate a new node and its child nodes from a node type.
 
 from opcua import Node
 from opcua import ua
-from opcua.common.uaerrors import UaError
 from opcua.common import ua_utils 
 
 
@@ -76,15 +75,15 @@ def _instantiate_node(server, parentid, rdesc, nodeid, bname, recursive=True):
         
         if recursive:
             parents = ua_utils.get_node_supertypes(node_type, includeitself = True)
+            node = Node(server, res.AddedNodeId)
+
+
+                
             for parent in parents:
                 descs = parent.get_children_descriptions(includesubtypes=False)
                 for c_rdesc in descs:
-                    #TODO: smells, is there a better way to test if a browse name is already present ?                    
                     # skip items that already exists, prefer the 'lowest' one in object hierarchy
-                    node = Node(server, res.AddedNodeId)
-                    try:
-                        node.get_child(c_rdesc.BrowseName) 
-                    except UaError as e:
+                    if not ua_utils.is_child_present(node, c_rdesc.BrowseName):                    
                         _instantiate_node(server, res.AddedNodeId, c_rdesc, nodeid=ua.NodeId(namespaceidx=res.AddedNodeId.NamespaceIndex), bname=c_rdesc.BrowseName)
                     
         return Node(server, res.AddedNodeId)
--- a/opcua/common/instantiate.py
+++ b/opcua/common/instantiate.py
@@ -76,9 +76,6 @@ def _instantiate_node(server, parentid, rdesc, nodeid, bname, recursive=True):
         if recursive:
             parents = ua_utils.get_node_supertypes(node_type, includeitself = True)
             node = Node(server, res.AddedNodeId)
-
-
-                
             for parent in parents:
                 descs = parent.get_children_descriptions(includesubtypes=False)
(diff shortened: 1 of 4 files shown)


opcua/common/instantiate.py
```diff
@@ -5,8 +5,42 @@
...
+    """
+    basetypes = []
+    parents = node.get_referenced_nodes(refs=ua.ObjectIds.HasSubtype, direction=ua.BrowseDirection.Inverse, includesubtypes=True)
+    if len(parents) != 0:  
+        #TODO: Is it possible to have multiple subtypes ? If so extended support for it
+       basetypes.append(parents[0])  
+       basetypes.extend( _get_sub_types(parents[0]) )
+       
+    return basetypes
+
```
oroulet: there is already such  a function in common/events. Can you check/merge/replace and move resulting function to common/ua_utils.py?

oroulet: and call is get_node_xxx it more explicit

bitkeeper: @oroulet  I will move the function to ua_utils an rename it to `get_node_supertypes`. The name get_sub_types was misleading because it was about the super types and not the subtypes.

Concerning the merge/replace; this is no applicable due the difference in functionality. But I will move also the events.py `get_node_subtypes `to ua_utils.py.

opcua/common/instantiate.py
```diff
@@ -51,32 +53,49 @@ def _instantiate_node(server, parentid, rdesc, nodeid, bname, recursive=True):
...
+        res = server.add_nodes([addnode])[0]
+        
+        if recursive:
+            parents = ua_utils.get_node_supertypes(node_type, includeitself = True)
+            for parent in parents:
+                descs = parent.get_children_descriptions(includesubtypes=False)
+                for c_rdesc in descs:
+                    #TODO: smells, is there a better way to test if a browse name is already present ?                    
+                    # skip items that already exists, prefer the 'lowest' one in object hierarchy
+                    node = Node(server, res.AddedNodeId)
```
oroulet: create the node outside of loop, you are recreating it many times!. Then add a comment to explain what you are doing like 'checking if child node is already instantiated'. Tok me some time to understand why you did that. Maybe split method and give explicit name.like "is_child_instantiated(node)"

opcua/common/instantiate.py
```diff
@@ -76,15 +75,15 @@ def _instantiate_node(server, parentid, rdesc, nodeid, bname, recursive=True):
         
         if recursive:
             parents = ua_utils.get_node_supertypes(node_type, includeitself = True)
+            node = Node(server, res.AddedNodeId)
+
+
+
```
oroulet: no more than one space. Try to run pychecker and/or autopep8 on your code

opcua/common/ua_utils.py
```diff
@@ -143,3 +144,19 @@ def _get_node_supertypes(node):
...
+
+def is_child_present(node, browsename):
+    """
+    return if a browsename is present a child from the provide node
+    :param node: node wherein to find the browsename
+    :param browsename: browsename to search
+    :returns returne True if the browsename is present else False 
+    """
+    is_present = True
+    #TODO: Is there a better way to test if a browse name is already present then with an exception?
```
oroulet: yes do one query: node.get_children_descriptions(xxx):
then check if your browse name is in the response

opcua/common/ua_utils.py
```diff
@@ -152,11 +152,12 @@ def is_child_present(node, browsename):
...
-    
+    is_present = False
+
+    child_descs = node.get_children_descriptions()
+    for child_desc in child_descs:
+        if child_desc.BrowseName == browsename:
+            is_present = True
+            break
+
     return is_present
```
oroulet: I think it is cleaner to just write:
if xxx:
    return True
return False

thus you avoid one variable and a breal. Then I will merge :-)

bitkeeper: I'm not a fan of multiple return statements, but it is your project :-)  So the requested change is made.

oroulet:
Everything is compromise and multiple returns is one tool..

PR #272, opened 2016-08-05
Fix to avoid circular import, including remove of *args

bitkeeper:
More clean construction then the current one.

### DIFF
--- a/opcua/__init__.py
+++ b/opcua/__init__.py
@@ -3,27 +3,6 @@ Pure Python OPC-UA library
 """
 
 from opcua.common.node import Node
-from opcua.common.manage_nodes import create_folder
-from opcua.common.manage_nodes import create_object
-from opcua.common.manage_nodes import create_variable
-from opcua.common.manage_nodes import create_object_type
-from opcua.common.manage_nodes import create_variable_type
-from opcua.common.manage_nodes import create_reference_type
-from opcua.common.manage_nodes import create_data_type
-from opcua.common.manage_nodes import create_property
-from opcua.common.manage_nodes import create_method
-from opcua.common.methods import call_method
-# FIXME: ugly hack to avoid circular import issue and still split file
-Node.add_folder = create_folder
-Node.add_object = create_object
-Node.add_variable = create_variable
-Node.add_property = create_property
-Node.add_method = create_method
-Node.call_method = call_method
-Node.add_data_type = create_data_type
-Node.add_object_type = create_object_type
-Node.add_variable_type = create_variable_type
-Node.add_reference_type = create_reference_type
 
 from opcua.common.methods import uamethod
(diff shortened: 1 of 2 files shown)


georgschoelly:
Is there a reason not to do simple assignments? Instead of

    class Node(object):
        ....
        def add_folder(self, nodeid, bname):
            return  opcua.common.manage_nodes.create_folder(self, nodeid, bname)

we could do:

    class Node(object):
        ...
        add_folder = opcua.common.manage_nodes.create_folder

and thinking a bit more about it, why don't we do that programatically? It's possible to create a [class decorator](http://www.ianbicking.org/blog/2007/08/opening-python-classes.html) which could be used like this: 

```
@extend(opcua.common.node.Node)
class Node(object):
    def some_method(self, args):
        pass
```

oroulet:
yes originally it was an assignment until we got a circular import issue with python2. Does it work? If it works, it is the cleanest thing

maljac:
I like the proposal from @georgschoelly. The only thing that is not so nice is probably the missing IDE support when using the `@extend` mechanism.

georgschoelly:
Another possibility:

~~~
class _NodeBasics(object):
    def method1(self):
        pass

class _NodeCreation(object):
    def method2(self):
         pass

class Node(_NodeBasics, _NodeCreation):
    pass
~~~

With some code such that `_NodeBasics` and `_NodeCreation` cannot be instantiated and that checks for methods / instance variables defined in both.

georgschoelly:
We could also do some subclassing magic:

oroulet:
@georgschoelly last proposition seems a bit overkill. the only issue here is that I did not want to have a file which is too big. I would prefer assignment, this is the solution with least code and that should not break anything. Does assignment work now?

oroulet:
otherwise I will just merge your PR

georgschoelly:
I'll have a look if it works with simple assignments, I should be able to do that within the next hour.

georgschoelly:
I wasn't able to get simple assignments to work and I currently don't have the time for it. I've improved on it slightly by using `*args*` and `**kwargs`, this ensures that if we change the methods in `manage_nodes.py`, we don't need to change it in `node.py` as well.

https://github.com/FreeOpcUa/python-opcua/pull/274

oroulet:
The issue with args and kwargs is that they hide what arguments the method are expecting to users. so I think I will merge bitkeeper solution.

bitkeeper:
Are we not making the problem bigger than it is ?
Just restoring the items is already cleaner then the current situation. Easy to read/maintain, IDE friendly. And it is the situation then we already had for sometime.

Maybe create a new issue to reduce the fat interface of node (take also the other functions of node in account). I believe that subclassing (like indicated by @georgschoelly ) is in the end the way to go compared to extend and simple assigments.

PR #275, opened 2016-08-09
add exception classes for all bad status codes

georgschoelly:
This enables us to catch status codes directly like this:

    try:
        raise UaStatusCodeError(0x80010000)
    except BadUnexpectedError:
        # do something


opcua/common/uaerrors/_auto.py
```diff
@@ -0,0 +1,596 @@
+#AUTOGENERATED!!!
```
oroulet: other files are names ***_auto.py, so maybe we should name this file uaerrors_auto.py. or rename the others if _xx is common convention.

oroulet:
Code looks good. just need to fix error

georgschoelly:
Done. I'm somewhat embarassed not having seen that, but at least it was caught by my own tests. :-)

opcua/common/uaerrors/_auto.py
georgschoelly: I do not feel strongly one way or another. I think we could argue that in the other cases we have a different module layout:

`ua` imports `ua.uaprotocol_auto` and `ua.uaprotocol_hand`, in this case there is no parent module that imports both, so we'd end up with 3 siblings:

   ua.common.uaerrors
   ua.common._uaerrors_hand
   ua.common._uaerrors_auto

georgschoelly: Whether one is a common convention, I don't know.

Issue #152, opened 2016-04-01
client PublishRequest timeout


oroulet:
check what timeout we send publishrequest with (9e8). Ideally we should use 0, try 0 again. Otherwise we need to use thread to resend publish requests that have timeouted on server.
This is critical if no event happen within timeout period

This issue is solved on server side.

georgschoelly:
The spec Part 4, 7.28 states:

> The Server shall check the `timeoutHint` parameter of a `PublishRequest` before processing a `PublishResponse`. If the request timed out, a `Bad_Timeout` Service result is sent and another `PublishRequest` is used.

Doesn't that just mean that we can catch `BadTimeout` in the publish callback handler and re-send another publish request?

oroulet:
@georgschoelly yes, might be a solution. Although using 0 seems even better... if servers support it

georgschoelly:
I created a quick pull request, but I don't have a setup ready to test this. Do you have something where you could easily test it? If not, I'm going to give it a try later.

destogl:
After #282 this should be solved?!

georgschoelly:
@destogl: Not really. The code still uses `9e8` as the timeout. As oroulet says, `0` (= no timeout) would be better.  I don't know what the problem was with `0`, so I don't know if we can just replace `9e8` with `0`.

oroulet:
I think some servers had issues with 0. But we can try again, so we follow correctly spec. Send pr

ramirahikkala:
I'm working with Festo PLC and it has an OPC UA server that doesn't allow 9e8 as a timeout but works with 0.

PR #282, opened 2016-08-12
#152 re-send publish request on timeout

georgschoelly:
Maybe part of a fix for #152

### DIFF
--- a/opcua/client/ua_client.py
+++ b/opcua/client/ua_client.py
@@ -10,7 +10,7 @@ from functools import partial
 
 from opcua import ua
 from opcua.common import utils
-from opcua.common.uaerrors import UaError
+from opcua.common.uaerrors import UaError, BadTimeout
 
 
 class UASocketClient(object):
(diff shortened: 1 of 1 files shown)


oroulet:
have you tested it? should be possible if you set timeout very low in client and another server

georgschoelly:
Not yet, but I will do so soon.

opcua/client/ua_client.py
```diff
@@ -400,23 +400,32 @@ def publish(self, acks=None):
...
+            self._uasocket.check_answer(data, "while waiting for publish response")
+        except BadTimeout: # Spec Part 4, 7.28
+            self.publish()
+
         try:
             response = ua.PublishResponse.from_binary(data)
             self.logger.debug(response)
         except Exception:
             self.logger.exception("Error parsing notificatipn from server")
             self.publish([]) #send publish request ot server so he does stop sending notifications
```
georgschoelly: Does anyone know in which cases this happens? It seems to me that catching `BadTimeout` has we now do above is the only case re-sending a publish request is appropriate.

oroulet: not that I know..

georgschoelly:
I just tested it and works as intended.

oroulet:
OK, our server does not support it, but it does not care about timeout on requests either....

PR #289, opened 2016-08-21
add standard address space part 14

oroulet:


oroulet:
adding it would slow down startup time

PR #311, opened 2016-10-05
Export Argument Extension Object

zerox1212:
Improve Extension Object support
Implement exporter for ua.Argument

### DIFF
--- a/opcua/common/xmlexporter.py
+++ b/opcua/common/xmlexporter.py
@@ -130,7 +130,7 @@ class XmlExporter(object):
         self._add_sub_el(node_el, 'DisplayName', displayname)
         if desc not in (None, ""):
             self._add_sub_el(node_el, 'Description', desc.decode('utf-8'))
-        #FIXME: add WriteMask and UserWriteMask
+        # FIXME: add WriteMask and UserWriteMask
         return node_el
 
     def add_etree_object(self, node):
@@ -155,6 +155,12 @@ class XmlExporter(object):
 
     def add_variable_common(self, node, el):
         dtype = node.get_data_type()
+
+        # FIXME hack because get_data_type() has issues
+        if dtype.NamespaceIndex > 50:
+            dtype.Identifier = dtype.NamespaceIndex
+            dtype.NamespaceIndex = 0
+
         if dtype.Identifier in o_ids.ObjectIdNames:
             dtype_name = o_ids.ObjectIdNames[dtype.Identifier]
             self.aliases[dtype_name] = dtype.to_string()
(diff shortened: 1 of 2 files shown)


opcua/common/xmlexporter.py
```diff
@@ -276,11 +282,16 @@ def value_to_etree(el, dtype_name, dtype, node):
 
 def _value_to_etree(el, dtype_name, dtype, val):
     if isinstance(val, (list, tuple)):
-        list_el = Et.SubElement(el, "uax:ListOf" + dtype_name)
-        for nval in val:
-            _value_to_etree(list_el, dtype_name, dtype, nval)
+        if type(dtype.Identifier) is int and dtype.Identifier > 21:  # this is a list of ExtensionObjects:
```
oroulet: can dtype.Identifier be something else than int? but ok

opcua/common/xmlexporter.py
```diff
@@ -155,6 +155,12 @@ def add_etree_object_type(self, node):
 
     def add_variable_common(self, node, el):
         dtype = node.get_data_type()
+
+        # FIXME hack because get_data_type() has issues
+        if dtype.NamespaceIndex > 50:
```
zerox1212: Use `parent.add_method()` and step through the code. Eventually `_create_method()` is called in manage_nodes.py where the InputArgument and OutputArgument properties are automatically created. When the properties are created the Identifier ends up in the `ns` field.

opcua/common/xmlexporter.py
zerox1212: I don't know. This check was already here from someone else, I just made it work for Methods because I don't fully understand extension objects.

opcua/common/xmlexporter.py
```diff
@@ -276,11 +282,16 @@ def value_to_etree(el, dtype_name, dtype, node):
 
 def _value_to_etree(el, dtype_name, dtype, val):
     if isinstance(val, (list, tuple)):
-        list_el = Et.SubElement(el, "uax:ListOf" + dtype_name)
-        for nval in val:
-            _value_to_etree(list_el, dtype_name, dtype, nval)
```
zerox1212: Do you mean make the "ListofX" element in `value_to_etree`? You typed `_value_etree` twice.

oroulet: the "ListOfX" is already written, but inside _value_to_etree you can test for extensionobject type and call you new methods from there.... should work.. if I did not forget anything...

zerox1212: The problem with that is the XML will output "ListOfArgument" instead of "ListOfExtensionObject" because the test `if isinstance(val, (list, tuple)):` will pass. Then the XML isn't valid.

zerox1212: The above code doesn't work. You will need to show me how to do it.

oroulet: your code above looks good. why doesn't it work?

oroulet:
Wait a little. I saw I wrote some code to export any ExtensionObjects that
is very close to work...

zerox1212:
My solution is working locally. I will wait for your code and no longer work on mine.

oroulet:
pushed a more general solution to exporting extension objects. Let me know how it works in your case

zerox1212:
I took a look and will test it. Soon I will close this PR.

oroulet:
Still broken. I added some tests and found a bunch of bugs I need to fix...

zerox1212:
For arguments it almost works. Only problem I see is that it makes the "uax:Description" element wrong. Trying to fix.

zerox1212:
I added PR to fix the only problem I had with your implementation. I'm closing this because your way is better and already merged.

Issue #297, opened 2016-08-30
[Server/Noob Question] How do I get methods to work in my custom types?


stillero:
Hello!

I recently started working on OPC UA and Python, so I'm still trying to wrap my head around most things, so please, bear with me. This might be a pretty stupid question, but I just can't get it right, so I need to ask anyway - how do I get methods to work in my custom types?

Here's a short the example, based on the "server-minimal" example:

![smoothexample](https://cloud.githubusercontent.com/assets/15153569/18081378/257c2a70-6e9a-11e6-9443-36c874bd4b90.PNG)

Whenever I call the method via Test Client, it always returns "BadNothingToDo".
But I know this method works just fine! If I add it to an object, like the SmoothSample above, the method runs successfully when called by the client. But that can't be the only solution, right?
I know I'm doing something wrong...

Please help me out!

Daniel

zerox1212:
You are creating a method node that doesn't have a UaMethod. You need to use a decorator when defining ```SmoothMethod1```

You can see this in server-method example. https://github.com/FreeOpcUa/python-opcua/blob/master/examples/server-methods.py
```python
# method to be exposed through server
def func(parent, variant):
    ret = False
    if variant.Value % 2 == 0:
        ret = True
    return [ua.Variant(ret, ua.VariantType.Boolean)]


# method to be exposed through server
# uses a decorator to automatically convert to and from variants

@uamethod
def multiply(parent, x, y):
    print("multiply method call with parameters: ", x, y)
    return x * y
```

stillero:
Thanks for the Input, I've tried your solution, but still no change in the outcome. Did I implement it wrong?

![smoothexample2](https://cloud.githubusercontent.com/assets/15153569/18089287/7904275e-6ebf-11e6-89c8-0f780ef7de0c.PNG)

zerox1212:
Your second attempt is wrong, because you have to declare method stuff manually.

```python
inargx = ua.Argument()
    inargx.Name = "x"
    inargx.DataType = ua.NodeId(ua.ObjectIds.Int64)
    inargx.ValueRank = -1
    inargx.ArrayDimensions = []
    inargx.Description = ua.LocalizedText("First number x")
    inargy = ua.Argument()
    inargy.Name = "y"
    inargy.DataType = ua.NodeId(ua.ObjectIds.Int64)
    inargy.ValueRank = -1
    inargy.ArrayDimensions = []
    inargy.Description = ua.LocalizedText("Second number y")
[...]

PR #312, opened 2016-10-05
Server link method

zerox1212:
To be used after instantiating from a custom type which has a method node.
See: https://github.com/FreeOpcUa/python-opcua/issues/297


zerox1212:
Sorry, I don't know how to use git very well. I thought this branch would be separate link method commit from other commits.

PR #313, opened 2016-10-05
Add link_method to server

zerox1212:
shortcut to linking python functions to UA Method nodes, for example
after instantiating an object from object type which has a method.

To be used after instantiating from a custom type which has a method node.
See: #297

### DIFF
--- a/opcua/server/server.py
+++ b/opcua/server/server.py
@@ -416,3 +416,6 @@ class Server(object):
 
     def unsubscribe_server_callback(self, event, handle):
         self.iserver.unsubscribe_server_callback(event, handle)
+
+    def link_method(self, node, callback):
+        node.server.add_method_callback(node.nodeid, callback)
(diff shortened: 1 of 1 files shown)


opcua/server/server.py
```diff
@@ -416,3 +416,6 @@ def subscribe_server_callback(self, event, handle):
 
     def unsubscribe_server_callback(self, event, handle):
         self.iserver.unsubscribe_server_callback(event, handle)
+
+    def link_method(self, node, callback):
+        node.server.add_method_callback(node.nodeid, callback)
```
oroulet: Looks a bit strange. Better to call
del.iserver.add_method_callback....

zerox1212: I will fix it.

zerox1212: Are you sure your idea will work?

The `add_method_callback` method is part of InternalSession, not InternalServer.

oroulet:
Self.isession then. ?

zerox1212:
I think this should be ok now?

PR #314, opened 2016-10-06
added connection loss polling to Client

AlexejStukov:
Simple implementation of connection loss polling using the KeepAlive thread (see #106). Server will be polled every `polling_interval` ms if `polling_interval` is >0.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -29,7 +29,7 @@ class KeepAlive(Thread):
     OPCUA defines timeout both for sessions and secure channel
     """
 
-    def __init__(self, client, timeout):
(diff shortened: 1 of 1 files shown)


oroulet:
What is the goal of this? Do some servers require extra polling?

AlexejStukov:
If you have a "bad" connection and want to know if the server is online or not (e.g. every 60 seconds). This does not renew the connection and so is more light on the resources.

oroulet:
As far as I can see this is only the thread that knows if connection has
gone wrong. How can you use such a thing?

AlexejStukov:
Hm, very good point and I don't see an easy way to get that exception to the main thread...

PR #316, opened 2016-10-07
Prevent the UASocketClient from hanging

AlexejStukov:
Added a timeout to the socket connection that prevents the ``UASocketClient`` from hanging on  ``_receive`` in ``_run``.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -79,7 +79,7 @@ class Client(object):
     which offers the raw OPC-UA services interface.
     """
 
-    def __init__(self, url, timeout=4):
+    def __init__(self, url, timeout=4, socket_timeout=10):
         """
 
         :param url: url of the server.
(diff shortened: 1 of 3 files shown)


oroulet:
Maybe should we use another timeout variable. Add a socket_timeout to class.
What should be default value of such timeout? Not sure...

AlexejStukov:
Is there a size limitation on the received packages?

oroulet:
A message has a max size if I remember well. You need to look at member
variables in headers or connection settings....

AlexejStukov:
The only thing i can find is ``self.MaxMessageSize = 0  # No limits`` and the likes... .

oroulet:
Yes I tried to set everything to no limits. What is your issue?

AlexejStukov:
I was searching for a size limitation in the sources to estimate a reasonable timeout. 
Shouldn't the package size be restricted by tcp? Then we could assume 64kb. So 10 seconds might be ok?

oroulet:
I think timeout is only user of establishing and stopping connection, so
packet size does not influence this. Something between 2 and 10 is probably
ok

AlexejStukov:
Ok, then I'll leave it at 10 seconds. If somebody needs a longer or shorter time he can change it easily.

Issue #317, opened 2016-10-09
Timeout Error with Imported Methods


zerox1212:
This one has me very confused.
I make a very basic address space in OPCUA Modeler, since it's running a server in the background I connect to it via UA Expert.
Modeler:
![image](https://cloud.githubusercontent.com/assets/13090472/19223531/bd65c3f2-8e37-11e6-8f84-45485ac1f8ec.png)
UA Expert connected:
![image](https://cloud.githubusercontent.com/assets/13090472/19223534/cedd4934-8e37-11e6-89a6-814bfa0b96b2.png)

This has no issues, clicking on the Input or Output Arguments returns node information as expected. I can even call the method (even though it just returns nothing to do).

I then export these 3 nodes to XML to method.xml. Then I start a server and call `server.import_xml("method.xml")`. The importer migrates everything to namespace 3, which should be fine. I connect UA Expert and I can browse the object and method nodes with no problems.
Browsing upper nodes with UA Expert:
![image](https://cloud.githubusercontent.com/assets/13090472/19223556/4b22b84e-8e38-11e6-8281-749797184b80.png)

However, as soon as I click on the InputArgument or OutputArgument nodes, the server crashes with this:
```
ERROR:opcua.server.binary_server_asyncio:Exception raised while parsing message from client, closing
Traceback (most recent call last):
  File "..\opcua\server\binary_server_asyncio.py", line 85, in _process_data
    ret = self.processor.process(hdr, buf)
  File "..\opcua\server\uaprocessor.py", line 87, in process
    return self.process_message(msg.SecurityHeader(), msg.SequenceHeader(), msg.body())
  File "..\opcua\server\uaprocessor.py", line 108, in process_message
    return self._process_message(typeid, requesthdr, algohdr, seqhdr, body)
  File "..\opcua\server\uaprocessor.py", line 184, in _process_message
    self.send_response(requesthdr.RequestHandle, algohdr, seqhdr, response)
  File "..\opcua\server\uaprocessor.py", line 41, in send_response
    response.to_binary(), message_type=msgtype, request_id=seqhdr.RequestId, algohdr=algohdr)
  File "..\opcua\ua\uaprotocol_auto.py", line 8933, in to_binary
    packet.append(fieldname.to_binary())
  File "..\opcua\ua\uatypes.py", line 996, in to_binary
    packet.append(self.Value.to_binary())
  File "..\opcua\ua\uatypes.py", line 837, in to_binary
[...]

PR #319, opened 2016-10-10
Fix getting node id identifier - importer fixes

zerox1212:
Please review! It took me a long time to track this down and I don't know if this solution is correct.

The original code would make a second NodeId object instead of just getting the Identifier, but perhaps that is needed in some other case???

### DIFF
--- a/opcua/common/xmlimporter.py
+++ b/opcua/common/xmlimporter.py
@@ -26,7 +26,9 @@ def ua_type_to_python(val, uatype):
 
 def to_python(val, obj, attname):
     if isinstance(obj, ua.NodeId) and attname == "Identifier":
-        return ua.NodeId.from_string(val)
+        k, v = val.split("=", 1)
+        identifier = int(v)
+        return identifier
     else:
         return ua_type_to_python(val, obj.ua_types[attname])
 
--- a/opcua/common/xmlimporter.py
+++ b/opcua/common/xmlimporter.py
@@ -15,13 +15,15 @@ def ua_type_to_python(val, uatype):
         return int(val)
     elif uatype in ("String"):
         return val
+    elif uatype in ("LocalizedText"):
+        return ua.LocalizedText(val)
     elif uatype in ("Bytes", "Bytes", "ByteString", "ByteArray"):
         if sys.version_info.major > 2:
             return bytes(val, 'utf8')
         else:
             return val
     else:
-        raise Exception("uatype nopt handled", uatype, " for val ", val)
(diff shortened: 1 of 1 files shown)


zerox1212:
Example of what happens in this issue: https://github.com/FreeOpcUa/python-opcua/issues/317

opcua/common/xmlimporter.py
```diff
@@ -15,18 +15,22 @@ def ua_type_to_python(val, uatype):
...
-        raise Exception("uatype nopt handled", uatype, " for val ", val)
+        raise Exception("ua type not handled", uatype, " for val ", val)
 
 
 def to_python(val, obj, attname):
     if isinstance(obj, ua.NodeId) and attname == "Identifier":
-        return ua.NodeId.from_string(val)
+        k, v = val.split("=", 1)
+        identifier = int(v)
+        return identifier
```
oroulet: this is unfortunaltely wrong. For some strange reason what is written as identifier is the string version of a node....

zerox1212:
What fix do you propose?

The problem is that when the extension object is instantiated (via this call `ext = getattr(ua, obj.objname)()`) the init creates a blank NodeId object. Then later `ua.NodeId.from_string(val)` gets called which returns a NumrericNodeId, but `setattr` seems to be assigning the __str__ value instead of the NumericNodeId object. This code is a little confusing for me to follow because there is a lot of builtin overrides, etc. Maybe the problem is the original NodeId from init is frozen and can't be changed?

oroulet:
pushed a fix there: https://github.com/FreeOpcUa/python-opcua/pull/323

oroulet:
btw I completely agree that the xml exporter and importer code is almost impossible to understand and hard to debug. I do not know why. Maybe it the nature of XML...

zerox1212:
I tested your fix for the node id and it's working. Thanks.

Your fix does not include support for import extension object descriptions. Will you add that in your own PR or use this one?

zerox1212:
I'm going to close this as it is no longer needed. However please add support for importing extension object descriptions (localized text) as soon as you get time. Thanks.

oroulet:
Description works in the tests in tests_xml.py, look at code. is it broken
for you?

oroulet:
@zerox1212 I added one more test to specifically check for description attribute of args in methods and it works. Do you still have issues? Maybe we still have a bug in xml format of description attribute?

zerox1212:
I will try and open issue if needed. I'm on different project for the moment.

PR #320, opened 2016-10-10
Export localized text objects correctly - exporter fixes

zerox1212:
Don't create a Locale xml element, just export the Description as text.
Small typo fixes.


oroulet:
It seems localized text are sometimes exported as a string and sometimes as an object with "enc" en "text" members... ideally we should support both,..

zerox1212:
I didn't read the spec. Other modelers seem to just export the string.

My guess is you need to export the object if you are using XML protocol or something.

oroulet:
In my tests with uamodeller, so localized strings where written as an
object in argument....

zerox1212:
Strange, with mine it only wrote the "text" into the description tag. No other parts of the LocalizedText object were created. Perhaps it has to do with system language?

Anyways I will close this and maybe just correct importer to handle the localized text object.

PR #322, opened 2016-10-10
migrate browsename namespace when importing to correct index

bitkeeper:
test included

### DIFF
--- a/opcua/common/xmlparser.py
+++ b/opcua/common/xmlparser.py
@@ -44,8 +44,8 @@ class NodeData(object):
 
         # referencetype
         self.inversename = ""
-        self.abstract = False 
-        self.symmetric = False 
+        self.abstract = False
+        self.symmetric = False
 
         # datatype
         self.definition = []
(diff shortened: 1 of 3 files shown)


oroulet:
thanks for a very complete PR. I am waiting a bit to merge that. I discovered that the namespace import stuff is broken. First we should not increase index in xmlparser but in xmlimporter. Then it seems to increase index even when namespaces already exit in address space....but I do not have time to look at that now...
In parser we should only parse
Then in importer we sould check if namespace exist, then cache the index and for every node to import, check to what namespace uri it refers, not simply increase its number!

bitkeeper:
@oroulet Something different; running the tests I noticed that the server isn't cleared between the tests. Starting the server before each test probably takes to much time. Is there a single method to delete all non standard nodes ? Or maybe use a combi of dump_address_space/load_address_space ?

Sometimes this influences the tests. For example `test_xml_export` breaks when running `tests/tests.py TestServer` , while `tests/tests.py TestServer.test_xml_export` runs fines.

oroulet:
Yes it is to expensive to to start new server at reach run. And in this case this helped to uncover another big so this is probably fine... There is a hack to make test with in master but we need to fix import of namespace asap

bitkeeper:
@oroulet   Could you open an separate issue about the broken namespace import ? I'm not sure I understand the problem, because importing multiple files within the same namespace just works fine here.

oroulet:
@bitkeeper done

oroulet:
Can you rebase then I can merge.

bitkeeper:
rebase completed

oroulet:
well done!
btw we should also migrate reference NodeId!. try to remember it if you move code.

Issue #321, opened 2016-10-10
command line test tool for OPC-UA method call missing in python-opcua


SCHUNK-Osswald:
Changing values in our (proprietary) opc-ua server is done via "OPC-UA method calls". For simple test it would thus be helpful to have a "uacall" command line tool, just like "uaread/uawrite/...". 
The rest of the code seems to be prepared for providing this (except for a minor bug in opcua.node.Node.call_method (missing self parameter)).

I feel sufficiently skilled to add this myself, so will try to provide some coding help here. (But please be forgiving, as I'm new to git/github/opcua)

oroulet:
Sounds like a good idea. It should not be difficult..

zerox1212:
If you don't want to learn git, just grab the github client. It works well.

The only thing that sucks about the gui client is it's often hard to synchronize you fork with the original master. In which case I open the GitShell and use the following command:

```
git remote add upstream url/to/original/repo
git fetch upstream
git checkout master
git reset --hard upstream/master
git push origin master --force
```

PR #325, opened 2016-10-11
Added command line support for OPC-UA method calls according to #321

SCHUNK-Osswald:
Finished the command line tool for calling OPC_UA methods as asked for in #321.

Method calling now works on PC (windows 7) and BBB (Debian Linux on ARM) with our proprietary OPC-UA server.

Not sure if this whole git-fork-clone-branch-commit-push stuff worked as expected. (Maybe my first changes ended in master rather than in the add_call_cli branch.)

Feedback / suggestions / corrections welcome.


setup.py
```diff
@@ -7,8 +7,8 @@
 else:
     install_requires = ["python-dateutil"]
 
-setup(name="freeopcua", 
-      version="0.10.17",
+setup(name="freeopcua",
+      version="0.10.17.1",
```
oroulet: no need to increase version here.

oroulet:
thanks for the PR. it looks OK. Btw, may I ask where does SCHUNK has an opc ua server? I have a bunch of schunck hardware, but they tend to use older protocol/connections like serial ;-)

setup.py
SCHUNK-Osswald: OK, sorry. Just did that so that pip performs the update...

SCHUNK-Osswald:
Yes, we tend to use older HW, but even for us serial (RS232) is pretty uncommon nowadays. Most of our modules use CAN or PROFIbus. Newer producst use PROFInet, Ethernet or USB.

And now OPC-UA is used within a research project that we participate in. Unfortunately I guess it will be some time until this reaches buyable SCHUNK products...

BTW: Which products do you use?

opcua/common/node.py
```diff
@@ -525,30 +525,30 @@ def delete(self, delete_references=True):
...
     def add_method(self, *args):
         return opcua.common.manage_nodes.create_method(self, *args)
- 
+
     def add_reference_type(self, parent, nodeid, bname):
         return opcua.common.manage_nodes.create_reference_type(parent, nodeid, bname)
-    
-    def call_method(parent, methodid, *args):
+
+    def call_method(self, parent, methodid, *args):
```
oroulet: This was not an error. Self was called parent her. I admit this was confusing. But you need to remind listener arg if you add self. Currently your pr breaks tests

opcua/tools.py
```diff
@@ -651,3 +651,77 @@ def uahistoryread():
...
+                    break
+
+        if ( method_id is None):
+            # last resort:
+            method_id = ua.NodeId( identifier=args.method )#, namespaceidx=? )#, nodeidtype=?): )
+
+        #print "method_id=%s\nval=%s" % (method_id,val)
+
+        result_variants = node.call_method( parent, method_id, *val )
+        print "resulting result_variants=%s" % result_variants
```
oroulet: No need for parent arg. The node itself is the parent

opcua/common/node.py
SCHUNK-Osswald: I don't know what you mean by 

> But you need to remind listener arg if you add self.

?
If no aditional _parent_ is needed then lets just rename _parent_ to _self_ (in your original version) or remove _parent_ in my version.
But I think naming the first parameter of a member function other than _self_ will always cause trouble.

opcua/tools.py
SCHUNK-Osswald: Yes ok, would be clear if _parent_ parameter was called _self_, as said above.

BTW: What am I supposed to do now (regarding git/pull/push...)? Should I change my add_call_cli branch, then commit, then push, then create pull request? Or what?

oroulet:
Looks like something went wrong with keyboard auto correction. The error
was that the first argument was called parent. Rename it to self, and
remove extra parent argument

oroulet:
@SCHUNK-Osswald Can you finish this?

SCHUNK-Osswald:
Am I done now or am I missing something?
(Still confused by this git workflow...)

oroulet:
Looks good now. But you have one fail left: Then I can merge
SyntaxError: Missing parentheses in call to 'print'

ERROR

test_uawrite (tests_cmd_lines.TestCmdLines) ... Traceback (most recent call last):

  File "tools/uawrite", line 7, in <module>

    from opcua.tools import uawrite

  File "/home/travis/build/FreeOpcUa/python-opcua/tools/../opcua/tools.py", line 721

    print "resulting result_variants=%s" % result_variants

                                       ^

SyntaxError: Missing parentheses in call to 'print'

oroulet:
Thanks for you contribution!

PR #327, opened 2016-10-13
also export namespace of referenced nodes

oroulet:


oroulet:
@bitkeeper export ns of referenced nodes seems to break importer. Can you run test
python tests/tests.py TestServer.test_xml_ns
look at generated xml and tell me if something is wrong? the last namespace is only used in a reference (a parent). I thought you showed me a similar example from uamodeller

oroulet:
@bitkeeper. I found out the hang is due to the node ordering in xmlparser always expect parents to be defined in xml. This is also code (the ordering stuff) that should be moved to importer. And check if parent is in address space. Not xml.

bitkeeper:
@oroulet yes I already expected that `_sort_nodes_by_parentid` also should be moved.

Issue #328, opened 2016-10-13
subscription stack trace in internal_subscription.deadband_callback


bitkeeper:
Today I got a nice stack trace from a python-opcua server  (at bottom of this message).

The problem is isolated to:
`D:\work\python-opcua\opcua\server\internal_subscription.py", line 209, in deadband_callback`

```python
   def deadband_callback(self, values, flt):
        if (values.get_old_value() is not None) or \
                ((abs(values.get_current_value() - values.get_old_value())) > flt.DeadbandValue):
            return True
        else:
            return False
```
When `get values.get_old_value()` is `None` you will get a exception. 
I guess the `not` in the `if` is a typo, but before I provide fix I want to be sure in case of `None` the return value should be `True` ?

```python
ERROR:opcua.server.binary_server_asyncio:Exception raised while parsing message from client, closing
Traceback (most recent call last):
  File "D:\work\python-opcua\opcua\server\binary_server_asyncio.py", line 85, in _process_data
    ret = self.processor.process(hdr, buf)
  File "D:\work\python-opcua\opcua\server\uaprocessor.py", line 87, in process
    return self.process_message(msg.SecurityHeader(), msg.SequenceHeader(), msg.body())
  File "D:\work\python-opcua\opcua\server\uaprocessor.py", line 108, in process_message
    return self._process_message(typeid, requesthdr, algohdr, seqhdr, body)
  File "D:\work\python-opcua\opcua\server\uaprocessor.py", line 320, in _process_message
    results = self.session.create_monitored_items(params)
  File "D:\work\python-opcua\opcua\server\internal_server.py", line 347, in create_monitored_items
    subscription_result = self.subscription_service.create_monitored_items(params)
  File "D:\work\python-opcua\opcua\server\subscription_service.py", line 67, in create_monitored_items
    return self.subscriptions[params.SubscriptionId].monitored_item_srv.create_monitored_items(params)
  File "D:\work\python-opcua\opcua\server\internal_subscription.py", line 69, in create_monitored_items
    result = self._create_data_change_monitored_item(item)
  File "D:\work\python-opcua\opcua\server\internal_subscription.py", line 157, in _create_data_change_monitored_item
    self.trigger_datachange(handle, params.ItemToMonitor.NodeId, params.ItemToMonitor.AttributeId)
  File "D:\work\python-opcua\opcua\server\internal_subscription.py", line 82, in trigger_datachange
    self.datachange_callback(handle, variant)
[...]

PR #329, opened 2016-10-14
fixed internal_subscription deadband_callback exception 

bitkeeper:
Fixed internal_subscription deadband_callback exception when old value are None.
See for more info #328

### DIFF
--- a/opcua/server/internal_subscription.py
+++ b/opcua/server/internal_subscription.py
@@ -205,7 +205,7 @@ class MonitoredItemService(object):
                     self.isub.enqueue_datachange_event(mid, event, mdata.queue_size)
 
     def deadband_callback(self, values, flt):
-        if (values.get_old_value() is not None) or \
+        if (values.get_old_value() is None) or \
                 ((abs(values.get_current_value() - values.get_old_value())) > flt.DeadbandValue):
             return True
         else:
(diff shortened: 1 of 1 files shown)


bitkeeper:
Haha the blame says that you made the change ... 

The following code fragment shows why I think it is the correct value. 

The location ( internal_subscription.py:datachange_callback) where the `deadband_callback` is used, uses a `deadband_flag_pass = True` when no filter is present.
```
  if mdata.filter is not None:
                    deadband_flag_pass = self.deadband_callback(mdata.mvalue, mdata.filter)
                else:
                    deadband_flag_pass = True
```

Then this leads to the assumption when there isn't an old value that `deadband_flag_pass` is also `True`. Any case now it leads to an exception

Orginal code:
```
     if (values.get_old_value() is not None) or \
                ((abs(values.get_current_value() - values.get_old_value())) > flt.DeadbandValue):
            return True
        else:
            return False
```

Fixed code:
```
     if (values.get_old_value() is None) or \
                ((abs(values.get_current_value() - values.get_old_value())) > flt.DeadbandValue):
            return True
        else:
            return False
```
And it not only fixes the exception is also makes sure that the Deadband is working again. Because now if an old value is present it will return True and makes the part after the `or` pretty useless.

btw the orginal author of the that part is joaovitor8, it looks like that this in no longer active.

oroulet:
Then this is s proof that blame is not a perfect tool :-)

PR #330, opened 2016-10-14
Move nsrel to xmlimport

bitkeeper:
First attempt to move namespace correction code out of xmlparser to xmlimport.

Part not finished sorting the nodes before import (_sort_nodes_by_parentid). This is a problem if the nodes not already are sorted in the xml file (default behavioral of UaModeler is sorted).

Please review and/or test.


oroulet:
well done. Have a nice vacation!

PR #335, opened 2016-10-18
Added: Support for Double and Float types in XML.

drakes00:
Added support for Double and Float value extensions in XML importer.

### DIFF
--- a/opcua/common/xmlimporter.py
+++ b/opcua/common/xmlimporter.py
@@ -13,6 +13,8 @@ from opcua.common import xmlparser
 def ua_type_to_python(val, uatype):
     if uatype.startswith("Int") or uatype.startswith("UInt"):
         return int(val)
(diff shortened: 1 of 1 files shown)


drakes00:
Hi, I went across an XML file with extensions used to specify a value.
Those were Double and I got error "uatype not handled" from xmlimporter.py.
I juste added a typecast into python float and it seems to have solved my problem.
I tried with different values and they seem to be read OK.

The XML file is "demoserver.xml" from UnifiedAutomations (attachment).
[demoserver.zip](https://github.com/FreeOpcUa/python-opcua/files/535493/demoserver.zip)


Please tell me if I missed anything.

Cheers,
Maxime

oroulet:
looks OK. strange that I forgot those.... the code of xml is messy

oroulet:
thank you

PR #336, opened 2016-10-18
Fix Guid implemenetation

zerox1212:
Improve handling, importing, and exporting of Guids, minor pep fixes

I left the "ua types" style implementation because it was already present. In my opinion treating Guid as a primitive would probably be better, but this will work for now.

One important note that this implementation doesn't protect against doing this:
```myvar = myobj.add_variable(idx, "MyGuid", ua.Variant(uuid.uuid4(), ua.VariantType.Guid))```
This will create a Variant of type Guid which has a Value containing a UUID object instead of a Guid object. This will result in binary errors and I couldn't come up with a nice way of protecting the user from it. The only protection from it is if they do this:
```myvar = myobj.add_variable(idx, "MyGuid", ua.Variant(uuid.uuid4()))```
In this case the `guess_type` code will raise an error and tell the user to supply a ua.Guid().

A second important note is that `node.get_value()` will return a Guid object, not the actual UUID as a string. Perhaps Guid class needs to override the get builtin...

See https://github.com/FreeOpcUa/python-opcua/issues/334 for more discussion.

### DIFF
--- a/opcua/ua/uatypes.py
+++ b/opcua/ua/uatypes.py
@@ -163,12 +163,35 @@ class EventNotifier(_MaskEnum):
 
 class Guid(FrozenClass):
 
-    def __init__(self):
-        self.uuid = uuid.uuid4()
+    def __init__(self, value=None):
+        if value is None or value == "":
+            # if no string is supplied, create a new UUID
+            self.uuid = uuid.uuid4()
+        else:
+            try:
+                # attempt to make a UUID with the supplied value, if hex string has errors use a default string
+                self.uuid = uuid.UUID(value)
+            except ValueError as e:
+                # TODO should notify user of error, but this module has no logger, raise UaError stops execution
+                self.uuid = uuid.UUID("00000000-0000-0000-0000-000000000000")
+
         self._freeze = True
 
     def to_binary(self):
-        return self.uuid.bytes
+        # NOTE THAT self.uuid.bytes WILL NOT WORK FOR UA BINARY PROTOCOL; BYTE ORDER IS NOT CORRECT
+        # convert python UUID 6 field format to OPC UA 4 field format
+        guid = self.uuid
+        f1 = uabin.Primitives.UInt32.pack(guid.fields[0])
+        f2 = uabin.Primitives.UInt16.pack(guid.fields[1])
+        f3 = uabin.Primitives.UInt16.pack(guid.fields[2])
+        f4a = uabin.Primitives.Byte.pack(guid.fields[3])
+        f4b = uabin.Primitives.Byte.pack(guid.fields[4])
+        f4c = bytearray(struct.unpack('8B', struct.pack('>Q', guid.fields[5]))[2:8])  # no primitive .pack available
+        f4 = f4a + f4b + f4c
+
+        # concat byte fields
+        b = f1 + f2 + f3 + f4
+        return b
 
     def __hash__(self):
         return hash(self.uuid.bytes)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by zerox1212** [comment]
As usual, I need some help on how to fix tests...

File: opcua/ua/uatypes.py [hunk 1]
```diff
@@ -163,12 +163,35 @@ class EventNotifier(_MaskEnum):
...
-        return self.uuid.bytes
+        # NOTE THAT self.uuid.bytes WILL NOT WORK FOR UA BINARY PROTOCOL; BYTE ORDER IS NOT CORRECT
+        # convert python UUID 6 field format to OPC UA 4 field format
+        guid = self.uuid
+        f1 = uabin.Primitives.UInt32.pack(guid.fields[0])
+        f2 = uabin.Primitives.UInt16.pack(guid.fields[1])
+        f3 = uabin.Primitives.UInt16.pack(guid.fields[2])
+        f4a = uabin.Primitives.Byte.pack(guid.fields[3])
+        f4b = uabin.Primitives.Byte.pack(guid.fields[4])
+        f4c = bytearray(struct.unpack('8B', struct.pack('>Q', guid.fields[5]))[2:8])  # no primitive .pack available
```
oroulet: from doc "Q 	unsigned long long 	integer 	8". 	so this is probablye a UInt32. Can you try?

opcua/ua/uatypes.py
```diff
@@ -163,12 +163,35 @@ class EventNotifier(_MaskEnum):
...
+        if value is None or value == "":
+            # if no string is supplied, create a new UUID
+            self.uuid = uuid.uuid4()
+        else:
+            try:
+                # attempt to make a UUID with the supplied value, if hex string has errors use a default string
+                self.uuid = uuid.UUID(value)
+            except ValueError as e:
+                # TODO should notify user of error, but this module has no logger, raise UaError stops execution
+                self.uuid = uuid.UUID("00000000-0000-0000-0000-000000000000")
```
oroulet: At lower level it is almost always better to raise exceptions.
you should write something like:
if value is None:
    self.uuid = uuid.uuid4()
elig isinstance(value, uuid): # maybe test size too
    sekf,uuid = val
elif isintance(value, (str, bytes)):
    self.uuid = uuid.UUID(value)
else:
   raise ValueError.....

oroulet:
Is there a difference between an UA guid and a python one? If they are the same it would be great to convert on the fly as we do with datetime.

opcua/ua/uatypes.py
zerox1212: It don't think it will work. I basically cheated here. uuid.fields[5] is actually a 6 byte int. That is why I pack it to bytes and disregard the unused bytes.

opcua/ua/uatypes.py
zerox1212: If uuid module receives a bad value it raises ValueError. I did it this way so that the program doesn't crash if the Guid is invalid. The main reason for this is for apps like modeler. If you make a Guid in modeler and type something wrong you will now just get a default value instead of crashing the entire program.

If it's better to crash the program than set a default then I can change it.

zerox1212:
Python UUID has everything that UA spec Guid has. I have now programmed it in ua_binary as a primitive and your way of using ua_types. Both ways work, I was basically waiting for you to tell me which direction to go.

One thing I did not look into is how Guid NodeIds work. This is the only reason this PR keeps UA Guid class, because I don't know what needs to be done to keep Guid NodeId working.

zerox1212:
See https://github.com/FreeOpcUa/python-opcua/pull/338

zerox1212:
Changed to different implementation, see https://github.com/FreeOpcUa/python-opcua/pull/338

Issue #286, opened 2016-08-16
[XML Import] Bug When Importing Variables/Properties


zerox1212:
When importing an XML to the server there is a bug that occurs if the XML doesn't specify a "Value" element. Note Unified Automation modeler software does not export "Value" by default.

The import parser adds the node with the XML data type, but it doesn't initialize the node with a default value for that data type. As soon as the server starts the node's data type gets set to null. This causes some external clients (such as UA expert) to block writing to the node.

Options to fix:
1. Leave it as is and tell the user that variables/properties with a value of None will become Null data type.
2. If a node is imported with a data type but no value, set the nodes value to an appropriate default to avoid it becoming Null.
3. Always set a value when importing unless XML data type is explicitly set to Null.
4. More of a bug with UA expert, so don't change anything.

Thoughts?

zerox1212:
I might go ahead and make a PR for option 2 if no one has an opinion. It's annoying for testing.

oroulet:
Does the spec days something about this? Using null seems to make sense....

zerox1212:
Honestly I didn't check the spec. Do you know what part to check?

I suppose it does make sense, but it's annoying that you can't write to a node of type null in UA expert.

zerox1212:
Thinking about it more I think it's UA modelers fault. If I create a node of type Double, and set the value to 0 and save it. That should mean that I want to set the value to 0 in the XML, not Null. I'll send an email to Unifed Automation and ask them.

zerox1212:
I'm still struggling with this and Unified Automation didn't help.
I added a "default values" option to importer. Can we add it to python-opcua?


```python
class XMLParser(object):

    def __init__(self, xmlpath, server, enable_default_values=False):
```

Then this in parser:
```python
def _parse_node(self, name, child):
        """
        Parse a XML node and create a NodeData object.
        """
        ..........

        # set default Value based on type; avoids nodes with Value=None being converted to type Null by server
        if self.enable_default_values:
            if obj.datatype is not None:
                if obj.datatype in ("Int8", "UInt8", "Int16", "UInt16", "Int32", "UInt32", "Int64", "UInt64"):
                    obj.valuetype = obj.datatype
                    obj.value = 0
                elif obj.datatype in ("Float", "Double"):
                    obj.valuetype = obj.datatype
                    obj.value = 0.0
                elif obj.datatype in ("Boolean"):
                    obj.valuetype = obj.datatype
                    obj.value = False
[...]

PR #337, opened 2016-10-18
Add default value option to xml importer

zerox1212:
Allows user to decide how to handle xml import
enable_default_values=False means any xml Values that are not set (type None) will be set to data type Null by server
enable_default_values=True means any xml Values that are None will be initialized with a default value to preserve the xml data type

See https://github.com/FreeOpcUa/python-opcua/issues/286


zerox1212:
Is there some concern about merging this?

zerox1212:
My project I'm working with currently uses everything from the recent merges. If you can merge this I can sync my project and test with real software.

zerox1212:
I don't think rebase worked, I have never tried to use that before...

zerox1212:
It wasn't a bug in the other software.
The problem is that if a xml Node does not have a Value tag, importer sets the datatype of the variable, then it sets the value to python type None. Then when the node gets created in the server the server sees a value of None and overwrites the original node data type to Null. After that, the only way undo the node being Null datatype is to write to it from the server. No clients can write to a datatype null.

This change will try to preserve the datatype set in the XML.

zerox1212:
I can't figure out how to correct these conflicts... I'm just going to make a new PR from the latest master.

oroulet:
Try to make sample code to demonstrate issue first, so we can discuss the
best way to resolve it

oroulet:
I am currently very busy at work. I should take a deeper look at this
default value stuff

oroulet:
Can you write sample code to show that issue? Write it so we can use your
code as a test later

PR #338, opened 2016-10-19
Guid ua binary primitive

zerox1212:
For review.

This is UA Guid, but using python UUID. Binary protocol is done on the fly.

Can't merge yet because NodeIds need to be refactored to no longer use ua.Guid() class.

Please comment.


oroulet:
just had a short look at NodeId class code and it seems very easy to add support for guid. support seems to be missing for encoding currently, need to add one more "if" line

zerox1212:
Should I close the other PR and go this route? I will need to move a few other changes from that branch to this one.

Do you think we need backwards support for ua.Guid class? Or can we just delete it?

oroulet:
Delete it

zerox1212:
Added fixes you suggested, moved miscellaneous changes from https://github.com/FreeOpcUa/python-opcua/pull/336/ to this PR. Updated guid node id. Need help with encoding that you mention.

tests/tests_unit.py
```diff
@@ -128,8 +129,8 @@ def test_custom_variant_array(self):
         self.assertEqual(v, v2)
 
     def test_guid(self):
-        g = ua.Guid()
-        sc = ua.StatusCode()
+        # FIXME what tests to add here?
+        pass
```
oroulet: Well, test what you think is necessary. Create a guid object, convert to binary and back and test it is the same value. For example

tests/tests_unit.py
```diff
@@ -150,8 +151,8 @@ def test_nodeid(self):
         s = ua.StringNodeId(53, 0)  # should we raise an exception???
         s1 = ua.StringNodeId("53", 0)
         bs = ua.ByteStringNodeId(b"53", 0)
-        gid = ua.Guid()
-        g = ua.ByteStringNodeId(gid, 0)
+        gid = uuid.uuid4()
+        g = ua.ByteStringNodeId(str(gid), 0)
```
oroulet: What is that? Create a guid nodeid or put all guid tests in method over

zerox1212:
Almost ready, but seems python 2.7 doesn't support the very nice int method `to_bytes()` .

oroulet:
Yes it looks good!

tests/tests_unit.py
zerox1212: I need help here. I do not understand what this test did in the first place...

oroulet:
Looks broken. Write a test you think is good

zerox1212:
I think it is good enough for merge. NodeId tests are the same as before. I'm not sure how to improve them.

zerox1212:
Actually don't merge just yet, I have to test xml exporter.

zerox1212:
Please review, should be ready for merge (I don't know why pypy failed...).

This does bring up a question though:
What to do with instantiate? If an the object that is being instantiated should we automatically create a new UUID? Perhaps we need to add that before merging this.

oroulet:
We can fix instanciate later. Maybe there is a bug in instantiate where we
change type of nodeid. But this is not related

opcua/common/ua_utils.py
```diff
@@ -87,13 +87,21 @@ def string_to_val(string, vtype):
...
         val = ua.LocalizedText(string)
     elif vtype == ua.VariantType.StatusCode:
         val = ua.StatusCode(string)
     elif vtype == ua.VariantType.Guid:
-        val = uuid.UUID(string)
+        # if no guid string is supplied, make a random guid (this is required for modeler)
+        if string is None or string == '':
+            val = uuid.uuid4()
+        else:
+            val = uuid.UUID(string)
```
zerox1212: Just want to point this out and make sure it's acceptable. It might be better to throw an exception, but in modeler it's a pain to type in a valid Guid. Same for datetime. However this function might be used by something else and have unintended affects.

oroulet:
No it is not since null nodeid might be desired. But we could add code in
uamodeller to generate default value

On Thu, Oct 20, 2016, 16:09 Andrew <notifications@github.com> wrote:

*@zerox1212* commented on this pull request.
------------------------------

In opcua/common/ua_utils.py
<https://github.com/FreeOpcUa/python-opcua/pull/338#pullrequestreview-5076612>
:

>      elif vtype == ua.VariantType.LocalizedText:
         val = ua.LocalizedText(string)
     elif vtype == ua.VariantType.StatusCode:
         val = ua.StatusCode(string)
     elif vtype == ua.VariantType.Guid:
-        val = uuid.UUID(string)
+        # if no guid string is supplied, make a random guid (this is
required for modeler)
+        if string is None or string == '':
+            val = uuid.uuid4()
+        else:
+            val = uuid.UUID(string)

Just want to point this out and make sure it's acceptable. It might be
better to throw an exception, but in modeler it's a pain to type in a valid
Guid. Same for datetime. However this function might be used by something
else and have unintended affects.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/FreeOpcUa/python-opcua/pull/338#pullrequestreview-5076612>,
or mute the thread

zerox1212:
I think it's finally ready.

One thing about modeler, I think it would be better to add a new function to python-opcua utils called `get_default` or something. Then modeler can just call `utils.get_default(type)` if the value field is null. This would match the current code organization much better (modeler is a shell, bulk of code in server). What do you think?

PR #339, opened 2016-10-24
Default value on import

zerox1212:
Add option to importer which tries to preserve variable nodes data types
when they don't have a value

This stops variables from being converted to type Null by server (after which no client can write to them until the server changes the type).

Anyone who has an address space generated from Unified Automation Modeler will want this option.

### DIFF
--- a/opcua/common/xmlimporter.py
+++ b/opcua/common/xmlimporter.py
@@ -50,12 +50,12 @@ class XmlImporter(object):
             aliases_mapped[alias] = self._migrate_ns(self.to_nodeid(node_id))
         return aliases_mapped
 
-    def import_xml(self, xmlpath):
+    def import_xml(self, xmlpath, enable_default_values=False):
         """
         import xml and return added nodes
         """
         self.logger.info("Importing XML file %s", xmlpath)
-        self.parser = xmlparser.XMLParser(xmlpath)
+        self.parser = xmlparser.XMLParser(xmlpath, enable_default_values)
 
         dnodes = self.parser.get_node_datas()
         dnodes = self.make_objects(dnodes)
--- a/opcua/common/xmlparser.py
+++ b/opcua/common/xmlparser.py
@@ -90,11 +90,22 @@ class ExtObj(object):
 
 
 class XMLParser(object):
+    """
+    XML Parser class which traverses an XML document to collect all information required for creating an address space
+    """
 
-    def __init__(self, xmlpath):
+    def __init__(self, xmlpath, enable_default_values=False):
+        """
+        Constructor for XML parser
+        Args:
+            xmlpath: path to the xml document; the document must be in OPC UA defined format
+            enable_default_values: false results in xml variable nodes with no Value element being converted to Null
+                                   true results in XML variable nodes to keep defined datatype with a default value
+        """
         self.logger = logging.getLogger(__name__)
         self._retag = re.compile(r"(\{.*\})(.*)")
         self.path = xmlpath
+        self.enable_default_values = enable_default_values
 
         self.tree = ET.parse(xmlpath)
         self.root = self.tree.getroot()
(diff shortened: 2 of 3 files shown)


zerox1212:
I fixed all conflicts. Possible to merge this?

bitkeeper:
@zerox1212 do you also have a test available ?

bitkeeper:
@zerox1212 perhaps a suggestion; 
Instead of directly using obj.datatype, find on which type(nodeid not string)) the obj.datatype is based on ( like Double, Float, Integer, UInteger, Boolean, Enumeration etc). In that way custom datatype are also supported.

Most logic in the parser is movedto the import; should this also be functionality of the importer instead of the parser ?

zerox1212:
I will try to make that correction.

I think this should be in the parser. The problem is when the parser doesn't find an XML `<Value>` tag. Then value is set to python type None. When a node has a Value of None it's data type gets changed to Null.

I will look at how to add a test.

oroulet:
Better spend time analysing the issue, writing options to handle corner
cases is probably a bad idea

bitkeeper:
Can you explain a little bit more, because I'm not sure about your issue ?

Both cases below the `node.get_data_type()` return `NumericNodeId(i=6)`, so the data type is still valid.
If you take a look at [http://opcfoundation.org/UA/2011/03/UANodeSet.xsd](http://opcfoundation.org/UA/2011/03/UANodeSet.xsd) you can see that both are completely valid nodes.  UAExpert show in that situation just a 'null'.

case 1 No Value envelope at all
```xml
    <UAVariable DataType="Int32" ParentNodeId="ns=2;i=5031" NodeId="ns=2;i=6013" BrowseName="1:MyVar">
        <DisplayName>ItemsPerHour</DisplayName>
        <References>
            <Reference ReferenceType="HasTypeDefinition">i=63</Reference>
            <Reference ReferenceType="HasComponent" IsForward="false">ns=1;i=5031</Reference>
        </References>
       
    </UAVariable>
```

case 2 Empty Value envelope 
```xml
    <UAVariable DataType="Int32" ParentNodeId="ns=2;i=5031" NodeId="ns=2;i=6013" BrowseName="1:MyVar">
        <DisplayName>ItemsPerHour</DisplayName>
        <References>
            <Reference ReferenceType="HasTypeDefinition">i=63</Reference>
            <Reference ReferenceType="HasComponent" IsForward="false">ns=1;i=5031</Reference>
        </References>
       
        <Value>         
        </Value>         
    </UAVariable>
```

Maybe you mean `type(node.get_value() )` because that off-course will be `None`  ? 
It that is the case and your application depends on the type of get_value, then personally I find it more a issue of the application then of the python-opcua library.

For example this is used  by the opcua_modeler (btw great initiative !) an import with default values enabled, followed by an export will result in a possible different xml as output).

What about adding the following util function to python-opcua:
`native_type = value_type_to_python_type (nodeid )`

Default value can then easily be generated application side (in case of an `None` value) by calling, without changing the original empty value:
`value = node_type_to_python (nodeid )()`

zerox1212:
Run the server example and import Case 1. Then open UA Expert and connect to your server. Try to write a 31 to that node. It won't let you.

oroulet:
As with earlier, I would like to understand the issue first. If we have a
bug we should fix it, not add options around it.
I did some small tests.
We allow to create variants with value none and type float, for example,
and this seems wrong, but the method to_ binary will fail anyway.
We can create variable with a given data type and a variant null. This is
correct, the prosys client is happy with that, but uaexpert does not allow
writing to such a variant. Looks like a bug in uaexpert.
Anything else?

zerox1212:
I see now that the Variable node data type is indeed preserved. I was mistaken there. I thought the server was changing the nodes data type based on the variant. So it appears that the problem is Ua Expert.

In other words this PR is wrong and my only thoughts on this topic are that Modeler probably needs some work to make sure user has control over his variable values. Currently the text box defaults to 9.99, when it should default to Null and not generate the XML Value tag unless use explicitly enters a value.

Thanks.

PR #340, opened 2016-10-25
default_values

oroulet:
@zerox1212 
This is an attempt to solve your issue. Can you try to describe the issue again?

### DIFF
--- a/opcua/ua/uatypes.py
+++ b/opcua/ua/uatypes.py
@@ -9,8 +9,6 @@ import os
 import uuid
 import re
 import itertools
-if sys.version_info.major > 2:
-    unicode = str
 
 from opcua.ua import ua_binary as uabin
 from opcua.ua import status_codes
@@ -20,6 +18,8 @@ from opcua.ua.uaerrors import UaStatusCodeError
 from opcua.ua.uaerrors import UaStringParsingError
 
 
+if sys.version_info.major > 2:
+    unicode = str
 def get_win_epoch():
     return uabin.win_epoch_to_datetime(0)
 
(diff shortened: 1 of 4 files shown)


zerox1212:
This is probably a good error to have, but it won't solve the issue.

When a node's value gets set, the server will update the data type, correct? (unless it is strong typed) This is where the problem is. Import makes nodes with a variant value of None. The server sees this and updates the nodes type to Null. If the server writes 10.5 to the same node, it's data type gets changed to Double. However my application has a lot of information coming from the clients, if the datatype is Null my clients can't write 10.5 to the node.

oroulet:
Maybe the data type is wrong when importing a node without value?
What happens is you create a nice using
Node.add_varisble(1, 'mynide', None, ua.varianttype.null)
Can this node be written by your client?

oroulet:
This is not correct. A variable node has a data type attribute. Then the
variant has also a data type called variant type in freeopcua. As far as I
know the variant type can be null, a client should not read the variant
type but the data type to write a new value. Looks like uaexpert has a bug
here

opcua/ua/uatypes.py
```diff
@@ -1037,3 +1037,43 @@ def datatype_to_varianttype(int_type):
...
+
+
+def get_default_value(vtype):
+    """
+    Given a variant type return default value for this type
+    """
+    if vtype == VariantType.Null:
+        return None
+    elif vtype == VariantType.Boolean:
+        return True
```
zerox1212: I think False is probably a better default for Boolean.

oroulet:
Added a function to give default value in uatypes. We can use it in modeler
and maybe in create _ nodes.
Looking at your code I see I made a mistake for float
. Will fix it

zerox1212:
I'm working in modeler right now and just realized we have the same problem as UA Expert lol.

```python
def string_to_val(string, vtype):
    """
    Convert back a string to a python or python-opcua object
    Note: no error checking is done here, supplying null strings could raise exceptions (datetime and guid)
    """
    string = string.strip()
    if string.startswith("["):
        string = string[1:-1]
        var = []
        for s in string.split(","):
            s = s.strip()
            val = string_to_val(s, vtype)
            var.append(val)
        return var

    if vtype == ua.VariantType.Null:
        val = None
    elif vtype == ua.VariantType.Boolean:
        val = bool(string)
    elif 4 <= vtype.value < 9:
        val = int(string)
    elif vtype in (ua.VariantType.Float, ua.VariantType.Double):
        val = float(string)
    elif vtype in (ua.VariantType.String, ua.VariantType.XmlElement):
        val = string
    elif vtype in (ua.VariantType.SByte, ua.VariantType.ByteString):
        val = bytes(string)
    elif vtype in (ua.VariantType.NodeId, ua.VariantType.ExpandedNodeId):
        val = ua.NodeId.from_string(string)
    elif vtype == ua.VariantType.QualifiedName:
        val = ua.QualifiedName.from_string(string)
    elif vtype == ua.VariantType.DateTime:
        val = parser.parse(string)
    elif vtype == ua.VariantType.LocalizedText:
        val = ua.LocalizedText(string)
    elif vtype == ua.VariantType.StatusCode:
        val = ua.StatusCode(string)
    elif vtype == ua.VariantType.Guid:
        val = uuid.UUID(string)
    else:
        # FIXME: Some types are probably missing!
        raise NotImplementedError
    return val
```
You can see here that our string to value conversion is checking Variant Type, not the Variable Type. Can you fix that in this PR?

zerox1212:
Actually, I think the problem comes from attrs_widget.py. So ignore what I posted above.

PR #341, opened 2016-10-25
Made subscribe handle most iterables

AlexejStukov:
like dicts, etc. Going for an EAFP-approach here.

### DIFF
--- a/opcua/common/subscription.py
+++ b/opcua/common/subscription.py
@@ -197,9 +197,11 @@ class Subscription(object):
 
     def _subscribe(self, nodes, attr, mfilter=None, queuesize=0):
         is_list = True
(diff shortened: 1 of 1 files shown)


AlexejStukov:
I would store my nodes rather in a dictionary than in a list, so accessing them becomes faster. Using ``[]`` fails on ``dict.values()`` and other such container types/generator expressions (you get ``[dict_values([...])]``). What should i use to check if it's an iterable type? Something like the following?
````python 
import collections
# ...
if isinstance(nodes, collections.Iterable):
````

oroulet:
Yes something like this. Try to google it ...

oroulet:
Dict.values returns an iterator. The correct solution here is probably to pass list(dict.values()) to method.

PR #342, opened 2016-10-25
Improve instantiate when parent has string node id + small fixes

zerox1212:
Is there a way we can do something like this code?

Basically, if you have an ObjectType, and you instantiate from that type, but you want String Node Id (much more readable), the children will also have a string node id.

It looks like this:
image

And when you instantiate the new node with a String Node Id:
image

I feel like this will work fine because the parent node will always be unique, but perhaps if you have some strange structures this would break? Having children with the same browsename would also break this, but I think `instantiate` is already broken in this case because `if not ua_utils.is_child_present(node, c_rdesc.BrowseName)` is already searching via browsename.

(the primary reason I want this is for Modeler)

### DIFF
--- a/tests/tests_common.py
+++ b/tests/tests_common.py
@@ -607,8 +607,29 @@ class CommonTests(object):
         prop1 = mydevicederived.get_child(["0:sensorx_id"])
         var1 = mydevicederived.get_child(["0:childparam"])
         var_parent = mydevicederived.get_child(["0:sensor"])
-        prop_parent = mydevicederived.get_child(["0:sensor_id"])        
+        prop_parent = mydevicederived.get_child(["0:sensor_id"])
 
+    def test_instantiate_string_nodeid(self):
+        # Create device type
+        dev_t = self.opc.nodes.base_object_type.add_object_type(0, "MyDevice2")
+        v_t = dev_t.add_variable(0, "sensor", 1.0)
+        p_t = dev_t.add_property(0, "sensor_id", "0340")
+        ctrl_t = dev_t.add_object(0, "controller")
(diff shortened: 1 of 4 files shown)


oroulet:
This is probably fine. Although I am not really sure we should do it for all nodeid types. If someone create a type with children using GUID node ids, he probably wants the instantiation to also use GUIDs. This is currently not supported, we need some changes in the automatic node creation algo

zerox1212:
Might need to look in the spec. In Unified Automation Modeler all types must be numeric node ids. So in their software you can't create an object type that has a GUID node id. I'm not sure if they did that for simplicity, or what...

In my code I'm making an assumption that if the new node is a String node id, then all the children should do the same.

tests/tests_common.py
```diff
@@ -607,8 +607,29 @@ def test_instantiate_1(self):
...
+        # instanciate device
+        nodes = instantiate(self.opc.nodes.objects, dev_t, nodeid=ua.NodeId("InstDevice", 2, ua.NodeIdType.String), bname="2:InstDevice")
+        mydevice = nodes[0]
+
+        self.assertEqual(mydevice.get_node_class(), ua.NodeClass.Object)
+        self.assertEqual(mydevice.get_type_definition(), dev_t.nodeid)
+        obj = mydevice.get_child(["0:controller"])
+        obj_nodeid_ident = obj.nodeid.Identifier
+        prop = mydevice.get_child(["0:controller", "0:state"])
+        self.assertEqual(obj_nodeid_ident, "InstDevice.controller")
```
zerox1212: This is where we test if the child was created with a "parent.child" style String NodeId

PR #353, opened 2016-11-04
Caching fixes

zerox1212:
Bring back dump and load to do their original function and add methods for caching.

These serve different purposes.
load and dump: just dump the entire address space to binary and then reload it
cache: tries to speed up loading default address space, it doesn't support saving custom nodes and has other drawbacks; for example you can't delete nodes if using the cache

Also added `isession.close()` to the server stop code. This ensures all clean up is done, including the deletion of internal subscriptions (required for dumping address space)

### DIFF
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -491,12 +491,11 @@ class AddressSpace(object):
 
     def dump(self, path):
         """
-        dump address space as binary to file; note that server must be stopped for this method to work
+        Dump address space as binary to file; note that server must be stopped for this method to work
+        DO NOT DUMP AN ADDRESS SPACE WHICH IS USING A SHELF (load_aspace_shelf), ONLY CACHED NODES WILL GET DUMPED!
         """
         # prepare nodes in address space for being serialized
-        for nodeid, v in self._nodes.items():
-            ndata = self._nodes[nodeid]
-
+        for nodeid, ndata in self._nodes.items():
             # if the node has a reference to a method call, remove it so the object can be serialized
             if ndata.call is not None:
                 self._nodes[nodeid].call = None
@@ -506,29 +505,37 @@ class AddressSpace(object):
 
     def load(self, path):
         """
-        load address space from file, overwriting everything in current address space
+        Load address space from a binary file, overwriting everything in the current address space
         """
         with open(path, 'rb') as f:
             self._nodes = pickle.load(f)
 
     def make_aspace_shelf(self, path):
         """
-        make a shelf for storing nodes from the standard address space; nodes are moved to a cache
-        by the LazyLoadingDict class when they are accessed
+        Make a shelf for containing the nodes from the standard address space; this is typically only done on first
+        start of the server. Subsequent server starts will load the shelf, nodes are then moved to a cache
+        by the LazyLoadingDict class when they are accessed. Saving data back to the shelf
+        is currently NOT supported, it is only used for the default OPC UA standard address space
+
+        Note: Intended for slow devices, such as Raspberry Pi, to greatly improve start up time
         """
         s = shelve.open(path, "n", protocol=pickle.HIGHEST_PROTOCOL)
-        for nodeid in self._nodes.keys():
-            s[nodeid.to_string()] = self._nodes[nodeid]
+        for nodeid, ndata in self._nodes.keys():
+            s[nodeid.to_string()] = ndata
         s.close()
 
-    def load_lazy_dict(self, path):
+    def load_aspace_shelf(self, path):
         """
-        load the standard address space nodes from the LazyLoadingDict as needed
+        Load the standard address space nodes from a python shelve via LazyLoadingDict as needed.
+        The dump() method can no longer be used if the address space is being loaded from a shelf
+
+        Note: Intended for slow devices, such as Raspberry Pi, to greatly improve start up time
         """
         class LazyLoadingDict(collections.MutableMapping):
             """
             Special dict that only loads nodes as they are accessed. If a node is accessed it gets copied from the
-            shelve to the cache dict. All user nodes are saved in the cache ONLY
+            shelve to the cache dict. All user nodes are saved in the cache ONLY. Saving data back to the shelf
+            is currently NOT supported
             """
             def __init__(self, source):
                 self.source = source  # python shelf
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -494,11 +494,11 @@ class AddressSpace(object):
         dump address space as binary to file; note that server must be stopped for this method to work
         """
         # prepare nodes in address space for being serialized
-        for nodeid in self._nodes.keys():
-            node = self._nodes[nodeid]
+        for nodeid, v in self._nodes.items():
+            ndata = self._nodes[nodeid]
 
             # if the node has a reference to a method call, remove it so the object can be serialized
-            if hasattr(node, "call"):
+            if ndata.call is not None:
                 self._nodes[nodeid].call = None
 
         with open(path, 'wb') as f:
@@ -511,9 +511,9 @@ class AddressSpace(object):
         with open(path, 'rb') as f:
             self._nodes = pickle.load(f)
 
-    def make_cache(self, path):
+    def make_aspace_shelf(self, path):
         """
-        make a shelve for storing nodes from the standard address space; nodes are moved to a cache
+        make a shelf for storing nodes from the standard address space; nodes are moved to a cache
         by the LazyLoadingDict class when they are accessed
         """
         s = shelve.open(path, "n", protocol=pickle.HIGHEST_PROTOCOL)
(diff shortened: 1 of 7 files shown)


opcua/server/address_space.py
```diff
@@ -491,23 +491,44 @@ def empty(self):
...
+
+            # if the node has a reference to a method call, remove it so object can be serialized
+            if hasattr(node, "call"):
+                node.call = None
+
+            s[nodeid.to_string()] = node
         s.close()
 
-    def load(self, path):
+    def load_cache(self, path):
```
oroulet: Maybe load_lazy() ?

opcua/server/internal_server.py
```diff
@@ -86,20 +86,29 @@ def setup_nodes(self):
...
 
-    def load_standard_address_space(self, cacheFile=None):
-        if cacheFile and path.isfile(cacheFile):
+    def load_standard_address_space(self, cachefile=None):
+        # check for a cache file, in windows the file extension is also needed for the check
+        cachefile_win = cachefile
+        if cachefile_win:
+            cachefile_win += ".dat"
+
+        if cachefile and (path.isfile(cachefile) or path.isfile(cachefile_win)):
```
oroulet: if XXX and (YYY or ZZZ)

with () otherwise you will not get expected result...

zerox1212:
Do you have any ideas how to fix LazyLoadingDict? Specifically how to handle iterations?
I made these edits so far but it doesn't work...
```python
class LazyLoadingDict(collections.MutableMapping):
            def __init__(self, source):
                self.source = source  # python shelve
                self.cache = {}  # internal dict

            def __getitem__(self, key):
                # try to get the item (node) from the cache, if it isn't there get it from the shelve
                try:
                    return self.cache[key]
                except KeyError:
                    node = self.cache[key] = self.source[key.to_string()]
                    return node

            def __setitem__(self, key, value):
                self.cache[key] = value

            def __contains__(self, key):
                return key in self.cache or key.to_string() in self.source

            def __delitem__(self, key):
                # only deleting items from cache is allowed
                del self.cache[key]

            def __iter__(self):
                test = list(self.source.keys())
                s_keys = []
                for key in test:
                    s_keys.append(ua.NodeId.from_string(key))
                all_keys = chain(iter(s_keys), iter(self.cache.keys()))
                return all_keys

            def __len__(self):
                # only returns the length of items in the cache, not unaccessed items in the shelve
                return len(self.cache)
```

zerox1212:
Please comment on https://github.com/FreeOpcUa/python-opcua/pull/353/commits/0fde446e3f2c105aad7aacc34c243414b7accc4e

This is an experiment because I want to use caching (way faster loading) and dump/load to save my server state.
This lets user use caching and back up address space with dump/load. It will update the shelf if any values from standard address space were changed via `dump` and then it will make a binary of all user files.

oroulet:
that caching stuff looks very complicated. What is the original issue?
If a server uses lazy loading and want to dump then there is probably a method in lazy loaderen to load everything in memory.

using the dumping stuff to save a custom tree of node sounds like a very bad idea. this library and OPC UA is not made to store data. It is made to enable interoperability. Data should be saved internally in a database or whatever system designed to store data.

zerox1212:
Fair enough. I will revert the last commit so we can merge my minor fixes.

Just know that if a user uses the caching feature, dump and load probably won't work.

zerox1212:
This should be ready for merge unless you have some comments.

opcua/server/address_space.py
```diff
@@ -491,43 +491,75 @@ def empty(self):
...
     def dump(self, path):
         """
-        dump address space as binary to file
+        dump address space as binary to file; note that server must be stopped for this method to work
         """
-        s = shelve.open(path, "n", protocol = pickle.HIGHEST_PROTOCOL)
+        # prepare nodes in address space for being serialized
+        for nodeid in self._nodes.keys():
+            node = self._nodes[nodeid]
+
```
oroulet: iterating though a dict can/should be done with 

    for k, v in mydict.items():
         whaterver

oroulet: do not call node this structure. nodedata or ndata for example.

opcua/server/address_space.py
```diff
@@ -491,43 +491,75 @@ def empty(self):
...
+            # if the node has a reference to a method call, remove it so the object can be serialized
+            if hasattr(node, "call"):
+                self._nodes[nodeid].call = None
+
```
oroulet: if nodedata.call is not None:

zerox1212:
Sorry, now should be ready.

opcua/server/address_space.py
```diff
@@ -491,43 +491,75 @@ def empty(self):
...
+        make a shelve for storing nodes from the standard address space; nodes are moved to a cache
+        by the LazyLoadingDict class when they are accessed
+        """
+        s = shelve.open(path, "n", protocol=pickle.HIGHEST_PROTOCOL)
         for nodeid in self._nodes.keys():
             s[nodeid.to_string()] = self._nodes[nodeid]
         s.close()
 
-    def load(self, path):
+    def load_cache(self, path):
```
oroulet: this naming are very not very explicit. what does make_cache does? maybe call it lazy_load()? and add a warning in doc that deleting nodes might crash, etc...
but then waht is the difference with load_cache?

opcua/server/address_space.py
zerox1212: Done.

opcua/server/address_space.py
zerox1212: make_cache only creates the python shelve with the standard address space. Load cache instantiates the lazy loader which uses the shelve make_cache created.

zerox1212: If you want I can call it `make_shelve` or something. Other option is to move it to load_cache and just create it there if it doesn't exist...

opcua/server/address_space.py
```diff
@@ -491,43 +491,75 @@ def empty(self):
...
+
+    def make_aspace_shelf(self, path):
+        """
+        make a shelf for storing nodes from the standard address space; nodes are moved to a cache
+        by the LazyLoadingDict class when they are accessed
+        """
+        s = shelve.open(path, "n", protocol=pickle.HIGHEST_PROTOCOL)
         for nodeid in self._nodes.keys():
             s[nodeid.to_string()] = self._nodes[nodeid]
         s.close()
```
oroulet: when should that method be used? What is the point of it?

zerox1212: It creates the python shelf containing the standard address space when the server first starts.

oroulet:
And why would someone need to create a shelve at startup?
I understand we have a dump method to create cache. A load method to load,
and a lazy load to load lazily (but with drawback). But where does that
make cache method fit?

zerox1212:
dump and make_cache create two different things.

dump dumps the entire address space, including the standard address space and user nodes, to a binary file.

make_cache creates a python shelf (dbm) that ONLY has the standard address space. They are not interchangeable without a lot of work.

zerox1212:
Honestly, maybe you would prefer that we just supply a premade shelf. Then `load_standard_addressspace` has 3 options: from code, from xml, from shelf.

opcua/server/address_space.py
zerox1212: Pickle in `dump` and shelve are completely different. A shelve is a primitive dict style database where every item is pickled individually. Those pickled items are accessed by key string. They are only unpickled when they are accessed.
https://docs.python.org/3.4/library/shelve.html

The new dict is the shelve database (shelve is designed to act like a dict, read the docs) stored on disk. They way @stanti made this code it will only ever have the standard address space.

oroulet:
Does it mean we should use load() if we have used dump()? And load_lazy()
if we used make_cache()?

zerox1212:
Yes, they have to be used separately. It also means that if user has his standard address space in shelve (using lazy loader) calling dump will try to pickle the lazy loader object and only items that have actually been accessed by lazy loader will get dumped. That is why I said they are not compatible unless we add code to make them work together (like the commit I reverted).

If you want a SINGLE dump format, we need to add code to so that `dump` loops through the address space and updates the shelf. Then `load` would need to restore the entire shelf to memory. Lastly `lazy_load` would use @stanti class and load from the shelve dynamically.

oroulet:
Ok so we should document which one should be used with another, any give
them symmetrical names. Dump and dump _ shelve then load and load _ shelve
for example.

oroulet:
@zerox1212 does the renaming to Dump() and dump_shelve() then load and load_shelve() make sense? Can you do it then we finally merge that one.

zerox1212:
I can do it later today. I got pulled from the project using this library so it will take me a little bit to make the commit you ask for.

oroulet:
I can do the changes if you think they make sense

zerox1212:
If you like it that way it's fine with me. Personally I liked my original naming (dump for binary, cache for shelve). This made it more clear that they are not compatible and that "cache" is really a special storage for the standard address space ONLY.

oroulet:
Rename to make _ cache and load _ cache then?

zerox1212:
If I have convinced you about how the shelve and dump work I would like that naming. :) I can make the change, give me 5 minutes.

zerox1212:
Actually, look at the current naming...it's very explicit.
`make_aspace_shelf`, I would just change 'load_lazy_dict` to 'load_aspace_shelf'. Is this OK? I will also add comments about dump/shelf usage.

oroulet:
Ok

zerox1212:
I think it is not possible to describe the function any better... I hope someone is gets use of this on Raspberry Pi lol.

zerox1212:
I know you said it should be up to the user to save/restore their nodes, but we could easily extend the shelving into a nice way to store/restore opc ua server state. Perhaps we could see if others have a use case for this. Thanks for reviewing everything, I hope I didn't annoy you with all the comments/commits.

Issue #352, opened 2016-11-04
Export of enumerations fails


bitkeeper:
`XmlExporter._value_to_etree`  treats an enumeration as an extension object:
```python
       if dtype.Identifier > 21:  # this is an extentionObject:
            _extobj_to_etree(el, type_name, dtype, val)
        else:
            val_el = Et.SubElement(el, "uax:" + type_name)
            _val_to_etree(val_el, dtype, val)
```
The compare doesn't work for custom datatypes (like a enum), also a compare for `>` makes no sence incase of if the nodeid is a `StringNodeId`.

Do you have a suggestion/direction howto how to fix this ?
Recursive search for the datatype until a Identifier in ns=0 is found and decide on that basis ?

Example output for an enummeration:
```xml
<UAVariable DataType="MyEnum" ParentNodeId="ns=1;i=5031" NodeId="ns=1;i=6012" BrowseName="2:State" UserAccessLevel="3" AccessLevel="3">
        <DisplayName>MyEnum</DisplayName>
        <References>
            <Reference ReferenceType="HasTypeDefinition">i=63</Reference>
            <Reference ReferenceType="HasComponent" IsForward="false">ns=1;s=5031</Reference>
        </References>
        <Value>
            <uax:Int32>0</uax:Int32>
        </Value>
</UAVariable>
```

oroulet:
Correct. I saw that code yesterday and wondered about it...

bitkeeper:
@oroulet 
Below is pseudo code that shows the current flow for exporting a simple variable:

As part of a fix I want to determine the basetype for the variable with a new function `get_variable_basetype `in `ua_utils`. It returns a nodeid with ns=0 and and i<=30.
Then a save compare can be made for i<=21.

For this fix I need to have either a node or server object in several of the helper functions like `_value_to_etree`, `member_to_etree` and `_val_to_etree`.

What is your prefered direction for this fix:

1. use as datatype a Node instead of a NodeId 
2. pass an additional server argument
3. make the helper methods part of XmlExporter
4. ...

Pseudo code of the variable export flow:
```python
def XmlExporter.add_variable_common       
  value_to_etree      # simple value flow 

def value_to_etree
  _value_to_etree     # simple value flow
    
def _value_to_etree
  if                   # problem is here
    _extobj_to_etree
  else
    _val_to_etree  # simple value flow
  
def  _val_to_etree
  if
                  # simple value flow ends here
  else
    for
      member_to_etree

[...]

PR #354, opened 2016-11-07
fix export Enum and custom datatype variables

bitkeeper:
fix for #352

### DIFF
--- a/opcua/common/xmlexporter.py
+++ b/opcua/common/xmlexporter.py
@@ -9,6 +9,7 @@ from copy import copy
 
 from opcua import ua
 from opcua.ua import object_ids as o_ids
+from opcua.common.ua_utils import get_node_supertypes
 
 
 class XmlExporter(object):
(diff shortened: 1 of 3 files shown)


opcua/common/xmlexporter.py
```diff
@@ -228,7 +229,26 @@ def add_variable_common(self, node, el):
...
+        Returns:
+            NodeId of datatype base or None in case base datype can not be determined
+        """
+        if datatype.NamespaceIndex == 0 and type(datatype) == ua.NumericNodeId and datatype.Identifier <= 30:
+            return datatype
+        dtype_supers_nodes = get_node_supertypes(self.server.get_node(datatype), includeitself=True, skipbase=True)
+        dtype_supers_nodeids = [node.nodeid for node in dtype_supers_nodes if node.nodeid.NamespaceIndex == 0 and  node.nodeid.Identifier <= 30]
+        if len(dtype_supers_nodeids) >= 1:
+            return dtype_supers_nodeids[0]
+        return None
```
oroulet: Did you copy the method from ua_utils.py? Can't you call it from ua_utils.py? you can modify/split the one in ua_utils.py if you need so

I think we should raise an exception if no data type is found, this is not normal. so raising an exception is correct and clearly shows there is an issue

tests/tests_xml.py
```diff
@@ -246,15 +246,24 @@ def test_xml_ext_obj(self):
         self.assertEqual(arg.Description, arg2.Description)
         self.assertEqual(arg.DataType, arg2.DataType)
 
+    def test_xml_enum(self):
+        o = self.opc.nodes.objects.add_variable(2, "xmlenum", 0, varianttype=ua.VariantType.Int32, datatype=ua.ObjectIds.ApplicationType)
+        self._test_xml_var_type(o, "enum")
```
oroulet: Can you add another test for an enum based on EnumValues and not string? us ua.AttributeWriteMask (i=347) for example.

bitkeeper: test is added

oroulet:
well done!

PR #355, opened 2016-11-08
export to xml based on namespaces

bitkeeper:
Default the `server.export_xml` requires a node list for export.  An alternative `server. export_xml_by_ns` export default all nodes (except from namespace 0). Also you can be more selective by provide a list of ns indexes or uris of the nodes that should be exported.

Examples:
```python
server. export_xml_by_ns export('all.xml') # export all nodes (excepts ns=0)
server. export_xml_by_ns export('all.xml', 4) # export only nodes in ns=4
server. export_xml_by_ns export('all.xml', [3,4]) # export only nodes in ns=3 and 4
server. export_xml_by_ns export('all.xml', ['myspace.org','myspaceto.org']) 
```
Helper method `get_nodes_of_namespace` for getting the nodes of a certain namespace is added to `ua_utils`

### DIFF
--- a/opcua/common/ua_utils.py
+++ b/opcua/common/ua_utils.py
@@ -238,3 +238,36 @@ def get_variable_basetype(server, datatype_id):
     if not dtype_supers_nodeids:
         raise ua.UaError("Datatype must be a subtype of builtin types %s" % datatype_id.nodeid)
     return dtype_supers_nodeids[0]
+
+def get_nodes_of_namespace(server, namespaces=[]):
+    """
+    Get the nodes of one or more namespaces .      
+    Args:
+        server: opc ua server to use
+        namespaces: list of string uri or int indexes of the namespace to export
+    Returns:
+        List of nodes that are part of the provided namespaces
+    """
+    ns_available = server.get_namespace_array()
+    ns_count = len(ns_available)
+
+    if not namespaces:
+        namespaces = ns_available[1:]
+
+    if type(namespaces) is not list and type(namespaces) in [str, int]:
+        namespaces = [namespaces]
+
+    namespace_indexes = []
+    for namespace in namespaces:
+        if type(namespace) is str:
+            namespace = server.get_namespace_index(namespace)
+        if namespace >= ns_count:
+            raise ua.UaError("Namespace index '%d' is invalid" % namespace)
+        namespace_indexes.append(namespace)
+
+    nodes = []
+    for nodeid in server.iserver.aspace.keys():
+        if nodeid.NamespaceIndex != 0 and  nodeid.NamespaceIndex in namespace_indexes:
+            node = server.get_node(nodeid)
+            nodes.append(node)
+    return nodes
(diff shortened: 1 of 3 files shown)


zerox1212:
Good idea. We should also integrate this option into modeler.

opcua/common/ua_utils.py
```diff
@@ -238,3 +238,36 @@ def get_variable_basetype(server, datatype_id):
...
+    Returns:
+        List of nodes that are part of the provided namespaces
+    """
+    ns_available = server.get_namespace_array()
+    ns_count = len(ns_available)
+
+    if not namespaces:
+        namespaces = ns_available[1:]
+
+    if type(namespaces) is not list and type(namespaces) in [str, int]:
```
oroulet: for some reasons it is advices to use isintance(a, (t1, t2)) instead of type ==. Not very important here. but good habit

opcua/common/ua_utils.py
```diff
@@ -238,3 +238,36 @@ def get_variable_basetype(server, datatype_id):
...
+        namespaces = [namespaces]
+
+    namespace_indexes = []
+    for namespace in namespaces:
+        if type(namespace) is str:
+            namespace = server.get_namespace_index(namespace)
+        if namespace >= ns_count:
+            raise ua.UaError("Namespace index '%d' is invalid" % namespace)
+        namespace_indexes.append(namespace)
```
oroulet: shorter and without new query:

    namespace_indexes = [ n for n in namespaces if isinstance(n, int) else ns_available.index(n)]

opcua/common/ua_utils.py
```diff
@@ -238,3 +238,36 @@ def get_variable_basetype(server, datatype_id):
...
+
+    nodes = []
+    for nodeid in server.iserver.aspace.keys():
+        if nodeid.NamespaceIndex != 0 and  nodeid.NamespaceIndex in namespace_indexes:
+            node = server.get_node(nodeid)
+            nodes.append(node)
+    return nodes
```
oroulet: Could also use list comprehension here... but your code is probably more understandable...

opcua/common/ua_utils.py
bitkeeper: This would lead to:
```python
 namespace_indexes = [ n if isinstance(n, int) else ns_available.index(n) for n in namespaces ]
 if max(namespace_indexes) >= len(ns_available):
    raise ua.UaError("Namespace index '%d' is invalid" % max(namespace_indexes))
```
Not sure if this makes the code more readable and adding comment to explain it feels unnatural...
Also the export is not really time critical.
But in the end it is your party, so I will make the change ;-)

bitkeeper: Are you talking about the `if isinstance(n, int) else ns_available.index(n)` before the `for` instead of after ?
A `if` in a _list comprehension_ filters the list. That is not what we want, we want to change the value which requires a  _conditional expression_ before the `for`.

The namespace index >= array size check is because I only want to support namespace that are registered. I have also created a test for it.

PR #356, opened 2016-11-10
If we receive an unknown NodeId, return an empty list instead of cras…

levkevi:
…hing


oroulet:
when is this an issue?
If we get a request for an unknown nodeid we should return an ua error. so I am afraid of the consequences of this change

levkevi:
I had a problem using Softing OPC UA Client where the server didn't seem to be to return a response. It basically looped forever while trying to browse the nodes. There may be better ways to fix it thought.

oroulet:
Keep it open. I haven't had time to look at it. But there is probably
something to fix here...

PR #357, opened 2016-11-10
Xml server example

levkevi:


zerox1212:
Can you give some more description about the purpose of this example?

levkevi:
There was no simple examples that mapped a method that was imported from a Xml file to a function. There is also no function that returns a list of value. The example show how to do a server that does that and what the server gets back when using the methods. 

I also handle the connection using a context manager so it always closes if there is an exception.

oroulet:
nice code

zerox1212:
FYI you can also put your method into the XML, then you can link the python function with `server.link_method`.

oroulet:
@zerox1212 good point. we should update the example

PR #363, opened 2016-11-15
Ignore the "Models" xml tags as they don't contain nodes

levkevi:


bitkeeper:
@oroulet to be future safe for other new extensions; 
Should we not change the test for if it is a node, by testing to be in a list instead of out the list:
From:
`if tag not in ["Aliases", "NamespaceUris", "Extensions", "Models"]:`
To:
`if tag in ["UAMethod", "UAVariable", "UAObject", ...]:`

oroulet:
yes it sounds better.

levkevi:
It crashes when the tag does not have the nodeid attribute. We should see if it is not empty instead, that would be even safer.

PR #364, opened 2016-11-16
Solution for hanging UASocketClient (#316)

AlexejStukov:
Maybe this works better than #316 ...


AlexejStukov:
This little fix lets you end the listener thread by calling ``Client.disconnect_socket``.

oroulet:
ok we try

PR #366, opened 2016-11-17
Changed values for RequestedLifetimeCount and RequestedMaxKeepAliveCount

AlexejStukov:
I have switched the values for `RequestedLifetimeCount `and `RequestedMaxKeepAliveCount`, because on an idle connection (no changes of the monitored items) the server will wait for `KeepAliveCount*PublishingInterval` and then send an empty message to the client. This empty message will trigger a PublishRequests. Buf if `LifeTimeCount` is smaller than `KeepAliveCount`, the server will close the subscription before sending the empty message.

Also added the values as `kwargs` to create_subscription.

> KeepAliveCount
If there is no data to send after the next PublishingInterval, the server will skip it. But KeepAlive defines how many intervals may be skipped, before an empty notification is sent anyway: to give the client a hint that the subscription is still alive in the server and that there just has not been any data arriving to the client.
LifeTimeCount
The client’s responsibility is to send PublishRequests to the server, in order to enable the server to send PublishResponses back. The PublishResponses are used to deliver the notifications: but if there are no PublishRequests, the server cannot send a notification to the client.
The server will also verify that the client is alive by checking that new PublishRequests are received – LifeTimeCount defines the number of PublishingIntervals to wait for a new PublishRequest, before realizing that the client is no longer active. The Subscription is then removed from the server.
(see [source](https://www.prosysopc.com/blog/opc-ua-sessions-subscriptions-and-timeouts/))

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -465,7 +465,7 @@ class Client(object):
         """
         return Node(self.uaclient, nodeid)
 
-    def create_subscription(self, period, handler):
+    def create_subscription(self, period, handler, lifetime_count=10000, max_keep_alive_count=3000):
         """
         Create a subscription.
         returns a Subscription object which allow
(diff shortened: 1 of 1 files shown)


oroulet:
I think a better ways to fix that kind of corner cases is to accept a CreateSubscriptionParameters() arg

    def create_subscrition(period, handler):
        if not isinstance(period, CreateSubscriptionParameters):
             make a new one
        return xxxx

and of course document
We should do not the same for the server

oroulet:
Looks good. Could you also update doc string?

AlexejStukov:
I've already added 
````
period argument is either a publishing interval in seconds or a 
CreateSubscriptionParameters instance. The second option should be used
if the opcua-server has problems with the default options.
````
to the docstring of the method. Is it enough?

AlexejStukov:
Thanks!

Issue #365, opened 2016-11-16
Client write callback in server


joaopmrod:
How do I trigger a handle function in the server when a client writes to a variable?

oroulet:
You subscribe to the variable on server side?

joaopmrod:
But then I will get a callback even when I write to it...

oroulet:
Yes. Do you have a good use case for only tracking writes from clients?

joaopmrod:
I'm writting a OPC server for a specific instrument. When a value changes in the instrument I want to update it in the OPC server and do nothing else. 

When a client changes is in the OPC server I want to trigger a callback function that will change that varible in the instrument.

joaopmrod:
Can I have a callback in the server when a client try to read one node?

oroulet:
If you want to monitor everything handing in server you can override the
InternalServer class and the session class then you can know everything
happening. But I am suspecting a bad design in your application if you need
such things....

zerox1212:
You could just keep track of values being written by the instrument. Then the subscription call back from a server write will always have the same value. If a client writes the subscription call back will have a different value and you will know it wasn't from the server it self.

oroulet:
But I am interested to know in what context, we have a value from an
instrument, which should be written by clients. That does not make sense in
my mind.

zerox1212:
My guess is that he is reading a parameter cyclically from the instrument and publishing it on the server. Then he wants to catch if an external client writes to the same address so he can change the instrument parameter.

joaopmrod:
Exactly

oroulet:
That does not explain anything! But thinking about it there might be cases
where one would like not to generate events. For example if the underlying
system has another interface that modify its variables, one probably want
to sync to changes from that interface without generating events to avoid
update loops.
To support that we need to add one more argument to Node.set_ value() and
server. write(), called for example no_event=False.

oroulet:
Maybe it is easier to add a set _ silent() method to the session in
internal server. Then we do not need to modify api

zerox1212:
I don't understand what you mean.

He basically wants to subscribe to a node for data change notification, but ignore all data changes from the server itself (internal server/session).

bitkeeper:
Isn't this the same request as #310 ?

The internal session already knows if the set value comes from external or internal:
```python
 def InternalSession.write(self, params):
        if not self.external:
            # If session is internal we need to store a copy og object, not a reference,
[...]

Issue #310, opened 2016-10-04
Datachange Source (Client vs. Server)


kevincolyar:
I'm looking for a way to differentiate between data changes that were made by a client connection and those which were made by the server itself.  

My application runs calculations on a 4s interval that updates the server's nodes with its latest data.  Those nodes can also be written to by the client and applied to the next calculations.  I'm looking to have my server update its nodes without triggering a bunch of datachange events, but at the same time know that a client is writting new data.

zerox1212:
I think you asked this before?

Your server needs to write directly to the address space dict or you need to change your concept.

kevincolyar:
I did but I must have deleted it.  I couldn't find the old issue.

Would you guys frown on submitting a pull request to add a `with_callbacks` kwarg flag to the server's `write` method and pass it all the way down to the `AddressSpace#set_attribute_value`?  That way I don't have to duplicate everything in `set_attribute_value`.

Basically:

In InternalServer:

```python
    def write(self, params, with_callbacks=True):
```

In AddressSpace#set_attribute_value:

```python
    if with_callbacks and old.Value != value.Value:  # only send call callback when a value change has happend
```

kevincolyar:
I've been able to directly update the address space nodes and can see the changes in a simple python-opcua client.  However, my TOP Server (https://www.softwaretoolbox.com/topserver/) does not see updates.  The TOP Server does see the changes if I update node normally.  

Here is how I'm updating the node values:

```python
        val = ua.DataValue(ua.Variant(float(v), ua.VariantType.Float))
        val.SourceTimestamp = datetime.datetime.utcnow()
        val.ServerTimestamp = datetime.datetime.utcnow()
        val.StatusCode = ua.StatusCode(ua.StatusCodes.Good)
        address_space._nodes[nodeid].attributes[ua.AttributeIds.Value].value = val
```

Do I need to call something to trigger update events?

oroulet:
In python-opcua there is no difference between an external and internal
client. The subscription code is already quite complex so you would need a
very good case to add even more complexity by differentiating client
origins.
From your description of you application it sounds like you should have two
nodes: one read only updated internally and one for external clients that
is subscribed to internally and copy data to the read only mode.
But as I said if you have a very good use case we may think about it....

zerox1212:
I gave him the same advice when he asked this before.
https://github.com/FreeOpcUa/python-opcua/issues/145

bitkeeper:
[...]

Issue #145, opened 2016-03-15
Updating Values Without Triggering Server Subscriptions


kevincolyar:
I currently have a handler that is called via a subscription when my values are changes by a connected client.  I'm looking for a way to update the same values from the server **without** triggering the subscription.  The subscriptions are what tell me a client action has modified a value, I don't want those to trigger when they are modified server side.

Ideally I'd like to only subscribe to data changes coming from clients.  Ideas?

oroulet:
You need to modify the db dict directly if you want such a thing. But why? Do you have a use case?

kevincolyar:
Is it possible to subscribe to client write events?  That would probably be a better solution if it's possible.

zerox1212:
I don't really understand why you would want to do that.
Wouldn't it be better to have two separate nodes?

Node 1 is read only that your server software writes to.
Node 2 is a normal client subscribed node.

Then your application can  update the server side node whenever a client writes to node 2.

oroulet:
No. Not possible but as Andrew asked: why would you want such a thing?

kevincolyar:
Closing this. I found a better way of going about what I was trying to solve.

PR #374, opened 2016-11-22
WIP: Suppress datachange events of internal source

bitkeeper:
Make it possible to suppress datachange events of internal source by set set_silent_events(True).

Is related to issue #310, #365 and #145.

### DIFF
--- a/tests/tests_server.py
+++ b/tests/tests_server.py
@@ -5,9 +5,10 @@ import time
 
 from tests_common import CommonTests, add_server_methods
 from tests_xml import XmlTests
-from tests_subscriptions import SubscriptionTests
+from tests_subscriptions import SubscriptionTests, MySubHandler
 from datetime import timedelta, datetime
 from tempfile import NamedTemporaryFile
+from concurrent.futures import TimeoutError
 
 import opcua
 from opcua import Server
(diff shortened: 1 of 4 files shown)


tests/tests_server.py
```diff
@@ -435,6 +436,31 @@ def get_ns_of_nodes(nodes):
...
+
+        myhandler.reset()
+        v1.set_value(5)
+        node, val, data = myhandler.future.result(timeout=10)
+        self.assertEqual(val, 5)
+
+        myhandler.reset()
+
+        # TODO: this feels not good. Maybe added set_silent_events to a higher class ?
+        self.opc.iserver.isession.set_silent_events(True)
```
bitkeeper: @oroulet 
Suggestion about this location ? Seems a little bit deep and very internal to call the set_silent_events.

oroulet: Add a helper method to server. But I am wondering about the name. Maybe disable_data_change_notification(true) is more explicit

bitkeeper:
@oroulet 
This implementation was a little bit too short-sighted, because this will disable all events also to the external subscriptions.

oroulet:
a good solution is not much more complicated, you need to add another argument to attribute_service.write to tell not to generate events. Call it no_events=False for example.

bitkeeper:
To be sure; the goal is that if in code an node.set_value is done:
* For external subscriptions stills receives updates.
* The internal subscriptions should be ignored. 

I have already added an additional argument to attribute_service.write and 
attribute_service.set_attribute_value. 
Can you give me a little bit more direction on how to approach this issue ?

oroulet:
I am not sure, I really would like to avoid modifying the subscription stuff, it is already far too complex..

bitkeeper:
I close this pr for now ... until I have more time and inspiration

PR #376, opened 2016-11-28
Removed duplicate code from XML parser

drakes00:
Allows to parse ListOfDateTime values and seems to correct a typo for Bytes.

--

More details in issue #377 

By the way, XML with "Byte" and "SByte" gives me warning of non implemented types, however they are mentioned in documentation from OPC Fondation while "Bytes" is not. 
Is that a macro-type for both?

### DIFF
--- a/opcua/common/xmlparser.py
+++ b/opcua/common/xmlparser.py
@@ -8,38 +8,21 @@ import sys
 
 import xml.etree.ElementTree as ET
 
+from opcua.common import ua_utils
+from opcua import ua
+
+
+def ua_type_to_python(val, uatype_as_str):
+    """
+    Converts a string value to a python value according to ua_utils.
+    """
+    return ua_utils.string_to_val(val, getattr(ua.VariantType, uatype_as_str))
 
 def _to_bool(val):
-    return val in ("True", "true", "on", "On", "1")
-
-
-def ua_type_to_python(val, uatype):
-    if uatype.startswith("Int") or uatype.startswith("UInt"):
-        return int(val)
-    elif uatype.lower().startswith("bool"):
-        return _to_bool(val)
-    elif uatype in ("Double", "Float"):
-        return float(val)
-    elif uatype == "String":
-        return val
-    elif uatype in ("SByte", "Byte", "Bytes", "ByteString", "ByteArray"):
-        if sys.version_info.major > 2:
-            return bytes(val, 'utf8')
-        else:
-            return val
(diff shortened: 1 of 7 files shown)


oroulet:
Bytes is probably a typo

drakes00:
Seems to have introduce a failing test on python3.4 (I'll test both versions next time).
I'm looking at the bug.

oroulet:
There is no issue with python3, we have some time dependant tests, if the tests run too slowly, sometimes one fail..... should fix it one day...
I think everything looks correct in your PR

oroulet:
would be great if you could add a few tests in tests_xml for th bugs you wanted to fix, should be easy

drakes00:
Ok, I wasn't able to reproduce the timing fail.
Sure, I'll add few more tests.

opcua/common/xmlparser.py
```diff
@@ -206,65 +201,76 @@ def _parse_attr(self, el, obj):
...
-            obj.value = _to_bool(child_el.text)
+        obj.valuetype = ntag
+        if ntag == "Null":
+            obj.value = None
+        elif ntag in (
+                "SByte", "Byte",
+                "Int16", "UInt16", "Int32", "UInt32", "Int64", "UInt64",
+                "Float", "Double",
+                "Boolean",
+                "DateTime" #FIXME Check format in a specific elif?
```
drakes00: Not sure since ByteString and String are part of ua.ua_binary.Primitives and they are in another elif.
However Primitives1 might do the trick if we treat DateTime in its own elif (might also be prettier for future checks if any).
I'll give it a try

oroulet:
maybe ByteString and String do not need to be part of another elif? and
this can always be solved by testing before..

drakes00:
Hum, I forgot to add pytz to requirements.

drakes00:
Trying with pytz with travis requierement but I din't though it would introduce a new dependency.
Maybe that should be removed, not sure.
What do you think?

drakes00:
Forgot to say, tests cases for QuantifiedNames are commented since their parsing seems to not be implemented and I'm not sure how to do it.

oroulet:
merging this one, it is good. Although I am wondering about the depenency on pytz...

drakes00:
I just checked without and it seems to pass ok so I guess it could be removed.
The xmlparser will just accept DateTime unaware which is technicaly refused in specs but I have no idea of the consequences thereafter.

It's up to you on this one,
Cheers

PR #378, opened 2016-12-01
Improve deadband handling, skip if deadbandtype is none.

bitkeeper:
Client can set a deadband filter to several types (Part 4 7.16.2). Currently this mode isn't checked.
This pr makes the check of those types explicit.

Also some errors can be prevent by this, for example when string type provides a filter with type None. The current situation results in some math checks with strings, which results in an exception.


oroulet:
I do not use that so I trust you

PR #379, opened 2016-12-01
xmlimport set default acceslevel to read

bitkeeper:
If in the information model UAVariable has no explicit set accesslevel, the level is read.

### DIFF
--- a/opcua/common/xmlimporter.py
+++ b/opcua/common/xmlimporter.py
@@ -77,7 +77,7 @@ class XmlImporter(object):
                 continue
             nodes.append(node)
         return nodes
-    
+
     def make_objects(self, node_datas):
         new_nodes = []
         for ndata in node_datas:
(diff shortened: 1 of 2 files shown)


opcua/common/xmlimporter.py
```diff
@@ -173,8 +173,12 @@ def add_variable(self, obj):
             attrs.ValueRank = obj.rank
         if obj.accesslevel:
             attrs.AccessLevel = obj.accesslevel
+        else:
+            attrs.AccessLevel = ua.AccessLevel.CurrentRead.mask
```
bitkeeper: For the XML schema it is part of the spec.
Are you sure about make it also default for VariableAttributes() ? It people already used bit masks, it can set an unwanted read access ?

oroulet:
I am not sure, but I think that default is currently undefined, which might
be surprising in some cases...or not...

bitkeeper:
I created pr #381 with only the changes to the constructor VariableAttributes. 
You probally want merge that one instead of this pr.
If it is fine with you, you can forget this pr.

Issue #373, opened 2016-11-21
ExtensionClasses do not support namespaces


levkevi:
It looks like the ExtensionClasses do not support namespaces.

To add support for it, we'd need to either use a tuple as a key for ObjectIdNames and ObjectIds or add a new system for the ExtensionClasses that are not in the default namespace. What should we do? It looks like object_ids.py is autogenerated.

zerox1212:
Did you find anything about it in the UA spec?

bitkeeper:
ua.ObjectIdNames and ua.ObjectIds are only shorthands for ids in the default namespace 0. That is why they don't contain namespaces.  The class uatypes.ExtensionObject contains a TypeId as NodeId.

So your are probably talking about something different ?

levkevi:
Ok so I should create another data structure to support the namespaces in custom objects. I used those to register my classes but it seems that it was not the proper place to do so.

levkevi:
Erm, I think that I should actually use a structure to return my data. I'm not exactly trying to send an object, I want to encode/decode a structure into a python object.

zerox1212:
Are you saying you want a custom UA object to be treated as a native python object? This functionality isn't part of the server. However, if that is what you want to do I have example code you can look at which creates python objects from UA objects, then keeps them synchronized via subscription. That way in my application I only need to work with the python class.

oroulet:
Custom structure should always be constructed as node under 0:data_types/0.Structures and to decode custom structure, we should look in the addressspace and from the node tree build automatically decode/encode. My issue is that I have not understood yet how to get the order of nodes. nodes are not ordered in a tree..... but maybe we could use the NodeId order....

so I am expected that read_attribute returned a variant containing an Extension object that we feed to a method called decode_extension object that uses the address space tree to decode and return a python object.

same for encoding

zerox1212:
Do you know what part of the spec this might be in @oroulet ? I can take a look.

oroulet:
No idea. It should be in mapping part but I could not find anything

zerox1212:
Which "part" is mapping in?

zerox1212:
Docs say it's based on the definition order. The example looks clear to me.
![image](https://cloud.githubusercontent.com/assets/13090472/20533383/f4cb0040-b0a2-11e6-92be-5a11c4b8e637.png)

oroulet:
[...]

PR #380, opened 2016-12-01
xmlexport improve compat with xml schema

bitkeeper:
Improve compatibility with:
* https://opcfoundation.org/UA/2011/03/UANodeSet.xsd
* https://opcfoundation.org/UA/2008/02/Types.xsd
Impove this compatibility make it possible to read the information models into other tools like UaModeler.

Mostly fixes to garantee the order of the elements according the xsd:
* NamespaceUris: Make sure that the namespacuris are on top (above the aliases)
* UAVariable: Place the values below the references
* uax:Boolean:  values should be 'true' and 'false' (lower case)
* Argument: place the members in the correct order

Btw correct format of the xml file can be checked with XML validator against the XSD.

### DIFF
--- a/opcua/common/xmlexporter.py
+++ b/opcua/common/xmlexporter.py
@@ -236,6 +236,7 @@ class XmlExporter(object):
         Add a UA variable element to the XML etree
         """
         var_el = self._add_node_common("UAVariable", node)
+        self._add_ref_els(var_el, node)
         self.add_variable_common(node, var_el)
 
         accesslevel = node.get_attribute(ua.AttributeIds.AccessLevel).Value.Value
@@ -254,7 +255,6 @@ class XmlExporter(object):
         var = node.get_attribute(ua.AttributeIds.Historizing)
         if var.Value.Value:
             var_el.attrib["Historizing"] = 'true'
-        self._add_ref_els(var_el, node)
 
     def add_etree_variable_type(self, node):
         """
@@ -313,7 +313,8 @@ class XmlExporter(object):
             ref_el = Et.SubElement(aliases_el, 'Alias', Alias=name)
             ref_el.text = nodeid.to_string()
 
-        self.etree.getroot().insert(0, aliases_el)
+        # insert behind the namespace element
+        self.etree.getroot().insert(1, aliases_el)
 
     def _add_ref_els(self, parent_el, obj):
         refs = obj.get_references()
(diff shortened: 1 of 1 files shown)


oroulet:
Thanks! I am really not an xml export, do feel free to explain what is
wrong. Maybe you can add automatic test to check generated files with xsd
in tests?

opcua/common/xmlexporter.py
```diff
@@ -404,8 +407,20 @@ def _extobj_to_etree(self, val_el, name, dtype, val):
...
-            self.member_to_etree(struct_el, name, ua.NodeId(getattr(ua.ObjectIds, vtype)), getattr(val, name))
+        items_keys = val.ua_types.keys()
+        skip_empty = False
+        if dtype == ua.NodeId(ua.ObjectIds.Argument):
+            skip_empty = True
+            items_keys = [name for name in ['Name',
+                                            'DataType',
+                                            'ValueRank',
+                                            'ArrayDimensions',
+                                            'Description'] if name in items_keys ]
```
oroulet: Why do we need to hard code the order? That seems strange....

bitkeeper: That isn't so strange: Just like `Argument.to_binary` also requires an fixed order of the members, the xmls schema (https://opcfoundation.org/UA/2008/02/Types.xsd) also requires an fixed order.

In a lot of cases the order doesn't matter, but if a tool checks the import against the scheme the xml parse will fail.

oroulet:
But it is a lot of code for one of 1000 cases, isn't it any logic in that
ordering? Maybe we can use the order of nodeids?

oroulet:
Ok we do not have nodeids here...

bitkeeper:
There are indeed a lot more cases, but using arguments (for methods) is that common. Now we have at least a working version for the most basic stuff. 

If more cases are required to be fixed (read really used by others) an order lookup will be required to prevent a code bloat. But lets only invest that time when it is required, I sure that we both have better things to at the moment ... 

The current matching FrozenClass currently contains a dictionary, so it will also not provide the order  information.

Maybe if the matching FrozenClass would contain the order of the member it would be possible to automate:
* `to_binary`
* `_binary_init`
* `__str__`
* `__init__`
* xml serialize order

zerox1212:
I have thought about this a few times, and have been wondering if it would be smart to refactor the entire library to use OrderedDict. It isn't very pythonic, but the UA spec seems to assume things like classes will have fixed orders.

We have the same issue with custom structures mentioned here after much research. https://github.com/FreeOpcUa/python-opcua/issues/373

oroulet:
Ordereddict is supposed to be a drop in replacement to dict, so it should
not be hard to use, but it may impact performance and I do really see where
it would help...

bitkeeper:
@oroulet To get this pr accepted: 
Should I revert the changes from this pr concerning the order of Argument members ?
The order fixes in this pr are indepent from it.

oroulet:
If you can split it, it would be great. I am unsure about the Argument ordering. It is a lot of code for something which is quite suspect..

bitkeeper:
Ok reverted the part about the member order of the Argument. Nevertheless at least the implementation of the member order of the `Argument `is wrong. Still an other implementation is required for that.

oroulet:
Make another pr for the order stuff
 Maybe we'll find a generic solution...

PR #381, opened 2016-12-02
set default access level of nodes to read

bitkeeper:
Currently the default AccessLevel and UserAccessLevel of node is undefined (=0 ).
This pr sets default to CurrentRead.


oroulet:
you cannot do that, you modified a file which is autogenerated. modify the struct at the end of uaprotocol_hand.py

bitkeeper:
ok oeps reverted change and applied it to uaprotocol_hand.py

PR #384, opened 2016-12-07
[XmlExporter] possibility to provide an order for extension object me…

bitkeeper:
Improved version of adding a specific order to extension object members.

`_extobj_to_etree` now use a helper method `_get_member_order` to order the members.
 `_get_member_order` checks if the base datatype is located in a dict  `XmlExporter.extobj_ordered_elements`. If you find a standard (ns=0) ext obj that has a wrong member order just add the correct order to `XmlExporter.extobj_ordered_elements`.

Currently `XmlExporter.extobj_ordered_elements` is only intended for standard extobjs in ns=0, but `_get_member_order` can easily be modified if there is a better (generated) source present. For example that also provide information about custom extension objects.

### DIFF
--- a/opcua/common/xmlexporter.py
+++ b/opcua/common/xmlexporter.py
@@ -14,6 +14,17 @@ from opcua.common.ua_utils import get_base_data_type
 
 class XmlExporter(object):
 
+    ''' If it is required that for _extobj_to_etree members to the value should be written in a certain
+        order it can be added to the dictionary below.    
+    '''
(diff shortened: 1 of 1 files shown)


bitkeeper:
@oroulet continue where pr #380 stopped: first try to go a more generic solution for the ordering issue.

oroulet:
This looks ok! Should I merge?

bitkeeper:
In that case; yes please

PR #386, opened 2016-12-16
implement browse_next in client

oroulet:


oroulet:
added som extra changes that should not have been there, will need to revert these later, but anyway it should work

PR #368, opened 2016-11-18
Custom type methods with examples

levkevi:

Issue #293, opened 2016-08-23
[help] Retrieving event data with custom objects


Terrub:
This is related to my earlier question that now works using hardcoded values, but is not at all dynamic just yet.

I've got the inverse problem now that the call I make returns amongst other things, a custom object. This object is described in the same XML-dictionary document that describes the method. Though I should be able to retrieve the description dictionary if it wasn't, I haven't looked into it.

The code below is what I do now, but it feels like this is not at all the right way to do this.
So obviously the question is: What is the right way to do this?

```
def unpack_custom_data(p_custom_data):
    data = utils.Buffer(p_custom_data)
    measurements = []

    name = ua.unpack_uatype("CharArray", data)
    type = ua.unpack_uatype("Int32", data)
    num_measurements = ua.unpack_uatype("Int32", data)
    for idx in range(0, num_measurements):
        measurement = ua.unpack_uatype("Float", data)
        measurements.append(measurement)
    percentage = ua.unpack_uatype("Float", data)
```

zerox1212:
I don't really understand. Now that you can call the server method with custom objects you need to unpack what it returns?

Terrub:
I believe so. One of the returned values is another custom structure that needs to be decoded from a bytestring to something digestible. It is that value (or the Body actually) of that ExtensionObject that I decode. I should probably create a custom class and put this logic in the `from_binary` method but I am not sure I'm doing this the right way to start with.

oroulet:
Start by writing a simple function, then we look at where to place it

oroulet:
event start with a simple function with anode as arg and return the xml encoding. So we can test this different places. From you email the xml format representing the encoding sems the same one as the one used to generate the address space, so we can copy code from it later

oroulet:
The following code describes roughly what I do:

First I get the list of arguments from the method using the inputarguments node:

```
input_argument_node = method_node.get_child(["0:InputArguments"])
arguments = input_argument_node.get_value()
```

Then I check if the argument has an encoding associated.

for argument in arguments:
…

```
    encodings = opc_ua_client.get_node(argument.DataType).get_references(
        refs=ua.ObjectIds.HasEncoding,
        direction=ua.BrowseDirection.Forward
    )
```

If it has an encoding associated, I grab one of the options and get the node ID from it:

```
chosen_encoding = encodings[0]
encoding_node = opc_ua_client.get_node(chosen_encoding.NodeId)

```
[...]

Issue #287, opened 2016-08-17
[help] Calling a server method with custom inputArguments


Terrub:
I originally asked this helprequest in the mailing-list and oroulet instructed me to create an issue out of this for more exposure.
So, by his request, this is the thread:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/freeopcua/zwRRI_sQbw4/LjN-FrBWCgAJ

Any and all help would of course be much appreciated.

Right now I could still use some help with setting up the custom class to pass on as an argument.
And if anybody needs testing or intel on getting this functionality properly intergrated, I'll gladly offer my assistance.

Terrub:
Quick update. I got it somewhat working using hardcoded values here n there. But I'm sure I can eventually refactor it into something more variable.

I posted a reply with some example code in the maillist post.

zerox1212:
I looked at your code. I don't know much about extension object, but I agree with @oroulet last post about how to find the hierarchy.  From what you said about the server people being surprised about the XML dictionary I'm wondering if that XML in the encoding node description is just by chance. It sounds like some obscure spec that any custom structure should be described in XML. If the creator of the server generated their UA model, they probably didn't even realize it made that XML for them.

Terrub:
It does indeed seem to be standard OPC UA behaviour. As far as I understand they created these custom structures using certified UA tools ([UAModeller](https://www.unified-automation.com/products/development-tools/uamodeler.html) I believe)
They are using a different stack though, in case that makes a difference, although I hope not.

As said, I can find an example of the xml structure in the book I mentioned. (Page 67 fig. 2.34)

I haven't had a chance to test the method @oroulet mentioned yet but I am going to later this week. Sadly I'm multi-tasking different things atm.
I'll let you know what I find but I'm not entirely sure this will work as I'm looking at a structure here, not a set of nodes, so ordering by NodeId is going to be interesting.

oroulet:
I may be wrong here since we have not implemented struts yet. Maybe the spec requires to define structure order in xml with a hasEncoding reference...... Let me know

Terrub:
Alright I browsed through the specs a little and I found two rather interesting things.
Mind you I'm referring to the specifications found over at https://opcfoundation.org/developer-tools/specifications-unified-architecture and you need to have an account to view them. For now I'm going to assume that you have used these specs to build what you have already.

So the two parts I've found that are interesting are:
- Part 3: Address Space Model
[...]

PR #388, opened 2016-12-23
automatic unpack of custom extensionObjects(Structures) WORK IN PROGRESS

oroulet:
fixes:
https://github.com/FreeOpcUa/python-opcua/issues/373
https://github.com/FreeOpcUa/python-opcua/issues/293
https://github.com/FreeOpcUa/python-opcua/issues/287
and is necessary for :
https://github.com/FreeOpcUa/python-opcua/pull/368

This is a work in progress. Currently only parses the xml file and generate python code which can be compiled online or imported. we need to:

client:

- [x] add a method that parse custom types and registers new classes online

server:

- [ ] add a method that generates the xml (to be used in code and by opcua-modeler)
- [ ] add a method that create the necessary data types nodes


zerox1212:
I didn't look too closely at your implementation, but doesn't the spec say that structures can't have nulls?

zerox1212:
Looks interesting. Only request I would have is to add some doc strings. Especially to `StructGenerator` methods. Also to be clear, what happens if the struct changes on the server side? Exception? Crash?

oroulet:
This is not ready for release. There is an error, and performance is an issue.....

PR #396, opened 2017-01-17
fix support for intermediate chunks

### LINKED ISSUE #395: Receive intermediate chunks support
Looks like the receive of intermediate chunks (Part 6 chapter 6 and 7) support is broken.

When a client sends a message with intermediate chunks to server it results in an exception:
```python
ERROR:opcua.server.binary_server_asyncio:Exception raised while parsing message from client, closing
  File "/opt/python27/lib/python2.7/site-packages/opcua/server/binary_server_asyncio.py", line 85, in _process_data
    ret = self.processor.process(hdr, buf)
  File "/opt/python27/lib/python2.7/site-packages/opcua/server/uaprocessor.py", line 101, in process
    raise utils.ServiceError(ua.StatusCodes.BadTcpMessageTypeInvalid)
ServiceError: UA Service Error
```
Problem is that in `uaprocessor.py|UaProcessor.process` it is expected that `self._connection.receive_from_header_and_body(header, body)` always return a `Message`. In case of an Intermediate chunk that isn't true because returned but `None`, which results in a exception ua.StatusCodes.BadTcpMessageTypeInvalid.

The fix is easy;

Orginal: 
```python
 else:
            self.logger.warning("Unsupported message type: %s", header.MessageType)
            raise utils.ServiceError(ua.StatusCodes.BadTcpMessageTypeInvalid)

```
Fix:
```python
 elif msg is not None:  # ua.MessageType.SecureMessage and ChunkType.Intermediate
            self.logger.warning("Unsupported message type: %s", header.MessageType)
            raise utils.ServiceError(ua.StatusCodes.BadTcpMessageTypeInvalid)
```
PR will follow.

Btw hard to test because MaxMessageSize and MaxChunkSize are hardcoded.

bitkeeper:
Fixes #395.

Don't know how to make a regression test for it (see #95 for more info). For now used an external client ,which uses the intermediate chunks, to test this item.

### DIFF
--- a/opcua/server/uaprocessor.py
+++ b/opcua/server/uaprocessor.py
@@ -85,7 +85,6 @@ class UaProcessor(object):
 
             elif header.MessageType == ua.MessageType.SecureMessage:
                 return self.process_message(msg.SecurityHeader(), msg.SequenceHeader(), msg.body())
-
         elif isinstance(msg, ua.Hello):
             ack = ua.Acknowledge()
             ack.ReceiveBufferSize = msg.ReceiveBufferSize
@@ -95,8 +94,7 @@ class UaProcessor(object):
 
         elif isinstance(msg, ua.ErrorMessage):
             self.logger.warning("Received an error message type")
-
-        else:
+        elif msg is not None:  # ua.MessageType.SecureMessage and ChunkType.Intermediate
             self.logger.warning("Unsupported message type: %s", header.MessageType)
             raise utils.ServiceError(ua.StatusCodes.BadTcpMessageTypeInvalid)
         return True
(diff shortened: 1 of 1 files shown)


opcua/server/uaprocessor.py
```diff
@@ -95,8 +94,7 @@ def process(self, header, body):
 
         elif isinstance(msg, ua.ErrorMessage):
             self.logger.warning("Received an error message type")
-
-        else:
+        elif msg is not None:  # ua.MessageType.SecureMessage and ChunkType.Intermediate
             self.logger.warning("Unsupported message type: %s", header.MessageType)
             raise utils.ServiceError(ua.StatusCodes.BadTcpMessageTypeInvalid)
         return True
```
oroulet: This looks hard to follow.  We should have an "else" clause and explicitely test for the None case if it is something that happens. But if you really think that the current solution is better I trust you.

bitkeeper: I will change it to (don't like an additional `return True` in the `elif`):
```python
        elif msg is None:  
            pass # msg is a ChunkType.Intermediate of an ua.MessageType.SecureMessage
        else
            self.logger.warning("Unsupported message type: %s", header.MessageType)
            raise utils.ServiceError(ua.StatusCodes.BadTcpMessageTypeInvalid)
```

oroulet:
Ok

Issue #393, opened 2017-01-17
Missing ClientNonce in OpenSecureChannel message


markusgraube:
I am trying to connect to a OPC UA server on a Wago PLC. I am using freeopcua-0.90.1 under Python 3.5.2+. Other clients can connect well to the server.

However, using freeopcua leads to following error:
``` python
In [1]: from opcua import Client

In [2]: c = Client("opc.tcp://10.0.51.9:4840")

In [3]: c.connect()                           
Received an error: MessageAbort(error:StatusCode(BadOutOfMemory), reason:None)
Received an error: MessageAbort(error:StatusCode(BadOutOfMemory), reason:None)
---------------------------------------------------------------------------
TimeoutError                              Traceback (most recent call last)
<ipython-input-3-fe870a4d911f> in <module>()
----> 1 c.connect()

/home/mgraube/.local/lib/python3.5/site-packages/opcua/client/client.py in connect(self)
    225         self.connect_socket()
    226         self.send_hello()
--> 227         self.open_secure_channel()
    228         self.create_session()
    229         self.activate_session(username=self.server_url.username, password=self.server_url.password, certificate=self.user_certificate)

/home/mgraube/.local/lib/python3.5/site-packages/opcua/client/client.py in open_secure_channel(self, renew)
    267         nonce = utils.create_nonce(self.security_policy.symmetric_key_size)   # length should be equal to the length of key of symmetric encryption
    268         params.ClientNonce = nonce      # this nonce is used to create a symmetric key
--> 269         result = self.uaclient.open_secure_channel(params)
    270         self.security_policy.make_symmetric_key(nonce, result.ServerNonce)
    271         self.secure_channel_timeout = result.SecurityToken.RevisedLifetime

/home/mgraube/.local/lib/python3.5/site-packages/opcua/client/ua_client.py in open_secure_channel(self, params)
    222 
    223     def open_secure_channel(self, params):
--> 224         return self._uasocket.open_secure_channel(params)
    225 
    226     def close_secure_channel(self):

/home/mgraube/.local/lib/python3.5/site-packages/opcua/client/ua_client.py in open_secure_channel(self, params)
    163         # FIXME: we have a race condition here
    164         # we can get a packet with the new token id before we reach to store it..
--> 165         response = ua.OpenSecureChannelResponse.from_binary(future.result(self.timeout))
    166         response.ResponseHeader.ServiceResult.check()
    167         self._connection.set_channel(response.Parameters)

[...]

PR #398, opened 2017-01-20
Prevent nonce to be of length 0

markusgraube:
This is an idea of how to solve first problem of #393


oroulet:
This creates an api which does not behave as expected. We need to find something else

PR #407, opened 2017-02-12
Ignore file proposal

destogl:
Ignore .xml files generated by the tests. The files are renamed to have prefix _tmp_tests__ and ignore them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+tmp_test_*
 build*
 MANIFEST
 .idea*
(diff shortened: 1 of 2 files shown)


oroulet:
Ideally it would be great if you could put those file under tests/tmp
directory... But maybe it is too much work?

destogl:
I tried, but than was some paths issues in importer and exporter itself. I will check what would be the easiest way...

oroulet:
merging this one as is. We can alway find something better later

PR #416, opened 2017-03-06
[xmil_import] Fixed (S)Byte string_to_val

bitkeeper:
The xml import for the type Byte is broken, fixed ua_utils.string_to_val for Byte and SByte.

Step to reproduce:
- Add  xml_test_byte to the xml test suite

**Backgound:**
The current implementation treat Byte as single byte version of a ByteString, but Byte is just a nummeric value limited to 255 values. The same story for SByte.

In the https://opcfoundation.org/UA/2008/02/Types.xsd the definition is:
```xml
  <xs:element name="Byte" nillable="true" type="xs:unsignedByte" />
```

Where ByteString is:
```xml
  <xs:element name="ByteString" nillable="true" type="xs:base64Binary" />
```  

`xs:unsignedByte` is just a nummreic value without encoding, see for more info https://www.w3schools.com/xml/schema_dtypes_numeric.asp.

### DIFF
--- a/opcua/common/ua_utils.py
+++ b/opcua/common/ua_utils.py
@@ -78,15 +78,15 @@ def string_to_val(string, vtype):
             val = True
         else:
             val = False
-    elif vtype in (ua.VariantType.Int16, ua.VariantType.Int32, ua.VariantType.Int64):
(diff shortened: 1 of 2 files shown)


oroulet:
Sorry I missed that PR.

Issue #424, opened 2017-03-29
AnalogItemType with optional field EngineeringUnit cannot be parse from xml


XB32Z:
Hi, 

I have the following error
`File "C:\Python27\lib\site-packages\opcua\common\xmlimporter.py", line 234, in _set_attr
    for vtype, v2 in val:
ValueError: need more than 1 value to unpack`
when I am trying to import a xml such as:
```
<UAVariable DataType="EUInformation" ParentNodeId="ns=1;i=6076" NodeId="ns=1;i=6166" BrowseName="EngineeringUnits">
        <DisplayName>EngineeringUnits</DisplayName>
        <References>
            <Reference ReferenceType="HasProperty" IsForward="false">ns=1;i=6076</Reference>
            <Reference ReferenceType="HasModellingRule">i=78</Reference>
            <Reference ReferenceType="HasTypeDefinition">i=68</Reference>
        </References>
        <Value>
            <uax:ExtensionObject>
                <uax:TypeId>
                    <uax:Identifier>i=888</uax:Identifier>
                </uax:TypeId>
                <uax:Body>
                    <uax:EUInformation>
                        <uax:NamespaceUri>http://www.opcfoundation.org/UA/units/un/cefact</uax:NamespaceUri>
                        <uax:UnitId>5066068</uax:UnitId>
                        <uax:DisplayName>
                            <uax:Text>mm</uax:Text>
                        </uax:DisplayName>
                        <uax:Description>
                            <uax:Text>millimetre</uax:Text>
                        </uax:Description>
                    </uax:EUInformation>
                </uax:Body>
            </uax:ExtensionObject>
        </Value>
    </UAVariable>
```
This node is a property attached to a AnalogDataType variable.
Thx for your help

zerox1212:
I looked quickly and it requires debugging because it appears the importer is trying to make a list instead of an object.

XB32Z:
I added a pull request #425 to solve this issue. The val was instance of unicode not str
Thx for your support and for doing this :)

zerox1212:
I think you should try opening your XML with notepad++ and then save it as a UTF-8 encoding. Then try to import it.

PR #425, opened 2017-03-30
Fixed minor issue #424 EngeineeringUnits could not be parsed

XB32Z:
The value was not str but unicode
Pierre

### DIFF
--- a/opcua/common/xmlimporter.py
+++ b/opcua/common/xmlimporter.py
@@ -10,6 +10,10 @@ import opcua
 from opcua import ua
 from opcua.common import xmlparser
 
+import sys
+
+if sys.version_info.major > 2:
+    unicode = str
 
 class XmlImporter(object):
 
@@ -216,7 +220,7 @@ class XmlImporter(object):
         # tow possible values:
         # either we get value directly
         # or a dict if it s an object or a list
-        if isinstance(val, str) or isinstance(val, unicode):
+        if isinstance(val, (str, unicode)):
             pval = xmlparser.ua_type_to_python(val, obj.ua_types[attname])
             setattr(obj, attname, pval)
         else:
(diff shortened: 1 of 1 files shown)


zerox1212:
I will let @oroulet comment, but I think this should maybe be fixed in `xmlparser`.

oroulet:
But thanks for the fix, if you can make it work with python 3 without to much code, I will merge

zerox1212:
I think a better solution would probably be to add `.encode('utf-8)` some where or maybe the XML parser should enforce an encoding.

XB32Z:
Hi all, 

Sorry I didn't checked on PYthon3. I found a [solution ](http://stackoverflow.com/questions/4843173/how-to-check-if-type-of-a-variable-is-string) which includes importing another compatibility library (six) and replacing str by six.string_type. Does this fit you? Encoding the file in utf8 did not solved the issue. 

Best

oroulet:
om several files we have: 

if sys.version_info.major > 2:
    unicode = str


lets do the same here.
So you can write isintance(xxx, (str, unicode)) instead of or

zerox1212:
I would still like to know the root cause of why the parser would pass a unicode object instead of a string object to the importer.

oroulet:
You are right there might be something strange here.

XB32Z:
This field (<uax:Text>) can have values such as "m²" maybe the parser always reads it as unicode because it can have special characters?

zerox1212:
After some reading this should be fixed in the parser. Python 2.7 and Python 3+ have different versions of Etree which handle encoding differently.

@XB32Z can you attached your XML file? The encoding of an XML should always be declared at the top. This declaration is used by the parser internally.

XB32Z:
Cf quote in #424 
A simple node from AnalogDataType with EUEngineeringUnit properties will give the error

zerox1212:
Do you have this at the top of the file you are importing?
`<?xml version="1.0" encoding="UTF-8"?>`

XB32Z:
No I did not, however I added it and I still have the error

zerox1212:
If I get time I will look at how to make the Etree encoding work correctly for 2.7 and 3. For now your fix will work.

PR #434, opened 2017-04-13
Add ua to python mirroring example

zerox1212:
Please review this example and let me know if it is clear and worth adding to our collection. I found it very useful in my personal projects and would like to expand it.

Do not merge because I didn't test it yet.

Ultimate goal would be to generate python objects that subclass `UaObject` from UA Modeler for fast application development.

### DIFF
--- a/examples/server-ua-python-mirror.py
+++ b/examples/server-ua-python-mirror.py
@@ -11,8 +11,10 @@ from opcua import ua, Server
 
 # The advantage of this is that the software can be designed in a tool like UA Modeler. Then with minimal setup, a
 # python program will import the XML and mirror all the objects in the software design. After this mirroring is achieved
-# the user can focus on programming in python knowing that all all data from UA clients with reach the python object,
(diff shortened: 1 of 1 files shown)


oroulet:
I could not see anything bad with that approach. You could override __setattr__ so anytime a member is written it is written to server. Or use @property

zerox1212:
I thought about overriding `setattr`, but I was worried about performance. For example if the user wants to do some operations on a variable without lots of client updates. This way can publish it explicitly.

oroulet:
True. A complete class could have a batch parameter, that sync attributes on the fly or when calling a sync method.

oroulet:
btw calling the method write is probably more correct than publish since the call is called uawrite

zerox1212:
I will change this to `write` and test it so we can merge.

zerox1212:
This example works and is ready to merge. It could probably use some optimizations, but at least it's a start. Adding more helper objects to this SDK should make it easier to use.

Issue #426, opened 2017-04-03
Problems with custom enumeration from xml import


devel0peris:
Hey there.  I import the following XML into my server:
``` xml
...
    <UAObjectType BrowseName="0:CustomType" NodeId="i=20001">
        <DisplayName>CustomType</DisplayName>
        <Description>CustomType</Description>
        <References>
            <Reference IsForward="false" ReferenceType="HasSubtype">i=58</Reference>
            <Reference ReferenceType="HasProperty">i=20003</Reference>
            <Reference ReferenceType="HasProperty">i=20007</Reference>
        </References>
    </UAObjectType>
    <UADataType BrowseName="0:Prio" NodeId="i=20004">
        <DisplayName>Prio</DisplayName>
        <Description>Prio</Description>
        <References>
            <Reference IsForward="false" ReferenceType="HasSubtype">i=29</Reference>
            <Reference ReferenceType="HasProperty">i=20005</Reference>
        </References>
    </UADataType>
    <UAVariable ArrayDimensions="3" BrowseName="0:EnumStrings" DataType="LocalizedText" NodeId="i=20005" ParentNodeId="i=20004" ValueRank="-3">
        <DisplayName>EnumStrings</DisplayName>
        <Description>EnumStrings</Description>
        <References>
            <Reference IsForward="false" ReferenceType="HasProperty">i=20004</Reference>
            <Reference ReferenceType="HasTypeDefinition">i=68</Reference>
            <Reference ReferenceType="HasModellingRule">i=78</Reference>
        </References>
        <Value>
            <uax:ListOfLocalizedText>
                <uax:LocalizedText>
                    <uax:Text>Prio1</uax:Text>
                    <uax:Locale/>
                </uax:LocalizedText>
                <uax:LocalizedText>
                    <uax:Text>Prio2</uax:Text>
                    <uax:Locale/>
                </uax:LocalizedText>
                <uax:LocalizedText>
                    <uax:Text>Prio3</uax:Text>
                    <uax:Locale/>
                </uax:LocalizedText>
            </uax:ListOfLocalizedText>
        </Value>
    </UAVariable>
    <UAVariable BrowseName="0:prio" DataType="i=20004" NodeId="i=20007" ParentNodeId="i=20001">
        <DisplayName>prio</DisplayName>
        <Description>prio</Description>
        <References>
            <Reference IsForward="false" ReferenceType="HasProperty">i=20001</Reference>
[...]

PR #439, opened 2017-04-25
fix ua_utils.data_type_to_variant_type for enum types

bitkeeper:
See #426 about data_type_to_variant_type .

### DIFF
--- a/opcua/common/ua_utils.py
+++ b/opcua/common/ua_utils.py
@@ -203,16 +203,8 @@ def data_type_to_variant_type(dtype_node):
     if base.nodeid.Identifier != 29:
         return ua.VariantType(base.nodeid.Identifier)
     else:
-        # we have an enumeration, we need to look at child to find type
-        descs = dtype_node.get_children_descriptions()
-        bnames = [d.BrowseName.Name for d in descs]
-        if "EnumStrings" in bnames:
-            return ua.VariantType.LocalizedText
-        elif "EnumValues" in bnames:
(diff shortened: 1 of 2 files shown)


opcua/common/ua_utils.py
```diff
@@ -203,16 +203,8 @@ def data_type_to_variant_type(dtype_node):
...
-        bnames = [d.BrowseName.Name for d in descs]
-        if "EnumStrings" in bnames:
-            return ua.VariantType.LocalizedText
-        elif "EnumValues" in bnames:
-            return ua.VariantType.ExtensionObject
-        else:
-            raise ua.UaError("Enumeration must have a child node describing its type and values")
-
+        # we have an enumeration, value is a Int32
+        return ua.VariantType.Int32
```
oroulet: from what you wrote today, this is correct. I am curious to see the consequences of that change...

bitkeeper: I think within the project itself not much, because I couldn't find any actual use of that code except in the tests.

oroulet:
It is probably the modeler and client GUI that use that method. Maybe the
command line too

zerox1212:
Please test this change with modeler. If I remember correctly modeler uses it when you edit attributes of a variable or property node. There is a good chance you will find this method used in the widgets library.

oroulet:
I am looking at modeler/client

oroulet:
Seems to fix some bugs in modeller, could not get any crash . Thanks the analysis and patch @bitkeeper

PR #442, opened 2017-04-26
Fixed codegenerator due changes in the xmlparser

### LINKED ISSUE #440: Code generator broken ?
Today I wanted to make some changes to the codegenerator output, but the generator doesn't work.

Because the generator isn't part of the current tests; Can it be that the code generator (schemas/generate_address_space.py:CodeGenerator) is broken due changes to the xmlparser last year ?

bitkeeper:
resolve #440


oroulet:
Can't you just add the changes to the pr?

PR #445, opened 2017-05-01
Fix codegeneration list of localized text

bitkeeper:
Fixed as crash for list of localized strings (see #426).

Old implemenation generated something like:
```python
attrs.Value = ua.Variant(['Server', 'Client', 'ClientAndServer', 'DiscoveryServer'], ua.VariantType.LocalizedText)
```
Due a change in the Variant implemention this cause an exception on to_binary. New implementation generates:

```python
attrs.Value = [ua.LocalizedText("Server"),ua.LocalizedText("Client"),ua.LocalizedText("ClientAndServer"),ua.LocalizedText("DiscoveryServer")]
```


oroulet:
Sorry forgot these PR

PR #446, opened 2017-05-04
Only add HistoryRead after event history is enabled

tjerkk:
I noticed in UaExpert that the attribute EventNotifier always gets the values SubscribeToEvents and HistoryRead.

Creating an event source with the call get_event_generator, the attribute EventNotifier is always set with SubscribeToEvents and HistoryRead. Probably HistoryRead must only be set after calling iserver.enable_history_event() to enable the history keeping for events.


oroulet:
Make sense

PR #450, opened 2017-05-12
Stop server on start failure and prevent exception

tjerkk:
Created two fixes:

Starting the server failed when for eg a socket was already is use. The interval server was already started and caused the application to 'hang'. Fixed by stopping the internal server on an exception while starting the binary server. 

Noticed an exception on the server after an unsubscribe for data changes (by UaExpert)  on a node already deleted by the server. Added a check in delete_datachange_callback to prevent the KeyError exception.

### DIFF
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -634,8 +634,9 @@ class AddressSpace(object):
 
     def delete_datachange_callback(self, handle):
         with self._lock:
-            nodeid, attr = self._handle_to_attribute_map.pop(handle)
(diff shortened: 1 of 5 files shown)


tjerkk:
@oroulet I have added an extra test which failed on the first attempt. Fixed it by using an other port in the test (port_num + 1). Are tests running in parallel on Travic CI and is it no problem to use other network ports?

oroulet:
Tests are not run in parallel, and I think you can use whatever port you want (in the higher range, some need root if I remember well)

Issue #453, opened 2017-05-27
Client not terminating on TimeoutError


haru-a8n:
I am trying to stress test an OPC UA server pushing both CPU and memory usage. One of the error I am seeing which I think is expected is TimeoutError. The problem is that the script is not terminating. It seems like there is a thread stuck somewhere.

The is error I am expecting but the script is not closing:

2017-05-26 16:58:01,793|MainThread|C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\ua_client.py|close_session|255|INFO|close_session
**********
Traceback (most recent call last):
  File "opcclientstress.py", line 55, in <module>
    client.disconnect()
  File "C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\client.py", line 254, in disconnect
    self.close_session()
  File "C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\client.py", line 471, in close_session
    return self.uaclient.close_session(True)
  File "C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\ua_client.py", line 258, in close_session
    data = self._uasocket.send_request(request)
  File "C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\ua_client.py", line 76, in send_request
    data = future.result(self.timeout)
  File "C:\ProgramData\Miniconda3\lib\concurrent\futures\_base.py", line 407, in result
    raise TimeoutError()

haru-a8n:
Additional log below:

Disconnecting
2017-05-25 16:08:26,413|MainThread|C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\ua_client.py|close_session|255|INFO|close_session
**********
Traceback (most recent call last):
  File "opcclientstress.py", line 55, in <module>
    client.disconnect()
  File "C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\client.py", line 254, in disconnect
    self.close_session()
  File "C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\client.py", line 471, in close_session
    return self.uaclient.close_session(True)
  File "C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\ua_client.py", line 258, in close_session
    data = self._uasocket.send_request(request)
  File "C:\ProgramData\Miniconda3\lib\site-packages\opcua\client\ua_client.py", line 76, in send_request
    data = future.result(self.timeout)
[...]

PR #454, opened 2017-05-29
close socket even when timeout occurs

haru-a8n:
When TimeoutError is thrown during the call to disconnect, the socket remains open. This will case the appliation to deadlock

This will address issue https://github.com/FreeOpcUa/python-opcua/issues/453


oroulet:
This make sense. Btw thank you for your efforts to increase reliability og python-opcua

haru-a8n:
Let me know if I need to do something more to get this merged.

PR #461, opened 2017-06-23
Let the use call a method using a QualifiedName object.

bassofil:
Before the fix, calling the method using a QualifiedName object as argument was not working, moreover it made the server saturate the RAM on my PC.


oroulet:
The patch is ok. But I do not understand why a string should saturate your server...

bassofil:
I really don't know, but I tried to use `call_method` in the following way:
```
object.call_method(QualifiedName("method", namespace_idx))
```
and I had such (big) problem.

PR #463, opened 2017-07-03
Corrections for generating debian packages

cedricboudinet:
Hello

I changed the name option in setup.py to opcua (instead of freeopcua), so that the generated debian package filename will be python3-opcua.
Without this patch the package installation for python3-opcua-widgets fails (requesting python3-opcua to be installed).

Moreover, this is more in accordance with opcua-client and opcua-widgets packages naming.

Is this fine or must the package be called freeopcua?

Cedric

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ install_requires = ["python-dateutil", "pytz", "lxml"]
 if sys.version_info[0] < 3:
     install_requires.extend(["enum34", "trollius", "futures"])
 
-setup(name="freeopcua",
+setup(name="opcua",
       version="0.90.3",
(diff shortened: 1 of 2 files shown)


oroulet:
I had a look. I think it will break pip. Since the package is called freeopcua there. But this is history, maybe we could rename it to opcua... not sure

destogl:
Probably would be the bast to have the same name for pip and for importing, otherwise there are always some minor issues with that...

oroulet:
Ok. Med to find out if we can rename pip package or if we need to make a
new one and delete the old one

oroulet:
changing pip name might make trouble for many people...but I will try to create a new package

PR #473, opened 2017-07-13
Updated instanciate.py to handle DataType Nodes

XB32Z:
Hi, 
I had a bug creating an instance of a Type I imported via XML. Some Nodes where of NodeClass DataType which would trigger the warning l70 then crash l 93 because the object returned was None. 
I hope that my fix is ok with the specification. 
Thx

### DIFF
--- a/opcua/common/instantiate.py
+++ b/opcua/common/instantiate.py
@@ -63,6 +63,9 @@ def _instantiate_node(server, parentid, rdesc, nodeid, bname, dname=None, recurs
     elif rdesc.NodeClass in (ua.NodeClass.Method,):
         addnode.NodeClass = ua.NodeClass.Method
         _read_and_copy_attrs(node_type, ua.MethodAttributes(), addnode)
+    elif rdesc.NodeClass in (ua.NodeClass.DataType,):
+        addnode.NodeClass = ua.NodeClass.DataType
+        _read_and_copy_attrs(node_type, ua.DataTypeAttributes(), addnode)
     else:
         logger.warning("Instantiate: Node class not supported: %s", rdesc.NodeClass)
         return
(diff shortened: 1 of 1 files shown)


opcua/common/instantiate.py
```diff
@@ -63,6 +63,9 @@ def _instantiate_node(server, parentid, rdesc, nodeid, bname, dname=None, recurs
     elif rdesc.NodeClass in (ua.NodeClass.Method,):
         addnode.NodeClass = ua.NodeClass.Method
         _read_and_copy_attrs(node_type, ua.MethodAttributes(), addnode)
+    elif rdesc.NodeClass in (ua.NodeClass.DataType,):
+        addnode.NodeClass = ua.NodeClass.DataType
+        _read_and_copy_attrs(node_type, ua.DataTypeAttributes(), addnode)
```
oroulet: This is OK. It was probably never fully written

opcua/common/instantiate.py
```diff
@@ -86,7 +89,8 @@ def _instantiate_node(server, parentid, rdesc, nodeid, bname, dname=None, recurs
                         nodeids = _instantiate_node(server, res.AddedNodeId, c_rdesc, nodeid=ua.NodeId(identifier=inst_nodeid, namespaceidx=res.AddedNodeId.NamespaceIndex), bname=c_rdesc.BrowseName)
                     else:
                         nodeids = _instantiate_node(server, res.AddedNodeId, c_rdesc, nodeid=ua.NodeId(namespaceidx=res.AddedNodeId.NamespaceIndex), bname=c_rdesc.BrowseName)
-                    added_nodes.extend(nodeids)
+                    if nodeids is not None:
+                        added_nodes.extend(nodeids)
```
oroulet: when would this happen? I think that if _instantiate does not return a node then something wrong has happend and we should raise an exception (If instantiate method has not done it already)

XB32Z: That happened when the node where of DataType (cf previous edit). The condition went to the warning and return

oroulet: yes but now that you fixed datatype then it should return something? instantiate should retur the node it instantiate, if it does not, then there is a bug somewhere...

oroulet: Look in code of that method, it might be missing support for DataType...
then maybe we should raise an exception, not only show a warning....

XB32Z: Hi, 
Yeah that shouldn't happen now
I'll make an exception

oroulet:
thanks. we try that solution

PR #475, opened 2017-07-19
replaced type(value) == Type check with isinstance

AlexejStukov:


AlexejStukov:
Thanks @oroulet. That was fast!

PR #490, opened 2017-08-26
replace to/from_binary() methods by global to/from_binary(obj) functions

oroulet:
Instead of auto-generating to_binary and from_binary method code, we autogenerate a python friendly description of binary format and call global function to generate binaries. This reduces quite a lot the total amount of code, probably reduces memory use and prepare code to support xml format.
Need to be carefull with performance though.,,

Very invasive changes, need time to stabilize...

the to/from_binary function requires ua structure to have an attribute ua_types containing a list of members and they UA type

    ua_types = (
        ('ProtocolVersion', 'UInt32'),
        ('ReceiveBufferSize', 'UInt32'),
        ('SendBufferSize', 'UInt32'),
        ('MaxMessageSize', 'UInt32'),
        ('MaxChunkCount', 'UInt32'),
        ('EndpointUrl', 'String'),

Another member ua_switches might be present if some field are optional:

    ua_switches = {
        'SymbolicId': ('Encoding', 0),
        'NamespaceURI': ('Encoding', 1),
        'Locale': ('Encoding', 2),
        'LocalizedText': ('Encoding', 3),
        'AdditionalInfo': ('Encoding', 4),
        'InnerStatusCode': ('Encoding', 5),


destogl:
Hi @oroulet,

can this structured be generated from specs?

Just an idea....

oroulet:
What do you mean? The attributes ua_types and ua_switches are generated from the xml. They are also easier to write by hand if needed than the to_binary methods

destogl:
That is exactly what I meant :)

oroulet:
ready to be merged

oroulet:
does not seem to be any performance changes at running tests

zerox1212:
Does this impact when we pack some special types, like DateTime or UUID?

oroulet:
Should not, but changes are very invasive, so it would be great if you could test! Especially against other clients/servers with encryption @destogl  @zerox1212

PR #494, opened 2017-09-09
Improve support for custom structures client side

oroulet:
Added/improve method load_type_definition() that can be called to read custom structures from server

Example:
 
       client.connect()
       client.load_type_definitions()
       s_mgr = client.nodes.objects.get_child("2:SerialNumbersManager")


oroulet:
@levkevi please tag interested people. The implementation should be correct as spec. My only concern now is that after parsing the xml from server, we create a python file that we import. I will add a commit to do everything in memory by default, and only generate files when desired.

btw a big change in binary generation is to be found here: https://github.com/FreeOpcUa/python-opcua/pull/490 But it does not change the principles in this PR

oroulet:
merging. This fixes a lot of things anyway

PR #505, opened 2017-10-27
treat TimeoutHint=0 as mandated by OPC Unified Architecture Specification

helmut-jacob:
According to OPC Unified Architecture Specification (Part 4: Services,
7.28 RequestHeader) a TimeoutHint value of 0 shall be treated as no timeout:
"The value of 0 indicates no timeout.".

This fixes issues with OPC UA clients never receiving any PublishResponse
if a TimeoutHint of 0 was used in the according PublishRequest.

### DIFF
--- a/opcua/server/uaprocessor.py
+++ b/opcua/server/uaprocessor.py
@@ -65,7 +65,7 @@ class UaProcessor(object):
                                      len(self._publish_result_queue))
                     return
                 requestdata = self._publishdata_queue.pop(0)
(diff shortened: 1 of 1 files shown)


oroulet:
Correct!

helmut-jacob:
No idea why tests on pyhton 3.4 are failing. Running the tests locally with 3.6 works just fine ...

oroulet:
The VM was so slow to run tests than one went into a timeout. We sould increase timeout. I do not think there is a better solution to that. A client is waiting for an event and we cannot wait forever either.

Issue #508, opened 2017-11-06
Auto-generated address space code skips some entries


ipa-mdl:
The auto-generated code follows the order of the XML tags and therefore references some nodes before they have been registered to address space.

Example from Part10 code:

[ReadyToHalted](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/server/standard_address_space/standard_address_space_part10.py#L1665)  adds a `HasCause` [reference](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/server/standard_address_space/standard_address_space_part10.py#L1700) to [Halt](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/server/standard_address_space/standard_address_space_part10.py#L1703) node, which will get [added later](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/server/standard_address_space/standard_address_space_part10.py#L1801).
That's why [_add_reference](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/server/address_space.py#L326) fails silently.

First of all the generated code should check the returned status codes to detect these problems.

I see some ways of fixing the code ordering:
* sort the nodes topologically
* implement a lookup-or-create mechanism to ensure that all node will be created (in the the generated code) before used
* postpone calls to `self.make_refs_code` to be run after all other steps

The topological ordering would also harden the XML importer, which should have the same problem.
I can try to implement this.

ipa-mdl:
> Sorting the entire address space will probably be too expensive

Not the entire address space, but the result of [get_node_datas](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/common/xmlparser.py#L132), i.e. in the scope of one XML document.

> Unless you mean to just generate the .py file with a new order.

Yes, for the standard address spaces this will be done during generation of the scripts.

oroulet:
@ipa-mdl looks like you made a good analysis. There are two XML parsers one to generate code and one to import directly. Generally XML parsing is the most ugliest part of the stack, so if you manage to improve things it would be great! But postponing ref creation might be the easiest solution... The other methods might get into lock if there is a reference loop...

PR #509, opened 2017-11-06
Improved handling of references

ipa-mdl:
add test for generated address space (#508)
Prints all skipped references(currently 1370) and nodes (0).

### DIFF
--- a/opcua/common/manage_nodes.py
+++ b/opcua/common/manage_nodes.py
@@ -367,7 +367,7 @@ def _guess_datatype(variant):
         return ua.NodeId(getattr(ua.ObjectIds, variant.VariantType.name))
 
 
-def delete_nodes(server, nodes, recursive=False):
+def delete_nodes(server, nodes, recursive=False, delete_target_references=True):
     """
     Delete specified nodes. Optionally delete recursively all nodes with a
     downward hierachic references to the node
@@ -378,7 +378,7 @@ def delete_nodes(server, nodes, recursive=False):
     for mynode in nodes:
         it = ua.DeleteNodesItem()
         it.NodeId = mynode.nodeid
-        it.DeleteTargetReferences = True
+        it.DeleteTargetReferences = delete_target_references
         nodestodelete.append(it)
     params = ua.DeleteNodesParameters()
     params.NodesToDelete = nodestodelete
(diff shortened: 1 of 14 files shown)


ipa-mdl:
I have added the patches for the generator, but I cannot run it.
Even for the master I get
```
  File "generate_address_space.py", line 289, in <module>
    c.run()
  File "generate_address_space.py", line 51, in run
    self.make_variable_code(node)
  File "generate_address_space.py", line 186, in make_variable_code
    self.make_common_variable_code(indent, obj)
  File "generate_address_space.py", line 141, in make_common_variable_code
    self.make_ext_obj_code(indent, ext)
  File "generate_address_space.py", line 169, in make_ext_obj_code
    val = _to_val([extobj.objname], k, v)
  File "generate_address_space.py", line 19, in _to_val
    return ua_type_to_python(val, cls.ua_types[attr])
TypeError: list indices must be integers or slices, not str
```

oroulet:
Sorry. there was a missing update after a huge change a few weeks ago. Fixed now in master

oroulet:
your PR looks good. rebase and test and we can merge it

ipa-mdl:
Still 93 references left..

ipa-mdl:
@oroulet: part 3 and 5 depend on each other, any idea how to solve this?

ipa-mdl:
some parents get references before they got defined.

oroulet:
Looks like you already have the solution with your design. store refs in your class and call add_references after all imports..... Isn't it?

ipa-mdl:
> Looks like you already have the solution with your design. store refs in your class and call add_references after all imports..... Isn't it?

Yes, there seems to be no alternative to storing the failing refs.
I don't know why the other tests fail, even for the [master](https://travis-ci.org/FreeOpcUa/python-opcua/builds/298224143).

ipa-mdl:
@oroulet: ready for review :)

ipa-mdl:
bind problems: https://github.com/travis-ci/travis-ci/issues/8711

ipa-mdl:
For some reason sub types which are imported from XML might have duplicated properties (identtical nodeid),

ipa-mdl:
Summary:
- implemented fix for #508, plus in XML importer
- enforcing uniqueness of references
- add Node interfaces for adding and deleting references 
- minor improvements

oroulet:
Do the tests pass now?

ipa-mdl:
> Do the tests pass now?

Yes, but only with the IPv4 hack
This PR changes the overall behaviour as new references gets added  etc.

oroulet:
Looks good. My only comment is that I do not like asserts. Especially on things that are not critical like missing reference in addresses space
 Better to write a warning

ipa-mdl:
Thanks for merging!

> My only comment is that I do not like asserts.

I hopefully have used them for invariants only :)

PR #511, opened 2017-11-13
Don't instantiate nodes without a modelling rule

ipa-mdl:
From the OPC-UA specification Part 3 (6.2.2)
> If no ModellingRule exists then the Node is neither considered for instantiation of a type nor for
subtyping.

This PR implements this behavior.
Tool-generated nodesets should already contain the appoproriate modelling rules.

I have patched the `Node` API to add them for types' children as well (with opt-out).

### DIFF
--- a/opcua/common/instantiate.py
+++ b/opcua/common/instantiate.py
@@ -31,21 +31,14 @@ def instantiate(parent, node_type, nodeid=None, bname=None, dname=None, idx=0):
     elif isinstance(bname, str):
         bname = ua.QualifiedName.from_string(bname)
 
-    nodeids = _instantiate_node(parent.server, parent.nodeid, rdesc, nodeid, bname, dname=dname)
+    nodeids = _instantiate_node(parent.server, Node(parent.server, rdesc.NodeId), parent.nodeid, rdesc, nodeid, bname, dname=dname, toplevel=True)
     return [Node(parent.server, nid) for nid in nodeids]
 
 
-def _instantiate_node(server, parentid, rdesc, nodeid, bname, dname=None, recursive=True):
+def _instantiate_node(server, node_type, parentid, rdesc, nodeid, bname, dname=None, recursive=True, toplevel=False):
     """
     instantiate a node type under parent
     """
-    node_type = Node(server, rdesc.NodeId)
-    refs = node_type.get_referenced_nodes(refs=ua.ObjectIds.HasModellingRule)
-
-    # skip optional elements
-    if len(refs) == 1 and refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
-        return []
-
     addnode = ua.AddNodesItem()
     addnode.RequestedNewNodeId = nodeid
     addnode.BrowseName = bname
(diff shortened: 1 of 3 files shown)


opcua/common/instantiate.py
```diff
@@ -84,14 +77,18 @@ def _instantiate_node(server, parentid, rdesc, nodeid, bname, dname=None, recurs
             for c_rdesc in descs:
                 # skip items that already exists, prefer the 'lowest' one in object hierarchy
                 if not ua_utils.is_child_present(node, c_rdesc.BrowseName):
+                    c_node_type = Node(server, c_rdesc.NodeId)
+                    # skip optional elements, or node without ModellingRule at top-level
```
oroulet: should we skip or raise an exception?

opcua/common/node.py
```diff
@@ -617,14 +617,20 @@ def add_reference(self, target, reftype, forward=True, bidirectional=True):
...
+        return self
 
-    def add_object(self, nodeid, bname, objecttype=None):
-        return opcua.common.manage_nodes.create_object(self, nodeid, bname, objecttype)
+    def add_folder(self, nodeid, bname, mandatory=True):
+        return  opcua.common.manage_nodes.create_folder(self, nodeid, bname)._add_modelling_rule(self, mandatory)
 
-    def add_variable(self, nodeid, bname, val, varianttype=None, datatype=None):
-        return opcua.common.manage_nodes.create_variable(self, nodeid, bname, val, varianttype, datatype)
+    def add_object(self, nodeid, bname, objecttype=None, mandatory=True):
```
oroulet: Just a side comment. We need to be careful adding to many arguments to this methods. They were meant as a "dummy" way to add nodes. We should be carefull not to make them take 100 args.
What about making add_modelling_rule() public? instead of adding arguments?
But you can argument for adding this ' mandatory' arg.....

oroulet: Thiking about it. I think having modelling rule as optional is probably an exception. So I think we should really not add an argument but simply expose the add_modelling_rule() method

ipa-mdl: The idea was to keep it backwards compatible.
I can refactor it to be stand-alone function, but then the user has to call it every time a node gets added to a type.

Another option is to just remove the argument and add the mandatory rule to all type children..
(Actually, this was my first implementation ;))

We could expose a `set_modelling_rule(bool)` function that is able to remove or change it afterwards.

oroulet: Exactly. Add automatically the mandatory rule. and let users that do not want it call set_modelling_rule. One might complain that this is not efficient, but those that want efficiency call use the opcua calls directly

opcua/common/instantiate.py
ipa-mdl: The standard wants them to be excluded. There are use cases that needs this behavior, e.g. in Part 10 (Programs).
So nodes without modeling rule are static members of the type. For nodes with optional modellling the server might include or skip them.

oroulet: ok

oroulet: maybe log with logger.info...

ipa-mdl:
I have now implemented `set_modelling_rule`, which will remove all exisiting rules beforehand.
In addition skipping optional will show log messages.

I kept the new commits separate for review, I will squash/rebase them later.

Issue #492, opened 2017-09-06
MaxMessageSize and MaxChunkCount set to 0


jwbeck:
In opcua/ua/uaprotocol_hand.py, the Hello class initializes the max message size and max chunk count to 0. This made my client script not work with the server I was using. After changing these values to match what I was seeing with UaExpert (MaxMessageSize = 16777216 and MaxChunkCount = 5000), my script worked.

oroulet:
0 is definitely the correct value as spec. But some servers/clients are broken

jwbeck:
The part where it would fail was when I would do create_session() or get_endpoints(), my server would return an error, and I would get a timeout error in the script.

daniel-brosche:
I had also problems with the values with the Unified Automation HP Server. It should be possible to configure these values. So it is more flexible in cases of embedded ressource restricted environments.
As far as I understand the code it should be not complicated to pass this parameters. Could be realized with 0 as default parameter.

From Part 6:

> The maximum size for any response Message. The Server shall abort the Message
> with a Bad_ResponseTooLarge StatusCode if a response Message exceeds this
> value.
> The mechanism for aborting Messages is described fully in 6.7.3.
> The Message size is calculated using the unencrypted Message body.
> 
Later on in the Spec:
> A value of zero indicates that the Client has no limit.

Therefore 0 is a valid value but not every server accepts this value. I see benefit to make this parameter configurable.

oroulet:
maybe add it as argument to hello method to Uaclient class. Then people can call UaClient.hello(whatever) directly

daniel-brosche:
I have associated this line with the issue:

https://github.com/FreeOpcUa/python-opcua/blob/10a9064b2e2d6ec1b63cad1bff398d175d026cb6/opcua/client/client.py#L358

Is this correct?

jwbeck:
https://github.com/FreeOpcUa/python-opcua/blob/f935c993dc8fd4f650b71ebc33dacf192dc6af1f/opcua/ua/uaprotocol_hand.py#L20

oroulet:
yes make a PR with a very high size (Check what value uaexpert use for example) and add possibility to pass it as arg to UaClient.hello() method

daniel-brosche:
Ok, I will do that :+1:

PR #517, opened 2017-11-18
Added properties to configure MaxMessageSize and MaxChunkCount

daniel-brosche:
Solves the issue #492. Now it is possible to set MaxMessageSize and MaxChunkCount via properties. Servers that don't accept 0 can be connected now when the new properties are used with proper values before establishing a connection to a server  (in my case I used `client.set_max_messagesize(65536)`). 

**Important:** Currently only `client.send_hello` makes usage of the new introduced properties. It needs to be checked if other operations also should make usage of this (properties) for consistency reasons.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -114,6 +114,9 @@ class Client(object):
         self._session_counter = 1
         self.keepalive = None
         self.nodes = Shortcuts(self.uaclient)
+        self._maxMessageSize = 0 # No limits
+        self._maxChunkCount = 0 # No limits
+
 
     def __enter__(self):
         self.connect()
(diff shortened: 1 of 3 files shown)


opcua/client/client.py
```diff
@@ -148,6 +151,20 @@ def set_password(self, pwd):
...
+        Set the max messagesize for the connection.
+        """
+        self._maxMessageSize = maxMessageSize
+    
+
+    def set_max_chunkcount(self, maxChunkCount):
+        """
+        Set the max chunkcount for the connection.
+        """
+        self._maxChunkCount = maxChunkCount
```
oroulet: Not sure what is best practice here. Should we use set method or public properties. I think pep8 says public properties

daniel-brosche: I tried to align to the current style of the codebase (e.g. `client.set_password(...)`). 
So it seems that get / set methods fits well.
Currently the default values of `_maxMessageSize` and `_maxChunkCount` is zero (unlimited). Besides the fact that it is a valid value according to the opc ua specification it seems that 65536 is a better value regarding the related issue. 
So the client will work with problematic servers that don't handle the 0 value by default. Not sure if this maybe is a problem for other opc ua servers that are supported by the last release without any additional configuration.
Because of some doubts I decided to set the default value to 0 so that it is absolutely the same state as before. This means in my case I always need to set the values to 65536 before for a successful connection.
I'm willing to make some more modifications but I need some input.

oroulet: set_password is either an error or a will to make that property more visible. Please use public variable and i will merge

Issue #518, opened 2017-11-18
Unpacking Primitive arrays is slow


brubbel:
Possibly related to https://github.com/FreeOpcUa/python-opcua/issues/514

Using list comprehension for array unpacking is very slow, for example when reading OPC arrays from history. Profiling narrows it down to the following lines:
https://github.com/FreeOpcUa/python-opcua/blob/9e5a26a97f1079a5db1f55b365411c5e2b5d0317/opcua/ua/ua_binary.py#L203-L207

l'll create a PR with better performance, where reading 1536 x [8192] arrays from history drops from 60 to <6 seconds.

oroulet:
Any help/proposition on this is welcome

brubbel:
Thanks, could you also restart the checks in the above PR? Travis seems to fail at random points, which I suspect is due to timeouts while running the tests.

PR #519, opened 2017-11-18
fixed slow unpacking to Primitive1 array

brubbel:
Fix tested for python 2.7.6
https://github.com/FreeOpcUa/python-opcua/issues/518


brubbel:
Check fails at random points due to timeouts on test server...

oroulet:
How much does this improve performance? Can you compare also with .90.3?

brubbel:
For my specific test case: 24 seconds on 0.90.3 (cpu-limited) to 6 seconds (limited by trollius on Py2.7).

PR #528, opened 2017-11-28
Replace leftover prints with logging

Nebukadneza:
In various places in the library, there were some print()s. These make it hard to control the output of the library externally, e.g. when supressing logging output for clean testing-logs.

This commit attempts to replace all leftover print()s by logging calls.

### DIFF
--- a/opcua/server/binary_server_asyncio.py
+++ b/opcua/server/binary_server_asyncio.py
@@ -113,7 +113,7 @@ class BinaryServer(object):
             sockname = self._server.sockets[0].getsockname()
             self.hostname = sockname[0]
             self.port = sockname[1]
-        print('Listening on {0}:{1}'.format(self.hostname, self.port))
+        self.logger.warning('Listening on {0}:{1}'.format(self.hostname, self.port))
 
     def stop(self):
         self.logger.info("Closing asyncio socket server")
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -21,7 +21,7 @@ use_crypto = True
 try:
     from opcua.crypto import uacrypto
 except ImportError:
-    print("cryptography is not installed, use of crypto disabled")
+    logging.getLogger(__name__).warning("cryptography is not installed, use of crypto disabled")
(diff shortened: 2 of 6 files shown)


opcua/common/connection.py
```diff
@@ -62,7 +62,7 @@ def encrypted_size(self, plain_size):
         size = plain_size + self.security_policy.signature_size()
         pbs = self.security_policy.plain_block_size()
         if size % pbs != 0:
-            print("ENC", plain_size, size, pbs)
+            logger.debug("ENC", plain_size, size, pbs)
```
oroulet: This one should be removed, not changed to logging

opcua/server/binary_server_asyncio.py
```diff
@@ -113,7 +113,7 @@ def _process_data(self, data):
             sockname = self._server.sockets[0].getsockname()
             self.hostname = sockname[0]
             self.port = sockname[1]
-        print('Listening on {0}:{1}'.format(self.hostname, self.port))
+        self.logger.info('Listening on {0}:{1}'.format(self.hostname, self.port))
```
oroulet: Not sure what should be done here. I had it as a warning earlier. That is the kind of things one want to see when running it main thread but does not want when running in another thread or submodule...

Nebukadneza: I’d have assumed that, especially when not trying around with one of the example-code-pieces, you’d configure your logger to your liking. What could make sense is having it as a logger.warning, so it’s the only higher-severity line emitted by the BinaryServer, and users can more easily write down what they wanna see — also for example in the commandline server or examples.

How does that sound to you?

oroulet: ok make it a warning again.

ratara:
I want to reduce the verbosity of the server for my unit tests. As far as i can see, every class creates an own logger instance.

The only way to set the log level for all loggers seems to be to set the log level of the root logger before importing the ua.server.

Because:
grafik
 
and

grafik


Is this the plan?

PR #532, opened 2017-12-07
Fix handling of CharArray in Structs

ipa-mdl:
If methods are used with custom data types, which contain strings, UAExpert encodes these as `CharArray`.

However, the current struct handling packs/unpacks only Variant types, which cover only a subset of the standard types for binary encoding.
This PR considers the Primitives types as well; Variant types take precedence.

In addition the server now exposes `load_type_definitions` (same as for the client) that can be called to setup extention object class automatically at server-side.

Without this fix the `CharArray` is handled as a structure and the methods call will never get a response due to an exception in between.

### DIFF
--- a/opcua/server/server.py
+++ b/opcua/server/server.py
@@ -22,6 +22,7 @@ from opcua.client.client import Client
 from opcua.crypto import security_policies
 from opcua.common.event_objects import BaseEvent
 from opcua.common.shortcuts import Shortcuts
+from opcua.common.structures import load_type_definitions
 from opcua.common.xmlexporter import XmlExporter
 from opcua.common.xmlimporter import XmlImporter
 from opcua.common.ua_utils import get_nodes_of_namespace
@@ -278,8 +279,8 @@ class Server(object):
             self.bserver.start()
         except Exception as exp:
             self.iserver.stop()
-            raise exp        
-        
+            raise exp
+
 
     def stop(self):
         """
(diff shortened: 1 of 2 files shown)


oroulet:
Great to see that you are looking at that kind of things! It i really new. Does it work for you? The server is still missing some methods to create and register custom sructures (It is on the plan, I even have some hours allocated for that but I am waiting for a go from customer)

ipa-mdl:
> Does it work for you?

It works. But it does not support custom enuma in custom data types.
That's why I wrote the classes for my custom types myself.

It's easy to register them:
```python
setattr(opcua.ua, "MyEnum", MyEnumImpl) # we should hide this hack in a function!
opcua.ua.register_extension_object("MyDataType", opcua.ua.NodeId.from_string("ns=1;i=1234"), MyDataTypeImpl)
```

(`MyEnum` is derived from `IntEnum`, `MyDataType ` is an object with `ua_types` array)

calling `load_type_definitions` automates generatation and registration for custom structures..

oroulet:
>> setattr(opcua.ua, "MyEnum", MyEnumImpl) # we should hide this hack in a function!

Yes. this is in fact one the things, the register_extension_objects and load_type_definitions do

If you have any ideas to improve support for custom enums, then a PR is welcome ;-). maybe simply a register_custom_enum (But their might be other types I did not think of.....)

ipa-mdl:
I have extended structures.py to support enums as well, but it I decided to not mix it into this bug fix.
I can share the extended code as well (can't access it now), but It is more a hack than a proper implementation. IMHO the current extension system needs to be refactored somehow to offer more flexibilty.
But this is beyond the scope of this PR ;)

oroulet:
>> IMHO the current extension system needs to be refactored somehow to offer more flexibilty
I just got refactored ;-( obviously I forgot some use cases...

ipa-mdl:
I will push my branch later and open an issue to discuss if we want the hack or need to implement a more elaborated approach.

PR #534, opened 2017-12-11
Reap Client-KeepAlive Threads for reconnect

Nebukadneza:
> In some situations, Clients seem to create multiple KeepAlive-Threads.
> The old ones are never properly stopped, and there are no references to
> them to be able to stop them. This causes programs not to quit properly
> or completely.
> 
> To Redeem, we now check for an existing KeepAlive-Thread whenever
> creating a new one — and try to stop the existing one beforehand.

We discovered this when running some combinations of clients in our test-bed. The tests would not terminate because some KeepAlive-Threads were leftover …

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -374,6 +374,11 @@ class Client(object):
         ep = Client.find_endpoint(response.ServerEndpoints, self.security_policy.Mode, self.security_policy.URI)
         self._policy_ids = ep.UserIdentityTokens
         self.session_timeout = response.RevisedSessionTimeout
+        # In case there was an existing session, try to reap existing KeepAlive-Thread
+        try:
+            self.keepalive.stop()
+        except AttributeError:
+            pass
         self.keepalive = KeepAlive(self, min(self.session_timeout, self.secure_channel_timeout) * 0.7)  # 0.7 is from spec
         self.keepalive.start()
         return response


zerox1212:
Might be better to put a message to `logger.info` instead of `pass`.

Nebukadneza:
My assumption was that the „pass“-case is the usual case, where the client opens his first and only connection. In this case, no old KeepAlive-Thread has to be explicitly stopped. If possible, I’d rather notify the user that we proactively stopped a Thread, which is if the try-case.

How would you feel about a small notification in the try-block, and leaving the except on pass, as ugly as it seems as first … ?

Thanks!

zerox1212:
It's up to @oroulet in that case.

Nebukadneza:
Oki, let’s wait for him. Thanks for your opinion nevertheless ^_^.

oroulet:
How can this happen? You must be calling two times create_session without closing it, isn't it?

oroulet:
BTW, this is anyway not correct way to solve that kind of things. You should check if object is not none, and if thread is running, not try to stop it under a try/catch

Nebukadneza:
Oh … of course you are right, in both ways.

I have made 2 mistakes here: First, i overlooked that self.keepalive is initialized as None in __init__ — so i could have indeed checked for None. Second, i was tracing for the calls of the session creation, but it got lost a little in the depths of the stacktrace of our system. Yes, _*I*_ did call twice, where i was assuming it happened due to a reconnect.

I’m sorry i misunderstood what i was seeing, as i made the wrong assumption there.

Thanks for looking into this anyways, closing now.

PR #537, opened 2017-12-22
fix typo: “sesssion”

jonemo:
Very minor change to fix three mis-spellings of "session", twice in log statements and once in a docstring.

Just to be paranoid, I wanted to run tests after these changes, and found that the Readme instructions for runnings tests are outdated. Hence, I changed those too.


oroulet:
Sorry i forgot that one. Merging now

PR #545, opened 2018-01-15
During call to connect, disconnect socket when create_session fails

harua8n:
Under certain conditions, normally when the server is under heavy load, it may fail creating session, if this happens we need to close the socket.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -207,8 +207,8 @@ class Client(object):
         """
         Connect, ask server for endpoints, and disconnect
         """
+        self.connect_socket()
         try:
-            self.connect_socket()
             self.send_hello()
             self.open_secure_channel()
             endpoints = self.get_endpoints()
@@ -221,8 +221,8 @@ class Client(object):
         """
         Connect, ask server for a list of known servers, and disconnect
         """
+        self.connect_socket()
         try:
-            self.connect_socket()
             self.send_hello()
             self.open_secure_channel()  # spec says it should not be necessary to open channel
             servers = self.find_servers()
@@ -235,8 +235,8 @@ class Client(object):
         """
         Connect, ask server for a list of known servers on network, and disconnect
         """
+        self.connect_socket()
         try:
-            self.connect_socket()
             self.send_hello()
             self.open_secure_channel()
             servers = self.find_servers_on_network()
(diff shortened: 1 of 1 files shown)


oroulet:
Shouldn't we put all socket operations under try catch then? Just wondering...

harua8n:
That might be the case but I can't test other scenarios though. If you want me I can do it.

harua8n:
@oroulet, as discussed I have wrapped socket operations in try catch. Kindly review.

harua8n:
Kindly review and let me know for inputs.

opcua/client/client.py
```diff
@@ -206,36 +207,42 @@ def connect_and_get_server_endpoints(self):
...
         Connect, ask server for a list of known servers, and disconnect
         """
-        self.connect_socket()
-        self.send_hello()
-        self.open_secure_channel()  # spec says it should not be necessary to open channel
-        servers = self.find_servers()
-        self.close_secure_channel()
-        self.disconnect_socket()
+        try:
+            self.connect_socket()
```
oroulet: Not sure if that one should be before or after try: ..... Before i guess. If it fails then no socket is created and no need to close it

opcua/client/client.py
```diff
@@ -246,7 +253,11 @@ def connect(self):
         self.connect_socket()
         self.send_hello()
```
oroulet: This one and the following line should be under try too. Isn't it?

oroulet:
Sorry i missed it. Added some comments now

harua8n:
Made changes as requested, kindly review again.

opcua/client/client.py
```diff
@@ -6,6 +6,7 @@
 except ImportError:  # support for python2
     from urlparse import urlparse
 
+from concurrent.futures import _base
```
oroulet: why did you add that import? otherwise it looks ok

harua8n: In line 258 (call to self.create_session()), can sometimes fail and I need to clean-up socket connection if this happens.

harua8n: @oroulet, any additional thoughts? I can remove it but that means the user needs to know to clean-up on failure outside of the class.

harua8n: @oroulet, I have submitted new PR without importing concurrent.futures, kindly review.

harua8n:
@oroulet, can you please review my latest changes?

Issue #548, opened 2018-01-18
URI being registered before user is able to change it.


JuanoD:
You are supposed to use the method server.set_application_uri("blablabla") to set ehe application uri, but in the __init__ method, it registers the default uri before you can change it, causing an extra uri to appear when you do server.get_namespace_array()

```python
        def __init__(self, shelffile=None, iserver=None):
        self.logger = logging.getLogger(__name__)
        self.endpoint = urlparse("opc.tcp://0.0.0.0:4840/freeopcua/server/")
        self.application_uri = "urn:freeopcua:python:server" # This one here will always be registered
        self.product_uri = "urn:freeopcua.github.no:python:server"
        self.name = "FreeOpcUa Python Server"
        self.application_type = ua.ApplicationType.ClientAndServer
        self.default_timeout = 3600000
        if iserver is not None:
            self.iserver = iserver
        else:
            self.iserver = InternalServer(shelffile)
        self.bserver = None
        self._discovery_clients = {}
        self._discovery_period = 60
        self.certificate = None
        self.private_key = None
        self._policies = []
        self.nodes = Shortcuts(self.iserver.isession)

        # setup some expected values
        self.register_namespace(self.application_uri) # Because this is here, inside the __init__
        sa_node = self.get_node(ua.NodeId(ua.ObjectIds.Server_ServerArray))
        sa_node.set_value([self.application_uri])
```
Result of the server.get_namespace_array()
```python
    ['http://opcfoundation.org/UA/', 'urn:freeopcua:python:server', 'http://opcfoundation.org/UA/DI/', 'http://fdi-cooperation.com/OPCUA/FDI5/', 'my:uri']
```
I think it would be better to enforce the user to register the uri after creating the server.
My main trouble here is I have required uris in my xml model, so I need to register them before my application uri.

I'll locally edit the library so I can have the uri as an argument on server creation, then I'll register my first required uri, and after I finish the xml importing, I'll change the uri to the one I want my server to have.

```python
    def __init__(self, desired_uri="urn:freeopcua:python:server", shelffile=None, iserver=None):
        self.logger = logging.getLogger(__name__)
        self.endpoint = urlparse("opc.tcp://0.0.0.0:4840/freeopcua/server/")
        self.application_uri = desired_uri
```
And then
```python
    server = Server(desired_uri="http://opcfoundation.org/UA/DI/")

    print(str(datetime.now()), "-> 0/3")
[...]

PR #553, opened 2018-01-31
fix for ordering of application_uri in namespace array

oroulet:


JuanoD:
As stated in #548, I think it is better for the application uri to be the last uri, since xml imports would register uri in import order, and I always import my uri after all required nodesets have been registered. Maybe this is not necessary, but I can always change it on library.

oroulet:
I think this is common practice (if not standard) to have application uri
as index 1. Anyway in you application you should not hardcode index but
query server for index given uri

PR #566, opened 2018-03-19
refactoring

maxim-s-barabash:
separation OPCUAProtocol and BinaryServer


oroulet:
Looks good. Thanks for the fix of that long standing uglyness. Can i ask how did you end up in that file?

maxim-s-barabash:
I needed to start listening on all interfaces that differed from the endpoint url. Exploring the Server / InternalServer / BinaryServer chain

Issue #572, opened 2018-04-05
Updating address space to 1.04


cirp-usf:
Hi,

In the project I'm working we need the 1.04 version of the standard address space. So I regenerated the files in [opcua/server/standard_address_space/](../tree/master/opcua/server/standard_address_space/) using [schemas/generate_address_space.py](../tree/master/schemas/generate_address_space.py) and the new NodeSet files. I needed to  manually reorder one of the generated files.

The generated files allow me to import the NodeSet of the [specification] I’m working with and I’ve been using them ever since.

I would like to contribute the generated files back. However now I realize other files may need to be changed as well, for instance [opcua/ua/object_ids.py](../tree/master/opcua/ua/object_ids.py). 
In order to make a PR, do I have to update other files as well? If so, which files must be updated and how?

Best regards,
Ubiratan

[specification]: http://www.euromap.org/en/euromap77

oroulet:
There is a file called download address space in schema. Then when calling generate_protocol.py, all these files should be updated

oroulet:
Make a pr anyway so we can see and discuss issues. Concerning testing. You can run python3 /tests/tests.py

oroulet:
It is automatically run when you create a pr btw...

cirp-usf:
PR made. Didn't pass the tests. I'll look at it.

cirp-usf:
I saw only 4 failures at the output of the tests. They are all related to the _AttributeWriteMask_ standard type. In version 1.04, it is defined in part 3 (8.20, p. 73) as a subtype of UInt32. I could not find it in 1.03, but in `schemas/Opc.Ua.NodeSet2.Part4.xml` of 1.03 this type was defined as a subtype of _Enumeration_. Two test files (`tests_common.py`, line 927 and `tests_xml.py`, line 305) explicitly attribute the Int32 data type to _AttributeWriteMask_. Those are the tests that fail.

Would it be OK to change those lines to use UInt32? With that change those tests pass, but I think that this would make the test in `tests_xml.py` pointless. This test is called `test_xml_enumvalues`, which _AttributeWriteMask_ no longer is.

oroulet:
OK I had a loook and you can change type. This does not change anything in binary protocol

PR #576, opened 2018-04-13
Standard address space  spec version 1.04 

cirp-usf:
For previous discussion, see #572 and #574


cirp-usf:
When opcua is imported, the failure to include some nodes at the first try generate some annoying warnings:
```
add_node: while adding node NumericNodeId(i=15957), requested parent node NumericNodeId(i=11715) does not exists
add_node: while adding node NumericNodeId(i=15958), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15959), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15960), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15961), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15962), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15963), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15964), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=16134), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=16135), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=16136), requested parent node NumericNodeId(i=15957) does not exists
```
The code passes the tests however.

cirp-usf:
When opcua is imported, the failure to include some nodes at the first try generate some annoying warnings:
```
add_node: while adding node NumericNodeId(i=15957), requested parent node NumericNodeId(i=11715) does not exists
add_node: while adding node NumericNodeId(i=15958), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15959), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15960), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15961), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15962), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15963), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=15964), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=16134), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=16135), requested parent node NumericNodeId(i=15957) does not exists
add_node: while adding node NumericNodeId(i=16136), requested parent node NumericNodeId(i=15957) does not exists
```
The code passes the tests however.

oroulet:
Clean and correct patch! Well done

oroulet:
Concerning warning. What is happening when a node is not inserted? The insertion is postponed? What happens if it fails the second time?

oroulet:
The easiest solution is to change warning to info. so it does not get printed by default. 
It is also possible to check if parent exist before trying to add the node, but I am afraid to add one more nodeid dict query here. It is already very slow... but maybe this does not change much...

cirp-usf:
I really don't like the warnings getting printed for the standard address space. But I'm afraid changing the warning to info would make it very difficult to find bugs when there actually is a problem with other xml node sets or nodes the user is trying to include. The warning is emitted [here](https://github.com/FreeOpcUa/python-opcua/blob/72629f5e80573563a2d8a08467bbb7a6244b85c4/opcua/server/address_space.py#L213). 

I'm calling the method `_add_node` directly from `try_add_nodes` instead of calling `add_nodes`. One possibility would be to add an extra parameter to `_add_node` that would make it emit an info instead of an warning. The parameter's default value would cause warnings to be emitted. Then I would make `try_add_nodes` use that parameter so when the standard address space is loaded there are no warnings. But when the user adds nodes via the usual interface, a warning would be emitted if the insertion fails.

Would such a solution be acceptable? This PR is so nice and organized. But I think I may need to add yet another commit...

oroulet:
Info is no problem. Anyway an explicit ua error status is returned

oroulet:
You can add another commit to fix warning

oroulet:
Thanks for good work

Issue #332, opened 2016-10-15
DataChangeFilter with trigger to timestamp.


brubbel:
Hi,
I'm fairly new to OPC UA and was looking into installing a custom DataChangeFilter to receive data even when the value has not changed.
Simple example:
```
        while True:
            time.sleep(1)
            count += 0.1
            myvar.set_value(round(count))
```
However, looking at the code in address_space.py, it seems that identical values are already pre-filtered before they even reach the datachange_callbacks():
```
if old.Value != value.Value:  # only send call callback when a value change has happend
    cbs = list(attval.datachange_callbacks.items())
```
[link to code section](https://github.com/FreeOpcUa/python-opcua/blob/1c6e5d26b5551cff19df73dc9511db1694abe4b9/opcua/server/address_space.py#L568-L569)

So, could it be that installing a filter which triggers on a timestamp or a zero deadband will fail in advance?

oroulet:
will fail. neither the spec not the code support such a thing. what are you trying to achieve?

brubbel:
Ok, I see that opc does not support repeated values. 
My use case is a wireless sensor which sends regular 
updates, which could have the same value. The way opc works masks the difference between a connection loss and a static value, but I guess the appropriate solution will be to combine a 'last seen' value along with the measurement.
However, this somewhat defeats the purpose of minimizing traffic.

oroulet:
The opcua foundation is now working on a real-time extension to opcua that
is not real-time at all since based on UDP/amqp but it will solve that kind
of cases. But someone has to implement this in python.... Or pay someone to
do it..

oroulet:
Btw You can use events. That will work fine, and you can discover lost
events and do history query

brubbel:
Thanks! I think this issue can be closed.
I may append some info later on, when I have a better view on opc-ua with a few alternative options as a reference for others with the same question.

zerox1212:
FYI

If your wireless sensor is where the OPC UA server lives, you should be able to just call node.set_value() with the same value and it will trigger client updates. In your example you can test it easily be seeing if the node's timestamp is changing every second.

PR #579, opened 2018-04-23
Respect monitored item trigger (DataChangeFilter)

draity:
As already mentioned in #332 according to https://forum.unified-automation.com/post282.html the `DataChangeFilter` allows clients to get subscription updates even if the value does not change. I need this kind of functionality because of legacy components, so I can't switch to e.g. Events. 

To achieve this I removed the filter for identical values and only avoid calling the callbacks of a value if the timestamps are exactly the same.  The actual filtering based on the `DataChangeFilter` is done similar to the `deadband` filter. I added a new subscription method which enables clients, tools and tests to subscribe with a `DataChangeFilter` of `StatusValueTimestamp`.

Please let me know if you have any doubts.

### DIFF
--- a/opcua/common/subscription.py
+++ b/opcua/common/subscription.py
@@ -177,6 +177,17 @@ class Subscription(object):
         """
         return self._subscribe(nodes, attr, queuesize=0)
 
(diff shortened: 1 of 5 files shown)


oroulet:
That one is tricky... Need to think about it..

PR #583, opened 2018-05-01
instantiation of optional nodes as parameter, do not set modelling rule automatically 

oroulet:
as implementation was too simple and sometimes wrong


ipa-mdl:
@oroulet: Is there any way to prevent optional nodes to be created from `Node` interface?

oroulet:
Not from Node. But it is not hard to use the instantiate function

PR #588, opened 2018-05-08
Allow importing XML from string

helmut-jacob:
Introduce an optional parameter xmlstring throughout the import_xml
chain (client, server, xmlimporter, xmlparser) to allow specifying
a XML string instead of a path.

If both are given the string takes precendence.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -538,12 +538,12 @@ class Client(object):
     def delete_nodes(self, nodes, recursive=False):
         return delete_nodes(self.uaclient, nodes, recursive)
 
-    def import_xml(self, path):
+    def import_xml(self, path=None, xmlstring=None):
         """
         Import nodes defined in xml
         """
         importer = XmlImporter(self)
-        return importer.import_xml(path)
+        return importer.import_xml(path, xmlstring)
 
     def export_xml(self, nodes, path):
         """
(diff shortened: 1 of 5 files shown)


oroulet:
I agree that having the possibility to path string is interesting but I am wondering about the API. I think that usually methods take an argument which is either a path or a string. but having two arguments where only one is used seems a bit strange...

helmut-jacob:
Thought about that too but how would we differentiate between path and xmlstring?
Something like: Interpret as XML string and if parsing fails try to use the argument as path?
Or the other way around: Try to parse as path and fall back to parsing the string as XML?
Any preferences?

oroulet:
You have a point. We should check how other libraries do...

helmut-jacob:
What do you think about checking for file presence first and then falling back to interpreting as XML string input?

oroulet:
Well we could just check if the string ends with . xml or start with <. Should be faster and even more reliable?

helmut-jacob:
A unix filename is valid even if it starts with "<", hence the explicit check for file existence.

helmut-jacob:
Good point. default behavior needs to stay the same.

Checking for ".xml" or "<?xml" doesn't sound sane to me. Hence, I'd vote for the previous approach and use an additional optional parameter and additionally make the xmlpath a named parameter.

Behavior stays the same for existing callers.
New callers can explicitely use xmlpath='foo' or xmlstring='bar'.

oroulet:
We can wait a bit before merging in case someone comes up with a smart solution...

oroulet:
Looks ok. A small detail: no space around = for default arg values

PR #597, opened 2018-05-22
Parsing of custom structures that contain custom enums

jobasto:
This commit implements parsing of custom enumerations so that it is now possible to import structure data types from a server via load_type_definitions() that reference these enumerations. Attached is an example of a server that exposes such a data type definition and a variable node using it via its address space. Additionally, there is a client that won't be able to parse the data without the patch.

[server-example-enum-struct.py.txt](https://github.com/FreeOpcUa/python-opcua/files/2027270/server-example-enum-struct.py.txt) - server with custom enums and structs
[struct_enum_example.xml.txt](https://github.com/FreeOpcUa/python-opcua/files/2027273/struct_enum_example.xml.txt) - xml file that is loaded by the server
[client-example-enum-struct.py.txt](https://github.com/FreeOpcUa/python-opcua/files/2027272/client-example-enum-struct.py.txt) - client


jobasto:
I've opened a new pull request because I was not happy with some information in the history.

oroulet:
You can always rebase to edit history and force push

PR #598, opened 2018-05-23
Parsing of custom structures that contain custom enums

jobasto:
This commit implements parsing of custom enumerations so that it is now possible to import structure data types from a server via load_type_definitions() that reference these enumerations.

Please see the tests for further details:
In tests_client.py/TestClient/setUpClass a new address space is loaded by calling add_server_custom_enum_struct. It loads an XML file which defines a structure "ExampleStruct" which itself contains an enum of type "ExampleEnum". In tests_client.py/TestClient/test_custom_enum_struct the server's custom type definitions are loaded by calling load_type_definitions(). The value of a node is fetched which uses the data type "ExampleStruct". This fails if this patch is not implemented because no instance of an "ExampleStruct" can be instantiated. With the patch, however, the test succeeds.


jobasto:
@oroulet Did the included tests clarify what I want to achieve or do you need more information? Please let me know if anything is unclear.

oroulet:
@jobasto looks good. Unfortunately I merged other simple pr and they have made a conflict. Can you rebase? Thanks

jobasto:
@oroulet Thanks a lot.

PR #607, opened 2018-06-02
option to disable anonymous connections to server

kleskjr:
A small change that makes choosing the type of encryption a bit more configurable. The biggest change is the introduction of the property _nosecurity_ whit default value of True to maintain compatability.

### DIFF
--- a/opcua/server/server.py
+++ b/opcua/server/server.py
@@ -99,6 +99,12 @@ class Server(object):
         sa_node = self.get_node(ua.NodeId(ua.ObjectIds.Server_ServerArray))
         sa_node.set_value([self._application_uri])
 
+        self.security_endpoints = ["None", "Basic128Rsa15_Sign",
+                                    "Basic128Rsa15_SignAndEncrypt", 
+                                    "Basic256_Sign", "Basic256_SignAndEncrypt"]
+        self.policyIDs = ["Anonymous", "Basic256", "Basic128", "Username"]
+
+
     def __enter__(self):
         self.start()
         return self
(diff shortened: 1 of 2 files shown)


oroulet:
Hi, Thanks for the PR.

I use very little encryption but I remember a BR explaining that it was hard to disable anonymous/no encryption and I always struggle to remember how this work.... Maybe it is time to fix this API and your PR seems a step in the good direction

As far as I understand your PR you expect users to setup the new property security_endpoints? maybe we should make it private and make a method with a docstring to increase visibility and documentation?

Do we really need this nosecurity parameter? Can't we just have a None in self.security endpojnts? ( I do not remember that code so I may be completely wrong...)

kleskjr:
thanks for the comments, oroulet!
Your suggestion about _None_ in _self.security endpojnts sounds_ sounds more elegant than the solution in the PR. I will push the change later today with a docstring description. I am opening another issue concerning authentication and trusted certificates.

opcua/server/server.py
```diff
@@ -74,7 +74,7 @@ class Server(object):
 
     """
 
-    def __init__(self, shelffile=None, iserver=None):
+    def __init__(self, shelffile=None, iserver=None, nosecurity=True):
```
oroulet: You forgot that line ;-)

kleskjr: yes, now I squashed all the changes into a single commit to decrease the noise. Hope it's fine now

opcua/server/server.py
```diff
@@ -201,54 +207,68 @@ def get_endpoints(self):
 
     def _setup_server_nodes(self):
         # to be called just before starting server since it needs all parameters to be setup
-        self._set_endpoints()
-        self._policies = [ua.SecurityPolicyFactory()]
+        if "None" in self.security_endpoints:
+            self._set_endpoints()
+            self._policies = [ua.SecurityPolicyFactory()]
+
         if self.certificate and self.private_key:
```
oroulet: we should replace that test with a warning. if something else than None is in the list and we are missing private key then we should write a warning to user

opcua/server/server.py
```diff
@@ -99,6 +99,12 @@ def __init__(self, shelffile=None, iserver=None):
         sa_node = self.get_node(ua.NodeId(ua.ObjectIds.Server_ServerArray))
         sa_node.set_value([self._application_uri])
 
+        self.security_endpoints = ["None", "Basic128Rsa15_Sign",
+                                    "Basic128Rsa15_SignAndEncrypt", 
+                                    "Basic256_Sign", "Basic256_SignAndEncrypt"]
+        self.policyIDs = ["Anonymous", "Basic256", "Basic128", "Username"]
```
oroulet: didn't you plan to make this variables private and add two methods? something like set_security_modes and set_user_identification_modes? I think this will improve discoverability and doc

opcua/server/server.py
kleskjr: Something like?
```
if (len(self._security_policy)>1) and self.private_key:
    self.logger.warning("Creating an open endpoint to the server, although encrypted endpoints are enabled.")
```

oroulet:
Looks very good! thank for the PR, that was something missing. Could you also add calls to these new methods into the server-example.py? so people see them.

kleskjr:
i skipped the definition of policyID, as they are currently not really implemented.

opcua/server/server.py
oroulet: and that one we forgot tp fix

oroulet:
I was too quick merging that one. I did not saw that it broke tests! @kleskjr Can you fix them?

oroulet:
@kleskjr looks like I fixed it by changing default enabled policies...

kleskjr:
cool, thanks!

opcua/server/server.py
kleskjr: weird, on my PC these missing lines (with the additional check) are shown, and yesterday I could see them on GitHub, while now they are missing.

oroulet:
there are a bunch of weirds things that happend. One of your commits went
also directly into the git log on master... while the others were
squashed...

kleskjr:
maybe it's my bad that I didn't create a separate branch for this changes, but worked directly on master instead. Now I am using branches for the different features/modifications.

Issue #612, opened 2018-06-05
ModifySubscription missing from server (i=793).


brubbel:
Hi,

It seems that support for the ModifySubscription service is not implemented in subscription_service.py.
According to the Part 7 of the OPC-UA specification, implementation is not optional to conform to the minimal 'Subscription Basic' profile (Part 7, rev 1.04, Table 10, page 13):

> SubscriptionBasic
> Support the following Subscription Services: CreateSubscription, ModifySubscription, DeleteSubscriptions, Publish, Republish and SetPublishingMode.

PS: this is not a request (not even a kind one) for implementation, but I want to report that at least one client (.NET) fails on this.

oroulet:
Can't you try to fix it? At least receiving the call and answering that everything is ok should be easy and make client happy

brubbel:
In this issue (https://github.com/FreeOpcUa/python-opcua/issues/594) you mentioned that some parts are autogenerated, is this also the case for subscription_service.py? Do you have a pointer for that?

oroulet:
I wrote that they could but it is not. Implementing it will be mainly copy paste from others and easy so try ;-)

PR #613, opened 2018-06-05
Subscription Service: ModifySubscriptionRequest

brubbel:
cfr. https://github.com/FreeOpcUa/python-opcua/issues/612
Note: requested parameters are shamelessly ignored and the original subscription settings are sent back to the OPC-UA client.


oroulet:
Lookd correct

PR #614, opened 2018-06-06
SecurityPolicy defined by enum class

kleskjr:
Using enum instead of strings for SecurityPolicy definition

### DIFF
--- a/examples/server-example.py
+++ b/examples/server-example.py
@@ -88,12 +88,8 @@ if __name__ == "__main__":
     #server.set_endpoint("opc.tcp://localhost:4840/freeopcua/server/")
     server.set_endpoint("opc.tcp://0.0.0.0:4840/freeopcua/server/")
     server.set_server_name("FreeOpcUa Example Server")
-    # set possible endpoint policies for clients to connect through
-    server.set_security_policy(["None",
-                                "Basic128Rsa15_Sign",
-                                "Basic128Rsa15_SignAndEncrypt", 
-                                "Basic256_Sign",
-                                "Basic256_SignAndEncrypt"])
+    # set all possible endpoint policies for clients to connect through
+    server.set_security_policy([0, 1, 2, 3, 4])
 
     # setup our own namespace
     uri = "http://examples.freeopcua.github.io"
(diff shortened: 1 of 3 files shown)


examples/server-example.py
```diff
@@ -88,12 +88,8 @@ def run(self):
...
     server.set_endpoint("opc.tcp://0.0.0.0:4840/freeopcua/server/")
     server.set_server_name("FreeOpcUa Example Server")
-    # set possible endpoint policies for clients to connect through
-    server.set_security_policy(["None",
-                                "Basic128Rsa15_Sign",
-                                "Basic128Rsa15_SignAndEncrypt", 
-                                "Basic256_Sign",
-                                "Basic256_SignAndEncrypt"])
+    # set all possible endpoint policies for clients to connect through
+    server.set_security_policy([0, 1, 2, 3, 4])
```
oroulet: that is not correct way to use enums ;-)

    server.set_security_policy([ua.SecurityPolicies.NoSecurity, ua.SecurityPolicies.Basic128Rsa15_Signm etc...])

kleskjr: I see. Just tried to decrease the number of letters written :)
Is the SecurityPolicyType class definition fine?

oroulet:
The declaration looks fine. But maybe you could just call it
SecurityPolicies to save a few keys...

opcua/ua/uatypes.py
```diff
@@ -957,3 +957,24 @@ def get_extensionobject_class_type(typeid):
...
+    The supported types of SecurityPolicy.
+    
+    "None"
+    "Basic128Rsa15_Sign"
+    "Basic128Rsa15_SignAndEncrypt"
+    "Basic256_Sign"
+    "Basic256_SignAndEncrypt"
+
+    :ivar Variant:
+    :ivar DiagnosticInfo:
```
oroulet: these 2 lines do not need to be there

kleskjr:
the reason I went with this name is that there is already `SecurityPolicy` class. 
I find `SecurityPolicyType` bit clearer than `SecurityPolicies` and might avoid confussion. Anyway, it's your call.

oroulet:
I see....I still think SecutiryPolicies is simpler. The other one should be renamed at one point

oroulet:
forget my last comment. do whaterver you want

PR #615, opened 2018-06-06
Dev

oroulet:


oroulet:
@jobasto I added a method to load enums from server. Not sure how this should be done... do you have any comments. In a way this is related to your PR, but in your PR you created the enum manually. Now I get them from server...

jobasto:
@oroulet I was getting the enum information from the xml file that was downloaded from the opc ua server (value of a DataTypeDictionaryType node, Component of Types->DataTypes->OPC Binary). That is the same xml file where also the structures are defined. As I now realize I didn't do a cross check between the xml and the server address space, i.e. if the enums defined in the xml also are defined as subtypes of the Enumeration node (ns=0, i=29). So we've got two places where the same enums can be defined. So what should we do? Ignore the enums in the DataTypeDictionary and use only the Enumeration node subtypes? Generate a union or intersection? I'm not quite sure.

PR #620, opened 2018-06-07
WIP: Client verification

kleskjr:
verifing clients's public certificate and public key before establishing a secure connection


oroulet:
I have no idea if check is correct but it looks simple and valid. For storing certificate I propose that you make a new class called CertificateStoreInMemory or similar with a few methods like add_certicate an check_certificate (cert) and set_checker. Then add a method to InternalServer called set_certicate_store.Thus people can make their own implementation later.  The certificate validation itself should be done in InternalServer or wherever you think is best

oroulet:
Maybe make a CertificateStoreInterface too with all methods raising NotImplemented

Issue #641, opened 2018-07-02
Client connect using username and password throws AttributeError


Salohcin:
When trying to connect to an OPC server with a username and password using the following code an AttributeError is thrown:

```python
client = Client("opc.tcp://127.0.0.1:53530/OPCUA/SimulationServer")
client.set_user("opctest")
client.set_password("test")
client.connect()
```

Traceback:

```pytb
Error
Traceback (most recent call last):
  File "\lib\unittest\case.py", line 58, in testPartExecutor
    yield
  File "\lib\unittest\case.py", line 580, in run
    testMethod()
  File "\code\opc\test\test_OPCModule.py", line 26, in testPlaintext
    self.client.connect()
  File "\lib\site-packages\opcua\client\client.py", line 259, in connect
    self.activate_session(username=self._username, password=self._password, certificate=self.user_certificate)
  File "\lib\site-packages\opcua\client\client.py", line 436, in activate_session
    return self.uaclient.activate_session(params)
  File "\lib\site-packages\opcua\client\ua_client.py", line 259, in activate_session
    data = self._uasocket.send_request(request)
  File "\lib\site-packages\opcua\client\ua_client.py", line 75, in send_request
    future = self._send_request(request, callback, timeout, message_type)
  File "\lib\site-packages\opcua\client\ua_client.py", line 54, in _send_request
    binreq = struct_to_binary(request)
  File "\lib\site-packages\opcua\ua\ua_binary.py", line 258, in struct_to_binary
    packet.append(to_binary(uatype, val))
  File "\lib\site-packages\opcua\ua\ua_binary.py", line 281, in to_binary
    return struct_to_binary(val)
  File "\lib\site-packages\opcua\ua\ua_binary.py", line 258, in struct_to_binary
    packet.append(to_binary(uatype, val))
  File "\lib\site-packages\opcua\ua\ua_binary.py", line 271, in to_binary
    return pack_uatype(vtype, val)
  File "\lib\site-packages\opcua\ua\ua_binary.py", line 185, in pack_uatype
    return extensionobject_to_binary(value)
  File "\lib\site-packages\opcua\ua\ua_binary.py", line 457, in extensionobject_to_binary
    Body = struct_to_binary(obj)
  File "\lib\site-packages\opcua\ua\ua_binary.py", line 258, in struct_to_binary
    packet.append(to_binary(uatype, val))
  File "\lib\site-packages\opcua\ua\ua_binary.py", line 271, in to_binary
[...]

PR #643, opened 2018-07-02
Change default UserIdentityToken.PolicyId value to String

Salohcin:
Fix for #641


oroulet:
Thanks for the debugging. I will have a look

PR #649, opened 2018-07-16
Updates asyncio to work with Python 3.7

crenwick:
Since Python 3.4.4, `asyncio.async()` has been a deprecated alias to `asyncio.ensure_future()` and in Python 3.7 the`asyncio.async()` alias was removed.


oroulet:
Ok. That was coming. The problem is that this is not supported in python 2.7. we may well start removing support for python 2.7...

oroulet:
Shit, in fact it worked with Python 2.7...

PR #656, opened 2018-07-23
User/Password authentication

Granjow:
This MR adds a new feature to the command-line tools to provide username/password by argument instead of by URI; details see respective commit messages.

It also prevents `uals` from crashing in a certain situation and improves the readme a bit.


oroulet:
Thanks

PR #657, opened 2018-07-25
Fix reference and parent parsing

ipa-mdl:
This reverts 0b6c4d8, because it overwrites the parent for all backward references.

> parent giving as ParentNodeId misses the ref type, so better to prioritize the ref descr

if ParentNodeId is given, I would expect a corresponding backward reference.
I have added a warning if this is not the case. In addtion the parent will be deleted.

As an alternative this special case could be handled later by searching for a corresponding forward reference from the parent..

### DIFF
--- a/opcua/common/xmlparser.py
+++ b/opcua/common/xmlparser.py
@@ -331,6 +331,8 @@ class XMLParser(object):
         return body
 
     def _parse_refs(self, el, obj):
+        parent, parentlink = obj.parent, None
+
         for ref in el:
             struct = RefStruct()
             struct.forward = "IsForward" not in ref.attrib or ref.attrib["IsForward"] not in ("false", "False")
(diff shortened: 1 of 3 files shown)


ipa-mdl:
I have now implemented a fallback behavior which tries to detect the proper parent first.

opcua/common/xmlparser.py
```diff
@@ -331,6 +331,8 @@ def _parse_body(self, el):
         return body
 
     def _parse_refs(self, el, obj):
+        parent, parentlink = obj.parent, None
```
oroulet: I trust you that you analysis is correct but that local variable assignment seems a bit broken or simply overcomplicated. you create a parent variable but then still use obj.parent at several places.. I cannot follow it and I suppose it should be possible to simplify this...and generally I am not very fan of creating local variables just to reduce variable length..

ipa-mdl: `parent`  and `parentlink`  are the fallbacks that will be used if no other backward reference matches (-> after the loop).

opcua/common/xmlparser.py
```diff
@@ -341,8 +343,12 @@ def _parse_refs(self, el, obj):
             if ref.attrib["ReferenceType"] == "HasTypeDefinition":
                 obj.typedef = ref.text
             elif not struct.forward:
-                # Rmq: parent giving as ParentNodeId misses the ref type, so better to prioritize the ref descr
-                #if not obj.parent:
-                obj.parent = ref.text
-                if obj.parent == ref.text:
-                    obj.parentlink = ref.attrib["ReferenceType"]
+                parent, parentlink = struct.target, struct.reftype
```
ipa-mdl: this just replicates the current bevahior to take the last reference

opcua/common/xmlparser.py
```diff
@@ -341,8 +343,12 @@ def _parse_refs(self, el, obj):
...
+                if not obj.parent:
```
ipa-mdl: in the next block I use the local variables to avoid the duplicated member and dictionary access..

opcua/common/xmlparser.py
ipa-mdl: > but then still use obj.parent at several places.. 

only to set it.

ipa-mdl: `parent`  gets initialized to `obj.parent` for the case that there are no backward references at all.
This way it will stay the same, if `obj.parentlink`  is `None`.

Issue #659, opened 2018-07-26
RuntimeError: dictionary changed size during iteration


brubbel:
Python3, fairly heavy usage and multithreading. Probably iterator-related. PR will follow.

```
RuntimeError: dictionary changed size during iteration
Exception in callback InternalSubscription._sub_loop()
handle: <TimerHandle when=598477.428340331 InternalSubscription._sub_loop()>
Traceback (most recent call last):
  File "/usr/lib/python3.5/asyncio/events.py", line 126, in _run
    self._callback(*self._args)
  File "/usr/local/lib/python3.5/dist-packages/opcua/server/internal_subscription.py", line 289, in _sub_loop
    self.publish_results()
  File "/usr/local/lib/python3.5/dist-packages/opcua/server/internal_subscription.py", line 315, in publish_results
    result = self._pop_publish_result()
  File "/usr/local/lib/python3.5/dist-packages/opcua/server/internal_subscription.py", line 322, in _pop_publish_result
    self._pop_triggered_datachanges(result)
  File "/usr/local/lib/python3.5/dist-packages/opcua/server/internal_subscription.py", line 338, in _pop_triggered_datachanges
    notif.MonitoredItems = [item for sublist in self._triggered_datachanges.values() for item in sublist]
  File "/usr/local/lib/python3.5/dist-packages/opcua/server/internal_subscription.py", line 338, in <listcomp>
    notif.MonitoredItems = [item for sublist in self._triggered_datachanges.values() for item in sublist]
```

oroulet:
The trace is missing the error...

devel0peris:
With version 98.5 I still get this error. I'm not iterating over a dictionary in my code. Could this still be the same error?
For privacy reasons, I have replaced project-related traces with `[...]`.

```
ERROR:opcua.common.subscription:Exception calling data change handler
Traceback (most recent call last):
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\opcua\common\subscription.py", line 132, in _call_datachange
    self._handler.datachange_notification(data.node, item.Value.Value.Value, event_data)
  [...]
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\opcua\common\node.py", line 578, in delete
    results = opcua.common.manage_nodes.delete_nodes(self.server, [self], recursive, delete_references)
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\opcua\common\manage_nodes.py", line 390, in delete_nodes
[...]

PR #660, opened 2018-07-26
fix: dictionary changed size during iteration

brubbel:
https://github.com/FreeOpcUa/python-opcua/issues/659
https://github.com/FreeOpcUa/python-opcua/issues/661

### DIFF
--- a/opcua/server/internal_subscription.py
+++ b/opcua/server/internal_subscription.py
@@ -335,16 +335,18 @@ class InternalSubscription(object):
     def _pop_triggered_datachanges(self, result):
         if self._triggered_datachanges:
             notif = ua.DataChangeNotification()
-            notif.MonitoredItems = [item for sublist in self._triggered_datachanges.values() for item in sublist]
-            self._triggered_datachanges = {}
+            with self._lock:
+                notif.MonitoredItems = [item for sublist in self._triggered_datachanges.values() for item in sublist]
+                self._triggered_datachanges = {}
             self.logger.debug("sending datachanges notification with %s events", len(notif.MonitoredItems))
             result.NotificationMessage.NotificationData.append(notif)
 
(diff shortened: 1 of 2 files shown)


oroulet:
Ok. This lock thing is dangerous. I will have a look

brubbel:
True, looking with a fresh mind, it is not quite right.

brubbel:
[Here](https://github.com/FreeOpcUa/python-opcua/blob/79e09e8082e383739fd7e36c9c1cf5467d470acd/opcua/server/internal_subscription.py#L205), [here](https://github.com/FreeOpcUa/python-opcua/blob/79e09e8082e383739fd7e36c9c1cf5467d470acd/opcua/server/internal_subscription.py#L243) and [here](https://github.com/FreeOpcUa/python-opcua/blob/79e09e8082e383739fd7e36c9c1cf5467d470acd/opcua/server/internal_subscription.py#L246) are calls from MonitoredItemService directly into InternalSubscription, thereby bypassing the RLock() of the InternalSubscription object. There is a lock in place, but it is the RLock() of the MonitoredItemService object.
So, a `with self._lock()` in [_enqueue_event](https://github.com/FreeOpcUa/python-opcua/blob/79e09e8082e383739fd7e36c9c1cf5467d470acd/opcua/server/internal_subscription.py#L385) should be sufficient.

oroulet:
But each lock is probably only to protect one sensitive structure. Do you have overview over which one?

brubbel:
I'm not using opcua events for the moment, so I guess the [enqueue_datachange_event](https://github.com/FreeOpcUa/python-opcua/blob/79e09e8082e383739fd7e36c9c1cf5467d470acd/opcua/server/internal_subscription.py#L205) is the cause.

Issue #664, opened 2018-07-27
KeepAlive: concurrent.futures._base.TimeoutError (client.py)


brubbel:
Background:
1. FreeOpcUa Client connects to server and is silent from now on.
2. Server disconnect.
3. After 0.7*1hr, the error below is printed from the KeepAlive thread.

PR for a graceful KeepAlive exit underway.

```
Exception in thread Thread-5:
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/***/venv3.6/lib/python3.6/site-packages/opcua/client/client.py", line 61, in run
    self.client.open_secure_channel(renew=True)
  File "/home/***/venv3.6/lib/python3.6/site-packages/opcua/client/client.py", line 302, in open_secure_channel
    result = self.uaclient.open_secure_channel(params)
  File "/home/***/venv3.6/lib/python3.6/site-packages/opcua/client/ua_client.py", line 235, in open_secure_channel
    return self._uasocket.open_secure_channel(params)
  File "/home/***/venv3.6/lib/python3.6/site-packages/opcua/client/ua_client.py", line 176, in open_secure_channel
    response = struct_from_binary(ua.OpenSecureChannelResponse, future.result(self.timeout))
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 434, in result
    raise TimeoutError()
concurrent.futures._base.TimeoutError
```

PR #665, opened 2018-07-27
fixes TimeoutError in KeepAlive thread

brubbel:
https://github.com/FreeOpcUa/python-opcua/issues/664

I took the liberty to throw a more relevant ConnectionError instead of TimeoutError in the open_secure_channel() method. Another option is an ua.UaError derived type?

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -1,5 +1,6 @@
 from __future__ import division  # support for python2
 from threading import Thread, Condition
+import concurrent.futures
 import logging
 try:
     from urllib.parse import urlparse
@@ -60,8 +61,8 @@ class KeepAlive(Thread):
             self.logger.debug("renewing channel")
             try:
                 self.client.open_secure_channel(renew=True)
-            except ConnectionError as e:
-                self.logger.debug('keepalive failed: {:s}'.format(str(e)))
+            except concurrent.futures.TimeoutError:
+                self.logger.debug("keepalive failed: timeout on open_secure_channel()")
                 break
             val = server_state.get_value()
             self.logger.debug("server state is: %s ", val)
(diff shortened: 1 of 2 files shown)


opcua/client/ua_client.py
```diff
@@ -173,7 +173,10 @@ def open_secure_channel(self, params):
 
         # FIXME: we have a race condition here
         # we can get a packet with the new token id before we reach to store it..
-        response = struct_from_binary(ua.OpenSecureChannelResponse, future.result(self.timeout))
+        try:
+            response = struct_from_binary(ua.OpenSecureChannelResponse, future.result(self.timeout))
+        except TimeoutError:
+            raise ConnectionError('OpenSecureChannelRequest() timeout')
```
oroulet: there is no try/except anywhere in that class, there is no reason to add one here. let users catch this (this is best practice)

opcua/client/client.py
```diff
@@ -58,7 +58,11 @@ def run(self):
             if self._dostop:
                 break
             self.logger.debug("renewing channel")
-            self.client.open_secure_channel(renew=True)
+            try:
+                self.client.open_secure_channel(renew=True)
+            except ConnectionError as e:
+                self.logger.debug('keepalive failed: {:s}'.format(str(e)))
+                break
```
oroulet: I am not really sure what it helps. What do you want? simply a cleaner output in case of crash? or do you want to detect connection error sooner?

brubbel: There is no negative impact. The only task of the KeepAlive thread is to keep the connection alive, but I was '''annoyed''' by the backtrace in the logs every once in a while.
I'll remove the ConnectionError and handle the timeout in the KeepAlive then.

brubbel:
As an aside, are there any plans in the direction of an asyncio-client? Also, what is the future of Py2.7 support (with the native `async def` and `await` syntax in mind).

oroulet:
opcua-asyncio will not support python < 3.6

PR #678, opened 2018-08-14
Fixes and perf improvments

oroulet:


jobasto:
@oroulet:

Yes, we should avoid eval if we can. The reason for the eval in the line

return "ua." + uatype + "(" + list(eval("ua."+uatype))[0] + ")"

was that there may be enums that don’t have a “0” element so we can’t always initialize it with

"ua." + uatype + "(0)".

If you apply list to an enum you get a list of the enumeration values and I was taking the first one to make sure that I can really instantiate an enum like that.



Geschäftsführer: Dr. Oliver Vietze · Severino Bruno
Sitz der Gesellschaft: Stockach

Handelsregister Freiburg im Br. HRB 71317
Ust. ID: DE 301255099

jobasto:
@oroulet: Yes, we should avoid eval if we can. The reason for the eval in the line
`return "ua." + uatype + "(" + list(eval("ua."+uatype))[0] + ")"`
was that there may be enums that don’t have a “0” element so we can’t always initialize it with
`"ua." + uatype + "(0)"`.
If you apply list to an enum you get a list of the enumeration values and I was taking the first one to make sure that I can really instantiate an enum like that.

oroulet:
@jobasto I understand why you are using enum but:
* I would really like to find a solution without eval
* In practice do we really have enums not starting with 0?

nevermind I just read in the doc that we should use the __member__ attrbute

jobasto:
@oroulet Cool, I didn't know that. Do you have a link for that doc?

oroulet:
https://docs.python.org/3/library/enum.html 8.13.7

jobasto:
@oroulet Thanks

PR #679, opened 2018-08-14
Fixed two bugs

yangpeiren:
## Bug 1 
in opcua/common/structures.py line 42 method `get_default_value`:

```python
elif  issubclass(eval("ua."+uatype), IntEnum):
```

this will case an exception if an uatype is not registered yet, another condition is added to make sure that the `uatype` is already in module `ua` by checking the dictionary of `ua`, the changed code:

```python
elif uatype in ua.__dict__ and issubclass(eval("ua."+uatype), IntEnum):
```

## Bug 2 
in opcua/server/address_space.py line 524 method `generate_nodeid` of class `AddressSpace`:

```python
def generate_nodeid(self, idx=None):
        if idx is None:
            idx = self._default_idx
        if idx in self._nodeid_counter:
            self._nodeid_counter[idx] += 1
        else:
            self._nodeid_counter[idx] = 1
        nodeid = ua.NodeId(self._nodeid_counter[idx], idx)
        with self._lock:  # OK since reentrant lock
            while True:
                if nodeid in self._nodes:
                    nodeid = self.generate_nodeid(idx)
                else:
                    return nodeid
```

Adding objects in one specific name space with a user defined global counter to generate node id like following:

```python
self.server.nodes.objects.add_object(ua.NodeId(global_counter, idx), node_name)
```

if this is executed more than 1,000 times, afterwards we want to add a method, the node id of the input and output arguments are lazy generated (see opcua/common/manage_nodes.py line 330 and 337), which means it will call the `generate_nodeid` method in class `AddressSpace`. In this case the self._nodeid_counter[idx] will be initialized (before this, the user uses an external global counter),  here it  checks the occupation of the ids from 1 on, if fails it will make one time recursion:

```python
while True:
                if nodeid in self._nodes:
                    nodeid = self.generate_nodeid(idx)
                else:
                    return nodeid
```

The maximum recursion allowed in python is 1000, this case will cause the maximum recursion depth exceeded exception. I have added code to find the existed maximum id, use this number plus one as the initial value for the counter to make the lazy generation working.

```python
      def generate_nodeid(self, idx=None):
        if idx is None:
            idx = self._default_idx
        if idx in self._nodeid_counter:
            self._nodeid_counter[idx] += 1
        else:
            # here get the biggest number from the address space, and add one
            identifier_list = sorted([i.Identifier for i in self._nodes.keys() if i.NamespaceIndex == idx])
            if identifier_list:
                self._nodeid_counter[idx] = identifier_list[-1]
            else:
                self._nodeid_counter[idx] = 1
        nodeid = ua.NodeId(self._nodeid_counter[idx], idx)
        with self._lock:  # OK since reentrant lock
            while True:
                if nodeid in self._nodes:
                    nodeid = self.generate_nodeid(idx)
                else:
                    return nodeid
```


oroulet:
@yangpeiren tests are failing your new nodeid method... not sure why...

yangpeiren:
Sorry I didn't consider the case that a node id could be a StringNodeId, now I fixed it by changing:

```python
identifier_list = sorted([i.Identifier for i in self._nodes.keys()  if i.NamespaceIndex == idx])
```
to

```python
identifier_list = sorted([i.Identifier for i in self._nodes.keys() if i.NamespaceIndex == idx and str(i.Identifier).isdigit()])
```

and ran the tests, this time it passes the tests. Please check it.

yangpeiren:
Yes, correct, and I have updated the readability of the code with your propose, please check it.

yangpeiren:
@oroulet Thank you for your quick reaction, and I have tested it with my use case, it works fine.

PR #681, opened 2018-08-20
New classes to support Creating extension objects by users

yangpeiren:
## Motivation

In my [use case](https://github.com/iirob/ros_opcua_communication/tree/interactive_ros/ros_opcua_impl_python_opcua) I need to read some external structures and create extension objects from them on-the-fly, I wrote two classes to support this as an extension of the standard python-opcua package. 

But I've seen that there are some other discussions to request this function in the python-opcua package, and in #388 there are two TODOs open:

server:
- [ ] add a method that generates the xml (to be used in code and by opcua-modeler)
- [ ] add a method that creates the necessary data types nodes

 which could correspondent to my classes `OPCTypeDictionaryBuilder` and  `DataTypeDictionaryBuilder`, therefore I created this pull request. 

Please first **Do Not Merge**, since I am not sure where you would like to put them in the project, I would suggest to put them in `opcua/common/structures.py`.

### Class `OPCTypeDictionaryBuilder`:
Generates xml files from a structure definition. The result of the xml string generated in the example `examples/server_create_custom_structures.py` is like this:
```xml
<opc:TypeDictionary DefaultByteOrder="LittleEndian" TargetNamespace="http://examples.freeopcua.github.io" xmlns:opc="http://opcfoundation.org/BinarySchema/" xmlns:tns="http://examples.freeopcua.github.io" xmlns:ua="http://opcfoundation.org/UA/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <opc:Import Namespace="http://opcfoundation.org/UA/" />
  <opc:StructuredType BaseType="ua:ExtensionObject" Name="BasicStructure">
    <opc:Field Name="ID" TypeName="opc:Int32" />
    <opc:Field Name="Gender" TypeName="opc:Boolean" />
    <opc:Field Name="Comments" TypeName="opc:String" />
  </opc:StructuredType>
  <opc:StructuredType BaseType="ua:ExtensionObject" Name="NestedStructure">
    <opc:Field Name="Name" TypeName="opc:String" />
    <opc:Field Name="Surname" TypeName="opc:String" />
    <opc:Field Name="Stuff" TypeName="tns:BasicStructure" />
  </opc:StructuredType>
</opc:TypeDictionary>
```

The `append_struct` method creates a new structure frame, and the `add_field` method allows to add structure members it the structure frame. The `get_dict_value` method retrieves the created xml string.

### One possible question

Maybe there is a question, why is there another parameter `struct_name` in `add_field` function, since we know after the structure creation we will directly add fields to the structure just then added?

The answer is to make the life easier for the users who use recursion to traverse some files or strings and create ua structures, for instance in my use case. But another method could be provided without the parameter `struct_name` in `add_field` function so that the users directly add structure members to it after the structure frame is created.

### Class `DataTypeDictionaryBuilder`

This class creates an `OPC Binary Dictionary` of the newly created extension objects. The result in the example `examples/server_create_custom_structures.py` is like this in the tool `UaExpert`:
built binary dictionary

This class has an internal `OPCTypeDictionaryBuilder` object, so as to automate the xml creation. 

The `create_data_type` method is to create new structure frame, and the `add_field` is to add structure members to the new structure frame. 

The `set_dict_byte_string` method will set the xml string created by the internal `OPCTypeDictionaryBuilder` object to the created dictionary, in the example case, it is the `MyDictionary` shown in the picture above. By this way, the standard UA tools like `UaExpert` can also correctly parsing the newly created objects. Below are two pics to show the result of the example.

The user could see the value of the extension objects in the attribute window directly:
value in attributes

Or by double-clicking the value in the data access view:
value in data access view

### Another possible question

Maybe there is another question, why not use the standard functions `Node.add_data_type`, `Node.add_variable` and `add_object` instead of directly creating nodes with some low level functions?

Since the dictionary uses quite different node types and reference types, for instance:

The dictionary description node is a ua variable, it has this type definition:

```python
desc_node.TypeDefinition = ua.NodeId(ua.ObjectIds.DataTypeDescriptionType, 0)
```

And for the object node, it has even different reference types:

```python
obj_node.ReferenceTypeId = ua.NodeId(ua.ObjectIds.HasEncoding, 0)
obj_node.TypeDefinition = ua.NodeId(ua.ObjectIds.DataTypeEncodingType, 0)
```

And the references inside the data types and dictionary are complex to program with the standard methods.

## The naming problem in `opcua/common/structures.py`

1. In `opcua/common/structures.py` the original method `_clean_name` (line 283) removes the illegal chars in the class name, but the user does not know the name changes afterward. Therefore the user could not locate the created class by the given name.

2. The method `_clean_name` only removes the illegal chars but does not make it to the python conventional camel case.

I would suggest using the method `_to_camel_case` in the `opcua/common/structure_extension.py`, it removes the illegal chars and makes the name string in camel case without damaging the original camel case in the string.

For example: `turtle_actionlib/ShapeActionFeedback -> TurtleActionlibShapeActionFeedback`, the original camel case in the part `ShapeActionFeedback` is kept with this method.

I also used `_to_camel_case` in designing `OPCTypeDictionaryBuilder` and  `DataTypeDictionaryBuilder`, so that the created extension objects could be directly used for generating python classes. In the UA nodes, the browser name will be modified by the `_to_camel_case` function to camel case, while the display name is set to the user given name, which makes the nodes identifiable for the users.

To solve problem 1, the `get_ua_class` method could be used (see example), each time if the user wants to use one specific class,  the method with the user given name will yield the generated class.

```python
my_ext_class = get_ua_class('turtle_actionlib/ShapeActionFeedback')
my_ext_obj = my_ext_class()
```

The `/structure_extension.py` is tested with the example, and with my use case, which includes more than 1,000 newly created structures, some structures are nested in several layers, some structures have structures lists.

### DIFF
--- a/examples/server_create_custom_structures.py
+++ b/examples/server_create_custom_structures.py
@@ -1,0 +1,96 @@
+from opcua import ua, Server
+from opcua.common.structure_extension import DataTypeDictionaryBuilder, get_ua_class
+from IPython import embed
+
+
+class DemoServer:
+
+    def __init__(self):
+        self.server = Server()
+
+        self.server.set_endpoint('opc.tcp://0.0.0.0:51210/UA/SampleServer')
+        self.server.set_server_name('Custom structure demo server')
+        # idx name will be used later for creating the xml used in data type dictionary
+        self._idx_name = 'http://examples.freeopcua.github.io'
+        self.idx = self.server.register_namespace(self._idx_name)
+
+        self.dict_builder = DataTypeDictionaryBuilder(self.server, self.idx, self._idx_name, 'MyDictionary')
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        quit()
+
+    def start_server(self):
+        self.server.start()
+
+    def create_structure(self, name):
+        # save the created data type
+        data_type_id = self.dict_builder.create_data_type(name)
+        return data_type_id
+
+    def add_field(self, type_name, field_name, struct_name, is_array=False):
+        self.dict_builder.add_field(type_name, field_name, struct_name, is_array)
+
+    def complete_creation(self):
+        self.dict_builder.set_dict_byte_string()
+
+
+if __name__ == '__main__':
+    with DemoServer() as ua_server:
+        # add one basic structure
+        basic_struct_name = 'basic_structure'
+        basic_data_type = ua_server.create_structure(basic_struct_name)
+        ua_server.add_field('Int32', 'ID', basic_struct_name)
+        ua_server.add_field('Boolean', 'Gender', basic_struct_name)
+        ua_server.add_field('String', 'Comments', basic_struct_name)
+
+        # add an advance structure which uses our basic structure
+        nested_struct_name = 'nested_structure'
+        nested_data_type = ua_server.create_structure(nested_struct_name)
+        ua_server.add_field('String', 'Name', nested_struct_name)
+        ua_server.add_field('String', 'Surname', nested_struct_name)
+        ua_server.add_field(basic_struct_name, 'Stuff', nested_struct_name)
+
+        # this operation will write the OPC dict string to our new data type dictionary
+        # namely the 'MyDictionary'
+        ua_server.complete_creation()
+
+        # get the working classes
+        ua_server.server.load_type_definitions()
+
+        # Create one test structure
+        basic_var = ua_server.server.nodes.objects.add_variable(ua.NodeId(namespaceidx=ua_server.idx), 'BasicStruct',
+                                                                ua.Variant(None, ua.VariantType.Null),
+                                                                datatype=basic_data_type)
+
+        basic_var.set_writable()
+        basic_msg = get_ua_class(basic_struct_name)()
+        basic_msg.ID = 3
+        basic_msg.Gender = True
+        basic_msg.Comments = 'Test string'
+        basic_var.set_value(basic_msg)
+
+        # Create one advance test structure
+        nested_var = ua_server.server.nodes.objects.add_variable(ua.NodeId(namespaceidx=ua_server.idx), 'NestedStruct',
+                                                                 ua.Variant(None, ua.VariantType.Null),
+                                                                 datatype=nested_data_type)
+
+        nested_var.set_writable()
+        nested_msg = get_ua_class(nested_struct_name)()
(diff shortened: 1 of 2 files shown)


oroulet:
Hei, 

Thanks for the PR. I recently wrote code in opcua-modeler to create custom structures. But your PR looks better and having this in python-opcua is better, maybe I should use your code to simplify code in opcua.modeler? Could you have a look at https://github.com/FreeOpcUa/opcua-modeler/pull/65 ?

Then I am wondering. Why not returning an object from create_data_type and call add_field from that object? instead of having to provide structure to add_name? it look very C-like...

yangpeiren:
Hey, I've seen your code in FreeOpcUa/opcua-modeler#65, in file `opcua-modeler/uamodeler/model_manager.py` line 385 `_create_typedictonary_children`, there is a FIXME requests for the links, actually this struck me also for a long time, as you can see in my code there are 8 different links to be set up. 

I think this two classes could be used to simplify opcua-modeler, the OPCTypeDictionaryBuilder is similar to the method _save_bsd in model_manager.py line 351, the class DataTypeDictionaryBuilder is similar to line 385 _create_typedictonary_children.

Yes your suggestion is very pertinent, I'll change it in a more OO style, besides, I will change the type_name from string like 'Int32' to the ua variables like ua.VariantType.Int32.

yangpeiren:
@oroulet I've changed the `add_field` method, now the create_data_type returns an object with `add_field` method, which could be used to add specific fields to the structure frame.

oroulet:
Great! Also it would be great if you could add some tests for it. I always find bugs when writing tests

yangpeiren:
Good idea, I'll try to write some.

yangpeiren:
@oroulet I wrote 25 test cases with 100% coverage to the methods and classes in structure_extension.py, P.S. the test is not included in the test.py, I keep both the source structure_extension.py and the test cases isolated, so that you can decide where to place them in the project.

oroulet:
Very good work! Concerning names, structure.py is already large so I propose to rename your file type_dictionary_buider.py and integrate to test.py. then I can merge

yangpeiren:
@oroulet I have renamed the module and added the tests to test.py, my local running of run-tests.sh shows no failure. please check it.

yangpeiren:
moment, there seem to be some compatibility problems with my tests between py2 and py3

yangpeiren:
OK now its fixed

PR #683, opened 2018-08-27
implement uacall --method-name

ipa-mdl:


oroulet:
What is he goal of that PR? Wan'tthere a way to provide name before?

ipa-mdl:
>  Wan'tthere a way to provide name before?

Surprisingly not.
Methods could only be specified by the node ID (namespace 0 only) or the index in the array of methods (as returned by `get_methods()`).

Issue #680, opened 2018-08-16
Help: using Siemens WinCC to access python server


ElisabethMZ:
I am having difficulty accessing an opcua server created in python using WinCC.
Has anybody had success getting a connection to work?

Siemens support insists the problem must be caused by improper handling of the certificate in python-opcua

Thank you for your support!

oroulet:
we need more info, what problem do you have? what error?

ElisabethMZ:
FYI: Other OPC UA clients can access the Python OPC UA server and the WinCC client can access other servers. Only the combination Python OPC UA Server and WinCC Client does not work.

I tried three different ways:
1. create the Python server without "load_certificate" and without "load_private_key
Security settings in WinCC Client: None / None

Result: Variables cannot be browsed, no connection to server
Error message in WinCC: "No certificate used by the server to establish secure access".

2. create the Python server with "load_certificate" and with "load_private_key
Security settings in WinCC Client: None / None

Result: Variables cannot be browsed, no connection to the server
Error message in WinCC: "The resolution of the IP address failed. Use the computer name instead." (When using the computer name instead of the IP address, the same error message appears)

3. create the Python server with "load_certificate" and with "load_private_key
Security settings in WinCC Client: Basic128Rsa15 / Sign and encrypt

Result: Variables can be browsed, no communication with the server during runtime, no error message

zerox1212:
You will probably have to use wire shark to see what WinCC is really doing.

oroulet:
Yes you will need to record a session with Wireshark in case 2 so we understand what they want. But no certificate is necessary when selecting None/None. This is a most probably a bug on their side

zerox1212:
I think someone on here once posted that WinCC requires encryption by default.

ElisabethMZ:
I am currently in contact with the Siemens Support and will keep you updated.
Thank you for your help :)

ElisabethMZ:
The following statement came from Siemens Support:
“My level of knowledge is that even with a None-None configuration, a certificate with "no encryption" is transmitted and therefore has less security requirements.
There are guidelines from the Foundation pages, which we are quite sure to follow. We are very sensitive about this, which means that if there is something that is not where it should be, then we break off the connection.
So I don't think WinCC is the cause.”

oroulet:
[...]

PR #687, opened 2018-08-30
discovery service: dynamically replace 0.0.0.0 endpointUrl in server response

brubbel:
cfr. https://github.com/FreeOpcUa/python-opcua/issues/680

If the opcua server is configured with the default endpoint 'opc.tcp://0.0.0.0:4840',
the discovery service will dynamically replace the endpointUrl with the url provided in the FindServersParameters request of the client. e.g. 'opc.tcp://192.168.10.123:4840'
If the opcua server is configured with a valid endpoint url, configuration remains untouched.

Solves the following issues:
1. Fixes the "Replaced Hostname" warning in UaExpert, even when the opcua server is listening on multiple interfaces.
image
2. Makes freeopcua compatible with OPC Scout (and possibly also with WinCC) if a certificate is used.
3. Provides a simple solution for a valid endpointUrl for NAT'ed systems (vpn etc), without a global discovery server.

### DIFF
--- a/opcua/server/internal_server.py
+++ b/opcua/server/internal_server.py
@@ -4,7 +4,7 @@ Can be used on server side or to implement binary/https opc-ua servers
 """
 
 from datetime import datetime, timedelta
-import pickle
+from copy import deepcopy
 import os
 import logging
 from threading import Lock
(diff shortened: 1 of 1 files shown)


brubbel:
Will need second pair of eyes for this :-)
`import ipaddress` only works for Py3. Is it acceptable to add a dependency to py2-ipaddress (pip install ipaddress) for Py2?

opcua/server/internal_server.py
```diff
@@ -168,19 +171,51 @@ def add_endpoint(self, endpoint):
...
-            for edp in self.endpoints:
-                edp1 = copy(edp)
-                url = urlparse(edp1.EndpointUrl)
-                url = url._replace(netloc=sockname[0] + ":" + str(sockname[1]))
-                edp1.EndpointUrl = url.geturl()
-                edps.append(edp1)
-            return edps
-        return self.endpoints[:]
+        # return to client the endpoints it has access to
+        edps = pickle.loads(pickle.dumps(self.endpoints))
```
oroulet: what is that? why dumping then loading again?

oroulet:
The first version was simple, but the last commit really complicate things... I am afraid to add so much code jsut to replace 0.0.0.0 for clients when this can simply be done by configuring the server

opcua/server/internal_server.py
brubbel: shallow copy() does not copy the endpoints, so either deepcopy() or pickle...

oroulet:
yes using deepcopy is clearly the correct way to implement this ;-)

PR #691, opened 2018-09-05
Add support for server username/password auth

igfarm:
This change adds support for username/password authentication to the server via new server.set_user_manager(user_manager) function.  The user_manager function takes two arguments, namely a username and password, and returns a boolean indicating if access should be granted.

### DIFF
--- a/opcua/server/internal_server.py
+++ b/opcua/server/internal_server.py
@@ -4,6 +4,7 @@ Can be used on server side or to implement binary/https opc-ua servers
 """
 from datetime import datetime, timedelta
 from copy import copy
+from struct import unpack_from, unpack
 import os
 import logging
 from threading import Lock
@@ -29,6 +30,13 @@ from opcua.server.subscription_service import SubscriptionService
 from opcua.server.standard_address_space import standard_address_space
 from opcua.server.users import User
 #from opcua.common import xmlimporter
+
+use_crypto = True
+try:
+    from opcua.crypto import uacrypto
+except ImportError:
+    logging.getLogger(__name__).warning("cryptography is not installed, use of crypto disabled")
+    use_crypto = False
 
 
 class SessionState(Enum):
@@ -56,6 +64,9 @@ class InternalServer(object):
         self.disabled_clock = False  # for debugging we may want to disable clock that writes too much in log
         self._known_servers = {}  # used if we are a discovery server
 
+        self.certificate = None
+        self.private_key = None
+
         self.aspace = AddressSpace()
         self.attribute_service = AttributeService(self.aspace)
         self.view_service = ViewService(self.aspace)
@@ -70,7 +81,7 @@ class InternalServer(object):
 
         self.history_manager = HistoryManager(self)
 
-        self.user_token_manager = None
+        self.user_manager = self.default_user_manager
 
         # create a session to use on server side
         self.isession = InternalSession(self, self.aspace, self.subscription_service, "Internal", user=User.Admin)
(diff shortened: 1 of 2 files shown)


opcua/server/server.py
```diff
@@ -641,3 +644,42 @@ def set_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value):
         so it is a little faster
         """
         return self.iserver.set_attribute_value(nodeid, datavalue, attr)
+    
+    def set_user_manager(self, user_manager):
```
oroulet: I tend to think these methods should be part of InternalServer. They really implement new feature. The Server class is supposed to only be a high level helper class

opcua/server/server.py
```diff
@@ -641,3 +644,42 @@ def set_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value):
...
+        """
+        set up a function which that will check for authorize users. Input function takes username
+        and password as paramters and returns True of user is allowed access, False otherwise.
+        """
+        self.user_manager = user_manager
+        self.iserver.user_token_manager = self.user_token_manager
+
+    def user_token_manager(self, token):
```
oroulet: if this method does something (it seems it does ;-)) then I think its name should contain a verb explain what it does: check_user_token(token) for example. But feel free to propose something better

opcua/server/internal_server.py
```diff
@@ -332,7 +333,10 @@ def activate_session(self, params):
         self.state = SessionState.Activated
         id_token = params.UserIdentityToken
         if isinstance(id_token, ua.UserNameIdentityToken):
-            if self.iserver.allow_remote_admin and id_token.UserName in ("admin", "Admin"):
+            if self.iserver.user_token_manager:
+                if self.iserver.user_token_manager(id_token) == False:
+                    raise utils.ServiceError(ua.StatusCodes.BadUserAccessDenied)
+            elif self.iserver.allow_remote_admin and id_token.UserName in ("admin", "Admin"):
```
oroulet: should'nt we implement this default "admin" user using a default user_manager? instead of having a special case?

oroulet:
thanks for the PR. This seems like a very simple solution the user/password issue, but please have a look at the comments.

opcua/server/internal_server.py
igfarm: That's a good idea.  I will implement that way.

opcua/server/server.py
igfarm: The reason the user_token_manager function (or check_user_token) used in InternalServer is in Server is that it needs access to the Server.private_key to decode the token password. Any suggestion on how to gain access to this variable from InternalServer?

oroulet: It is probably wrong to store private_key in server. Can you move it to internal server?

igfarm:
I moved the functions and variables  to the InternalServer as suggested. It looks cleaner now. The user_manager function is set up by calling InternalServer.set_user_manager().  The function itself requires an the extra isession parameter which is used to set up the  User.Admin type.  The default function is on InternalServer.default_user_manager()

opcua/server/internal_server.py
```diff
@@ -269,7 +281,49 @@ def set_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value):
...
+        """
+        self.user_manager = user_manager
 
+    def reset_user_manager(self):
+        """
+        reset user manager to default state 
+        """
+        self.user_manager = self.default_user_manager
+        
+    def default_user_manager(self, isession, username, password):
```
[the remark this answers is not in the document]
igfarm: @oroulet, where would you like the function to live?

opcua/server/internal_server.py
```diff
@@ -269,7 +281,46 @@ def set_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value):
...
+            except Exception as exp:
+                self.logger.warning("Unable to decrypt password")
+                return False
+
+        # call user_manager if one is defined
+        if self.user_manager is not None:
+            return self.user_manager(self, isession, userName, passwd)
+
+        # If nothing function defined, do the default thing
+        if self.allow_remote_admin and userName in ("admin", "Admin"):
```
oroulet: Why did you remove the default user manager function?I thought it was a good idea. It shows people how they can make they own

igfarm: I see. I can add it back in.... where should this function live?

igfarm:
I added the function back as an example. Thanks for your patience and guidance.

igfarm:
Anything I can do to resolve the conflicts?

oroulet:
@igfarm git pull --rebase origin master
you will get some conflicts to solve that should not be difficult
If you do have experience with conflict solving, remember to make a backup branch before so you do not lose any work

oroulet:
something is wrong there is still a conflict so I just merged ans sqaushed everythig by hand

oroulet:
thanks for the PR!

PR #695, opened 2018-09-11
Make recursive node deletion truly recursive

### LINKED ISSUE #694: Is `manage_nodes.delete_nodes` really recursive?
Hi,

I’ve just looked at the delete_nodes function as it’s also executed by `Server.delete_nodes`. It offers a `recursive` parameter, which i would assume leads to walking all hierarchical children of the nodes specified, and deletes these too. However, that does not seem to be the case:

```
def delete_nodes(server, nodes, recursive=False, delete_target_references=True):                                                                                                                               
    nodestodelete = []                                                                                                                                                                                         
    if recursive:                                                                                                                                                                                              
        nodes += _add_childs(nodes)                                                                                                                                                                            
[…]
def _add_childs(nodes):
    results = []
    for mynode in nodes[:]:
        results += mynode.get_children()
    return results
```

To me, this looks like it’s not recursive, but only deletes the specified nodes as well as their direct children. Childrens children wouldn’t be deleted … right? AFAIK that’s not what the library-user expects when seeing a `recursive=…` parameter.

Thanks nonetheless for this great library ^_^.
Best & Thanks,
-Dario

Nebukadneza:
Fixes #694 

Before it only deleted the first level of nodes below the specified
list, which was not what the parameter `recursive` conveyed to the user.

This also adds a test, which failed in current master, but succeeds in this branch.

### DIFF
--- a/opcua/common/manage_nodes.py
+++ b/opcua/common/manage_nodes.py
@@ -379,7 +379,7 @@ def delete_nodes(server, nodes, recursive=False, delete_target_references=True):
     """
     nodestodelete = []
     if recursive:
-        nodes += _add_childs(nodes)
+        nodes = _add_childs(nodes)
     for mynode in nodes:
         it = ua.DeleteNodesItem()
         it.NodeId = mynode.nodeid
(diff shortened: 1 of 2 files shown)


zerox1212:
I think there was a problem with here if the user's address space had circular references, but I agree that the parameter should do what it says. Have to see what @oroulet says.

PR #697, opened 2018-09-13
Removed security policies with Sha-1 algorithm

ephraim:
Basic256 and Basic128Rsa15 use both the sha-1 algorithm.
This is considererd not secure anymore since OPC UA Spec 1.04.
See:
http://opcfoundation.org/UA-Profile/UA/SecurityPolicy%23Basic128Rsa15
http://opcfoundation.org/UA-Profile/UA/SecurityPolicy%23Basic256

This patch removes both of them and adds the security policy
Basic256Sha256.
See:
http://opcfoundation.org/UA-Profile/UA/SecurityPolicy%23Basic256Sha256

### DIFF
--- a/tests/tests_crypto_connect.py
+++ b/tests/tests_crypto_connect.py
@@ -25,28 +25,28 @@ class TestCryptoConnect(unittest.TestCase):
 
     '''
     @classmethod
-    def setUpClass(cls):
+    def setUpClass(self):
         # start our own server
-        cls.srv_crypto = Server()
-        cls.uri_crypto = 'opc.tcp://127.0.0.1:{0:d}'.format(port_num1)
-        cls.srv_crypto.set_endpoint(cls.uri_crypto)
+        self.srv_crypto = Server()
+        self.uri_crypto = 'opc.tcp://127.0.0.1:{0:d}'.format(port_num1)
+        self.srv_crypto.set_endpoint(self.uri_crypto)
         # load server certificate and private key. This enables endpoints
         # with signing and encryption.
-        cls.srv_crypto.load_certificate("examples/certificate-example.der")
-        cls.srv_crypto.load_private_key("examples/private-key-example.pem")
-        cls.srv_crypto.start()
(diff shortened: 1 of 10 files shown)


oroulet:
thanks for the PR. changing the defaults and not advertising for sha1 stuff seems like a good idea. But what do we do if someone wants to connect to a sha1 only server?

ephraim:
So you want to keep the old algorithms? Although they are obsolete and unsecure? 
Python-opcua could throw a warning about the usage and keep on working.... 
What do you think?

oroulet:
I am not sure what is best. But yes maybe keeping the code, but removing them from examples and defaults and writing a big logger. warning?

ephraim:
Ok. Will change my patch accordingly on monday.

ephraim:
Here the new commit which does not remove the old policies. But added a deprecated warning and removed the usage of the old policies from the server example.

gitforce7:
When is this going to be released? It is the default policy to connect to an encrypted opc ua server, isn't it?

tests/tests_crypto_connect.py
```diff
@@ -25,28 +25,28 @@ class TestCryptoConnect(unittest.TestCase):
 
     '''
     @classmethod
-    def setUpClass(cls):
+    def setUpClass(self):
         # start our own server
-        cls.srv_crypto = Server()
```
oroulet: calling the class cls is convention in classmethod. not an error. This is should have been left as cls

oroulet:
@gitforce7 this PR is an improvment for encryption. the server works fine without

tests/tests_crypto_connect.py
ephraim: Sorry about that! Good to know.

ephraim: Should I change my patch, or generate a new pull request to fix this?

oroulet: if you could generate a new PR it would be great. Thanks for your work! cryptography is definitely a part of that library that need attention

PR #732, opened 2018-11-19
removed unused code: get_client_to_discovery

brubbel:
Removed 3-yo orphaned code.

### DIFF
--- a/opcua/server/server.py
+++ b/opcua/server/server.py
@@ -199,14 +199,6 @@ class Server(object):
             client.register_server(self)
             self.iserver.loop.call_later(self._discovery_period, self._renew_registration)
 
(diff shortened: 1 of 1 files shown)


oroulet:
I do not remember this well. but I though there was a reason for that one...

oroulet:
OK now I remember. The spec says that servers should register to discovery servers. and to register to discovery server they need a client, so this method is just a helper method to simplify registration.
but maybe either it should be documented or it should do more and be rename like for example: register_to_discovery_server(), although I thought there was already such a method....

brubbel:
There is indeed already a registration method. It makes the get_client_to_discovery obsolete.
https://github.com/FreeOpcUa/python-opcua/blob/fa52b1d863e1007cd4420c86f54485671329b00e/opcua/server/server.py#L173-L188

Issue #737, opened 2018-11-20
client.close_session() fails with TimeOutError when SecureChannel is not open.


brubbel:
Disconnecting a client fails with a timeout when the server previously closed the secure channel.
The reason is that during close_session(), the client tries to send a deletesubscriptions request to the server, although it can know in advance that it will fail, since SecureChannel is in 'closed' state.
One of the symptoms is that aborting with ctrl-c is slow, because of the TimeOutError.

The client should not send a deletesubscriptions request over a SecureChannel that is closed.
Will do PR.

```
Traceback (most recent call last):
  File "./opcua/client/client.py", line 272, in disconnect
    self.close_session()
  File "./opcua/client/client.py", line 494, in close_session
    return self.uaclient.close_session(True)
  File "./opcua/client/ua_client.py", line 277, in close_session
    data = self._uasocket.send_request(request)
  File "./opcua/client/ua_client.py", line 77, in send_request
    data = future.result(self.timeout)
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 434, in result
    raise TimeoutError()
concurrent.futures._base.TimeoutError
```

https://github.com/FreeOpcUa/python-opcua/blob/fa52b1d863e1007cd4420c86f54485671329b00e/opcua/client/ua_client.py#L265-L278

PR #738, opened 2018-11-20
fix: don't send unsubscribe request if SecureChannel is closed

brubbel:
Cfr. https://github.com/FreeOpcUa/python-opcua/issues/737


oroulet:
need to be rebased for some reasons.

brubbel:
rebase ok.

PR #743, opened 2018-11-20
fix client.disconnect() OSError: [Errno 9/107] EBADF/ENOTCONN

brubbel:
EBADF [errno 9]: socket connection is closed.
ENOTCONN [errno 107]: socket never connected.
cfr. https://www.gnu.org/software/libc/manual/html_node/Receiving-Data.html

### DIFF
--- a/opcua/client/ua_client.py
+++ b/opcua/client/ua_client.py
@@ -148,7 +148,7 @@ class UASocketClient(object):
         try:
             self._socket.socket.shutdown(socket.SHUT_RDWR)
         except OSError as exc:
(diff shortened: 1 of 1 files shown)


brubbel:
It seems that Py2.7 raises the same errors in a different format. Please don't merge yet.

brubbel:
From https://docs.python.org/3.6/library/exceptions.html#OSError

> Changed in version 3.3: EnvironmentError, IOError, WindowsError, socket.error, select.error and mmap.error have been merged into OSError, and the constructor may return a subclass.

Issue #733, opened 2018-11-19
Discovery clients should be outside of server.py code.


brubbel:
It seems to me that registration to OPC-UA discovery servers may be handled outside of the Server class, as a separate entity. Looking at the call graph, it would make the server.py code cleaner.
Will do a PR, then feel free to resist.

brubbel:
https://github.com/FreeOpcUa/python-opcua/blob/fa52b1d863e1007cd4420c86f54485671329b00e/tests/tests_server.py#L54-L56
Setting an app_uri for register_to_discovery is useless since the registration basically doesn't involve the OPC-UA server that one wants to register. It just sets up a new client to announce the url of said server to the discovery server. So:

OPC-UA server with endpoint x (the one we want to announce)
    \_ associated discovery client -> [endpoint x announcement] -> OPC-UA discovery server.

brubbel:
Confusing member names:
https://github.com/FreeOpcUa/python-opcua/blob/fa52b1d863e1007cd4420c86f54485671329b00e/opcua/server/server.py#L91-L92

discovery_clients should be called registration_clients. They take care of announcing/registering the server with an LDS server. discovery clients, on the other hand, connect to the LDS to find out the available server end points.

brubbel:
Race condition when periodic re-registrations to LDS (asyncio thread) and server is stopped from main thread.

```
ERROR:asyncio:Exception in callback Server._renew_registration()
handle: <TimerHandle when=52137.077456338004 Server._renew_registration()>
Traceback (most recent call last):
  File "/usr/lib/python3.6/asyncio/events.py", line 145, in _run
    self._callback(*self._args)
  File "./opcua/server/server.py", line 204, in _renew_registration
    client.register_server(self)
  File "./opcua/client/client.py", line 337, in register_server
    return self.uaclient.register_server(serv)
  File "./opcua/client/ua_client.py", line 364, in register_server
    data = self._uasocket.send_request(request)
  File "./opcua/client/ua_client.py", line 75, in send_request
    future = self._send_request(request, callback, timeout, message_type)
  File "./opcua/client/ua_client.py", line 66, in _send_request
    self._socket.write(msg)
  File "./opcua/common/utils.py", line 118, in write
    self.socket.sendall(data)
BrokenPipeError: [Errno 32] Broken pipe
```

brubbel:
[...]

PR #747, opened 2018-11-21
wip: opc-ua registration and discovery refactoring

brubbel:
See https://github.com/FreeOpcUa/python-opcua/issues/733
Fixes multiple issues with current discovery/registration implementation.
Pulls code outside server/client.py classes into a separate entity.

### DIFF
--- a/opcua/server/discovery_service.py
+++ b/opcua/server/discovery_service.py
@@ -1,0 +1,49 @@
+
+from opcua import ua
+
+class LocalDiscoveryService(object):
+
+    class ServerDesc(object):
+        def __init__(self, appDesc, uaDiscoveryConfiguration=None):
+            self.applicationDescription = appDesc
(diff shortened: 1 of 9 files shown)


oroulet:
start PR name with wip: and remove the tag when you think it is finished

opcua/server/discovery_service.py
```diff
@@ -0,0 +1,49 @@
+
```
oroulet: The name Service is a bit confusing since OPCUA services are the ones defined in spec. Mayve this is better called DiscoveryUtils DiscoveryRegistrator... I let you think about it

oroulet: We should also call services the classes offering the functions defined by spec

oroulet: OK now I read more of your code and maybe name service is correct

PR #752, opened 2018-11-23
pull user authentication out of InternalServer class

brubbel:
User authentication code should not be interleaved with InternalServer.

Also: private_key from InternalServer -> Server.


opcua/server/internal_server.py
```diff
@@ -339,6 +296,10 @@ def __init__(self, internal_server, aspace, submgr, name, user=User.Anonymous, e
         self.logger.info("Created internal session %s", self.name)
         self._lock = Lock()
 
+    @property
```
oroulet: PEP8 says user_manager for class members

oroulet:
changes look good

brubbel:
done.

oroulet:
when you are finished , remove the wip flag

brubbel:
I think the page must be refreshed to see the updated title.

PR #754, opened 2018-11-23
readability: node.server is in fact an InternalSession instance

brubbel:
code readability.

It is highly confusing to understand the code flow for the Node.server attribute, as the server is actually an "InternalSession (server)" instance. Replaced all ocurrences of node.server by node.isession.

There is a fallback property which redirects the server attribute to node.isession, with a one-shot deprecation warning.


oroulet:
The problem is that Node is also used on client side, and calling it isession there is also very confusing. Not sure what the correct wording should be. maybe connection... or simply session. but that is a very invasive change that will break things for many people so I am not sure we can merge it.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

brubbel:
Yes, I did not thought about that.
But server is also confusing, because server can be an `InternalSession` or an `UaClient` instance. Neither of which is a Server which many people (including myself) will suppose that this attribute refers to.

You are right about the name isession, maybe it is better to find a common basis for InternalSession and UaClient. 

`activeSession` for example could point to a server or client session.
At server side, InternalSession could inherit from activeSession base class.
At client side, the UaClient could inherit from activeSession ABC.

For me, it was a useful insight that a Node instance is related to a session, rather than to a server.

brubbel:
proposed scheme:
image

brubbel:
Settled for session_server, because it was already used here:
https://github.com/FreeOpcUa/python-opcua/blob/9357c990c6a5c7ad6e24c073bceddc17b59cbea3/opcua/common/type_dictionary_buider.py#L121-L130

Up to you to decide to merge or not :+1:

Issue #757, opened 2018-11-25
Variant(value=anotherVariant) never sets the is_array flag


brubbel:
When a Variant-A is constructed from another variant-B, the latter of which the value is a list or a tuple, the is_array flag of Variant-A is always set to False.

This is because the self.is_array checks if the value in the constructor is a list or tuple, but not whether Variant-B value embeds a list:
https://github.com/FreeOpcUa/python-opcua/blob/bfd5d1227fe7d83759bbaf1c341b93ac5e98ea09/opcua/ua/uatypes.py#L693-L697

Affected constructor:
https://github.com/FreeOpcUa/python-opcua/blob/bfd5d1227fe7d83759bbaf1c341b93ac5e98ea09/opcua/ua/uatypes.py#L688-L710

PR #759, opened 2018-11-25
is_array flag is not set in Variant(value=otherVariant) init

brubbel:
See https://github.com/FreeOpcUa/python-opcua/issues/757

### DIFF
--- a/opcua/ua/uatypes.py
+++ b/opcua/ua/uatypes.py
@@ -696,19 +696,18 @@ class Variant(FrozenClass):
     """
 
     def __init__(self, value=None, varianttype=None, dimensions=None, is_array=None):
-        self.Value = value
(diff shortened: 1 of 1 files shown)


brubbel:
WIP->fix (travis error was not related to fix.)

brubbel:
Last change makes sure that a `float` is not stored as `int` type when `VariantType.Float` or `VariantType.Double`.

I had been doing some address-space exports and imports and it seems that the load_standard_address_space causes Variants with int's which are actually floats/double.
Variant->to_binary->from_binary->Variant then resulted in a consistency check failure as the from_binary produces the correct Variant.Value type.

brubbel:
rebase ok. removing WIP.

PR #763, opened 2018-11-27
WIP: pickle shelffile -> standard_address_space.sql

brubbel:
Refactoring the shelffile from a monolithic blob to sqlite3 db.
Advantages over pickle:
- Fast
- Small (25% of fill_address_space code)
- Not dependent on Python2/3 version
- Supports transactional read/(write TODO)
- Supports real-time persistence of user address-space (TODO)
- Strong typed: only INTEGER, TEXT and opc-ua to_binary() BLOB
- Drop-in replacement for memory-based AddressSpace

Built-in integrity check on generate_address_space.py dump.

Usage: shelffile=True when starting server.


zerox1212:
Interesting work.
1. How difficult to support more than NumericNodeID? This will definitely be required to persist a user address space (which would be a great feature). In the past I always had to save/restore the binary file.  
2.  I'm not sure it will be as fast as the shelf, but should be good enough.

oroulet:
What is the point of that change? The shelves stuff was just a ugly hack for very slow systems. Btw sqlite is not typed at all. You store whatever you want whereever you want.

zerox1212:
I'm in support of something that we can build into address space persistence.   
It reduces the following which I have had in my use cases:  
1. Don't need to rebuild starting address space from code or waste time parsing XML every time server starts.  
2. Can use UA address space to save data like parameters, HMI settings, etc. that would otherwise have to be handled in two spots (like when UA is "giving access" to an underlying system). This is especially useful if you create custom "drivers" for UA like Modbus TCP where some configuration data needs to be persisted. In other words it reduces code duplication in some applications.

My only other input is that if we really go this direction it might be better to create an interface like history has. That way end user could potentially implement their own persistence layer.

brubbel:
I am somewhat concerned about the divergence between the main branch and the parallel work that is done on the asyncio-server. I really would be able to do `await`, given that Python2 end-of-life is not so far in the future. Or is there already an official asyncio-branch on pypy?

brubbel:
@zerox1212 
Yes, then why not merge the user-space shelf in the history.sql?
So you have all data at one place. Then just on-the-fly fetch the latest value from history at request of that node. No need to store that data somewhere else or restore all data on server start.

zerox1212:
I think history is better done separately. The OPC UA spec defines how history works in detail.

If we are strictly keeping to the spec I do not know if it says anything about persisting server state.

In the end I was just giving some inputs. I am not able to do much work on this library myself. However I can review some stuff about this topic because I was one of the people working on shelf, dump, and load because I needed to restore server state. I think in the end I actually used XML export/import for persistence.

brubbel:
Maybe I'm wrong, but why not storing user aspace in the same sql file?
In a separate sql-table of course. No need to change how history works.
So that 1 file has all related data.

zerox1212:
Because history SQL implementation is rather simplistic and is not a good model to build on. It makes one table per historized node. You would not want this model for potentially thousands of nodes. Only nodes with history require this design.

Plus if node names change or other such things data is orphaned in the database because there is not really a mechanism to remove data.

oroulet:
>> I have never seen an SQLite database in WAL mode to become fully corrupt.
I have not see it myself , bu I have heard many people complain about easy corruption in sqlite. 

opcua-asyncio is a python 3.6 only fork so we do not care about python2 there. I think the approach is correct but yest I am concerned about divergence, that is why I thing, large restructure work should wait for that work to stabilize. better work on improving opcua-asyncio than python-opcua

concerning the shelve stuff I thing the people wanting to save things in the opcua database have a fundamental misunderstanding of what opcua is. opua is a communication protocol, saving things inside the communication protocol does not make sense at all, the underlying system should save its state, opcua is just here to expose data from the underlying system. I am wondering if I made some errors in the API that make people think they should do that kind of things....

but I agree that using a sqlite db to store the standard address space might have an interest, this standard address space slows down starting time and uses memory for something that is ignored 99.99 % of the time. But onece again this should be something to implement in asyncio-opcua

@brubbel if you have some time or want to learn some modern python things, then look at the fork, try it and try to improve it. We also need to make a sync wrapper around it, there are a lot of things to things about here. probably make some classes with decorator and maybe the async Node should be called AsyncNode and the syn node Node to stay compatible with python-opcua... or just accept that we not completly API compatible anymore

brubbel:
I agree that opcua-ua is not meant as storage space. However, It is my opinion that the library must provide a simple means to override some parts such as the address space. With that in mind I think it is a good idea to support a strategy for that.
For example: address space must a dict-compatible + get_attribute_value + set_attribute_value etc.

I would like to start using the opcua-asyncio for the following reasons:
* I see there are some nice modifications: __aenter__ instead of start and __aexit__ for stop.
* In the current version there is no way for the user to use an async history and async address space.
Synchronous I/O really kills performance in my case.
* I want to provide my own asyncio thread, instead of relying on the one that is started in the background.

However:
* I don't have much time, as all of you. So I really appreciate your views on the best path.
* I am in favor of a working version at any time :-) Breaking API's is not a concern. Bugs are. With that in mind: do you consider the asyncio branch stable enough?

oroulet:
>> It is my opinion that the library must provide a simple means to override some parts such as the address space
I agree that this could be an improvment, if you do that you need to provide some API to notify the system for change in values so datachange events can be generated. Such a change will require quite a lot fo work

>> do you consider the asyncio branch stable enough?
I have no idea :cry:  I haven't had time to try it t but I looked at the changes and they all looked correct, 
@cbergmiller what is the status and plan currently
Ideally we should first work on the sync API so tests can be run without having to modify all of them.

zerox1212:
@oroulet said OPC UA is not really meant for persistence (except history I guess...). For me it's still a valid request to want persistence of user nodes, mainly because it can make prototyping UA software faster which was my use case. In other words I'm lazy and I don't want to write a persisting data layer and then create a UA address space that will basically be the exact same thing. Maybe my use case is just too special.

An address space design that could can be overridden would be a nice option.

brubbel:
So here it is, based on your comments:
Instead of providing a shelffile, you can now provide your own address space.

```
with AddressSpace() as myAspace:
    server = Server(aspace=myAspace)
```
Of course this address space is empty, but you can load from sqlite:
```
with AddressSpaceSQLite(sqlFile='/path/to/standard_address_space.sql') as myAspace:
    server = Server(aspace=myAspace)
```
Now you can start to populate the address space with your own nodes, as usual.
Then you can simply dump your own address space:
```
with AddressSpaceSQLite(sqlFile='/path/to/standard_address_space.sql') as myAspace:
    server = Server(aspace=myAspace)
    <...>
    with AddressSpaceSQLite(cache=aspace, sqlFile=path) as aspace_sql:
        aspace_sql.dump(namespaceidx=2)
```
And to load, simply stack it on top of the standard address space:
```
with AddressSpaceSQLite(sqlFile='/path/to/standard_address_space.sql') as stdAspace, \
  AddressSpaceSQLite(sqlFile='/path/to/my_address_space.sql', cache = stdAspace) as myAspace:
    server = Server(aspace=myAspace)
```

@oroulet said:
> if you do that you need to provide some API to notify the system for change in values so
> datachange events can be generated. Such a change will require quite a lot fo work
In the proposed implementation, you still read and write to the server via the usual set/get_value API, so all is fine.

@zerox1212 saild
> An address space design that could can be overridden would be a nice option so user like
> me can break the rules. :)
Just inherit from `AddressSpace`, which is basically a threadsafe dict and you're good to go.
`AddressSpaceSQLite` can be used as a reference implementation.

The current `AddressSpaceSQLite` implementation still requires a aspace.dump() to save the address space to disk. However, the path is now open to implement on-the-fly writes to sqlite3.

oroulet:
I haven't looked at this closely yet, but one more thing is that you should be really carefull with performance, this is an arear which we had to really optimized. I see that your replaced several dicts access with functions, this might slow downn things quite a lot for example. It was also optimized many places to avoid searching dicts.
so basically you will need to setup some performance comparison between your approach and the old one.

brubbel:
True, the first read will be slow as data is fetched from disk. After that there should be no difference.
What I'm still considering is a way to solve searches for non-existent nodes in the aspace (BadNodeIdUnknown), since there would be a repititive fallback to (slow) database.

Also, the current implementation is synchronous and the async-opcua could really shine here as you can continue servering while waiting for IO, or you could even restore cache during idle time.

brubbel:
TL;DR; address space `.dump()` is now obsolete :-) real-time updates of database/backend.

Usage:
```
with SQLite3Backend(sqlFile='/path/to/standard_address_space.sql') as stdBackend,
       SQLite3Backend(sqlFile='my_address_space.sql', readonly=False) as myBackend
       AddressSpaceSQLite(stdBackend) as stdAspace, \
       AddressSpaceSQLite(myBackend, cache = stdAspace) as myAspace:
    server = Server(aspace=myAspace)
    <...>
```

How it works:
1. stdBackend provides readonly access to the standard address space on disk.
2. myBackend provides r/w access to your custom address space on disk.
3. stdAspace reverts to stdBackend for nodes not in cache.
4. myAspace reverts to myBackend for nodes not in (stdAspace, cache)
5. Server uses myAspace as write-through cache.

Data becomes persistent by:
* automatic checkpoints every 90sec, if Server_ServerStatus_CurrentTime ticks are active (default)
* or by calling myBackend.wal_checkpoint() manually
* or by a clean shutdown.

Advantages.
* Share database with history.sql database becomes possible. Or not, whatever you want.
* (TODO) provide your custom backend, may even retrieve data from network etc.

Around 10k record updates/sec on my i3-3227U, when multiple statements in a single transaction.
So if you need high-speed updates, override the execute_read method to bypass the automatic commits and bundle writes in a single transaction.

zerox1212:
Looks nice. Let's wait for @oroulet  to review.

If you don't define `aspace` everything works as normal correct?

Only small points I see are that the `x.sql` files should not be committed to the repo and an usage example could be nice.

brubbel:
Correct, everything remains as before when the user does not define aspace.
Only the pre-generated standard_address_space.sql is added to the repo.

brubbel:
Added a shortcut for loading the standard address space.

```
with SQLite3Backend(sqlFile='my_address_space.sql', readonly=False) as myBackend, \
       StandardAddressSpaceSQLite() as stdAspace, \
       AddressSpaceSQLite(backend=myBackend, cache=stdAspace) as myAspace:
    assert(isinstance(myAspace, AddressSpace)) # <- interface like HistoryStorageInterface
    server = Server(aspace=myAspace)
    <...>
```

brubbel:
Added examples/server-persistent.py

* Open en close the server multiple times to fill my_address_space.sql
* Write values from your client into the server and see that they are also persisted.

oroulet:
The idea of putting the standard address space in a sqlite db is not bad, that is fine, but I hada a few questions:
* can we have persistence off by default? This is really against the idea of a protocol,
* how can users force update of sql address space? or how can we have the address space automatically detect the standard has changed (due to new realease or whatever) we need a version string somewhere

brubbel:
> can we have persistence off by default?

Examples without persistence:

**Option 1 (default):** no custom address space.
`server = Server(aspace=None)` or `server = Server()`

**Option 2:** read-only standard address space from sql. Non-persistent cache.
```
with StandardAddressSpaceSQLite() as stdAspace:
    server = Server(aspace=stdAspace)
```

**Option 3:** read-only std aspace + read-only custom aspace. Non-persistent cache.
```
with SQLite3Backend(sqlFile='my_address_space.py', readonly=True) as backend, \
    StandardAddressSpaceSQLite() as stdAspace, \
    AddressSpaceSQLite(backend=backend, cache=stdAspace) as myAspace:
server = Server(aspace=myAspace)
```

> how can users force update of sql address space?

As is done in this PR, I would suggest to keep the standard_address_space.sql within the distribution,
in the opcua/server/standard_address_space/ directory. It is generated together with the `standard_address_space_partx.py` files.

The version is already embedded in the address space:
image

oroulet:
>> The version is already embedded in the standard address space:

so maybe you should read that value at startup and regenerate address space if current addresse space is different than the one in code?
Maybe you can generate a file code ua_std_version.py in opcua/ua/server/address_space when callinge generate_address_space.py and compare that value to the version in sql db?

oroulet:
I may forget something, but my proposition was

if running sql
   load only opcua.ua.ua_std_version.py
   check that version to version in sql
   load standard address space and save it to disk if sql version < opcua.ua.ua_std_version.py

oroulet:
oops I did not see that you put a huge a blob file in repository. Yes this is a very bad idea ;-).
I did not read your entire code yet  and was expecting that you populate the sql db from the address space in memory... but was you idea to distribute the address space as an sql blob with python-opcua? That is not going to be very popular....
I think that you should populate the db at first startup if it does not exist or if outdated. you can also add a script to generate the blob if some people want to generate it by hand

oroulet:
as written earlier if you manage to implement  that correctly I thing we should considere using sql as default to reduse startup time. But then we really need to make sure it does not give and performance drawback for high speed read_write on client and server side.

brubbel:
confusion solved :-)
Yes, my idea was that standard_address_space.sql would come along with all those standard_address_space_partx.py files, basically representing the same info in a different format.

image

oroulet:
but as written above binary blobs in pytnon distributions are really not popular (neither in a git repository). if you want to save something in repository it should be some sql code for example, or give a script to generate it

oroulet:
I think that you should populate the db at first startup if it does not exist or if outdated. you can also add a script to generate the blob if some people want to generate it by hand

brubbel:
That would be a reasonable solution. I would however advise to keep the all the original standard_address_space.py code as default in the future, as it is slower but also easier to debug.

brubbel:
removed standard_address_space.sql from repo + history rewrite to remove from all intermediate commits.

oroulet:
@brubbel what do you save in db btw? I can read code but if you explain also I might better understand. do you save some stuff as python objects? or do you decompose eveything into basic types  organized in tables? maybe you should make a figure to document it and add to repository so people can understand

brubbel:
I'm considering several options.
As explained [here](https://python-packaging.readthedocs.io/en/latest/non-code-files.html), it could be the most straightforward solution to package the data file along with the pypy distribution.
That way, there is no problem that the user must generate the files themself and updates are also solved.
In the other case extra things todo:
* where to store the generated file?
* when to update the generated file?
* Keep it simple? The user can always opt to use his own standard_address_space.py with the tools provided, and just consider it like any other address space in the stack.

oroulet:
>>where to store the generated file?
>>when to update the generated file?
>> Keep it simple?

save everything in current directory, or allow users to pass directory as argument. This is very easy ang good enough. easier that finding the path in the installed package and usually correct for most applications

oroulet:
how big is that file btw?

brubbel:
> easier that finding the path in the installed package
That is what the StandardAddressSpaceSQLite() does for you.
The file is 3.8M.

brubbel:
Attributes table:
image
* _Id as primary key to avoid duplicates
* NodeId and AttributeId for search
* DataValue Object
* Human-readable column

brubbel:
References table:

image

* _Id as primary key = from+to+direction
* NodeId for lookup
* ReferenceTypeDescription = embedded object.
* Human-readable description

oroulet:
ok so in fact you decompose most objects apart from Variant.
Why only numerical nodeids btw? sqlite is not type sensitive so it should handle everything

oroulet:
what if a node is not numerical but string or somethings else?

brubbel:
Good remark :-) That is not solved yet, but there is no reason why it can't be used as lookup key (not supported now, let this be clear)
If you want to mix both numeric+string in 1 single node, waiting for external input :-)

oroulet:
in fact you probably cannot use only a numeric value as db index. a nodeid is namespace index AND a value which can be off different type, so you probably either need to use some blob as key (might break between python versions or whatever) or string (slow) or binary data (maybe faster but hard to read for humans). Maybe you could use tuple? (ns, key) is htis supported in sqlite?

brubbel:
Ah, but the numeric nodeid is not required!

The lookup key is the OPC-UA-defined binary representation of this nodeid (nodeid_to_binary). If this breaks, than it means that the opc standard binary encoding has changed.

You can use a string nodeid in there (I'll change some lines of code), but you cannot expect it to show up later on if you request the same node with it's numeric identifier. But that won't work with the current nodeid.__hash__() in the main branch either.

brubbel:
I also want to stress that you really need the BLOB as primary key, so that `INSERT OR REPLACE INTO` statements work fast. Original code used lookups with 2 columns (Identifier and nsidx) and this turned out to be 20x slower.

oroulet:
but can we really use the binary node as primary key? the same node can have different encoding... but maybe that is what you are addressing with your numerical stuff?

brubbel:
> that is what you are addressing with your numerical stuff?

Exact. You need a fixed BLOB for fast lookups, that's where all the performance is.
SQLite even supports LIKE '%xyz%' pattern matching for BLOBs but that is not used here.

oroulet:
That is probably broken, you can encode the same nodeid in different ways, using an extended nodeid for example. if you can use a blob why can't you use a tuple?

brubbel:
http://documentation.unified-automation.com/uasdkcpp/1.5.4/html/structOpcUa__ExpandedNodeId.html

> Detailed Description
> Extends the NodeId structure by allowing the NamespaceUri to be explicitly specified instead of using the NamespaceIndex.
> 
> The NamespaceUri is optional. If it is specified, the NamespaceIndex inside the NodeId shall be ignored.

It seems that expanded nodeid is basically nodeid + nsuri embedded. Shouldn't that be solved before the database access? Either store the namespace index or the nsuri, and access the data using one of the 2.

brubbel:
I can't imagine at this moment how this can work in the in-memory address space of the master branch.
Either you have a node without nsuri, or with nsuri. But you can you also create a nodeid and than have a dict lookup from expanded nodeid (without nsidx!) to map to the same nodedata?

brubbel:
> a nodeid is namespace index AND an identifier which can be off different type

The primary key, as shown above is a byte-wise concatenation of

oroulet:
The in memory stuff is using the hash as key. In fact that is probably what you should be using

brubbel:
See above:
> The hash cannot be used as key since it changes between python sessions for security reasons.
> see https://docs.python.org/3/reference/datamodel.html#object.__hash__

I did initially, but it didn't work for that reason
For now, this seems only be activated for string and binary hashes, not for integers, but I won't rely on that. The built-in hash is only guaranteed during 1 python session.

The hash key of a node is calculated here:
https://github.com/FreeOpcUa/python-opcua/blob/82ba92c4ccd7613161d383f1eeea765060e0d23d/opcua/ua/uatypes.py#L316-L317

where Identifier can be: 
https://github.com/FreeOpcUa/python-opcua/blob/82ba92c4ccd7613161d383f1eeea765060e0d23d/opcua/ua/uatypes.py#L257-L263

No NodeId.NamespaceUri involved in the hashing process.

zerox1212:
I'm guessing I know the answer to this, but can we just extend Node with a local ID field that has a GUID for DB look up? Maybe that's too hacky.

brubbel:
How would you make the link to GUID for nodes that are only on disk?
Say a remote server requests nodeid(i=123,ns=2) -> [?] -> sql lookup.

zerox1212:
Could we look for separate library that will hash in a predictable way i.e. not salt the hash?

brubbel:
Have been looking at fast hash algorithms like fnv-1a or murmur, but the only thing that they add is a risk for hash collisions and an extra dependency, and no way to calculate backwards to the original nodeid.

brubbel:
Even worse:
32bit hash -> 50% collision rate since a node has multiple attributes (calculated with 8 attr/node)
64bit hash -> 1e-10 (ok)
binary numericNodeId -> 0%

brubbel:
Support for all other nodeid types other than numeric/twobyte/fourbyte.

brubbel:
SQLite needs write access to read-only standard_address_space.sql file. Added the `immutable=true` flag when opening the database. Not supported by Python2.x, checked the source code of cPython for that.

Py2.x users can't use the wrapper:
```
with StandardAddressSpaceSQLite() as stdAspace:
    <...>
```
But will have to copy the standard_address_space.sql and do:
```
with SQLite3Backend(sqlFile='/path/to/standard_address_space.sql') as stdBackend, \
       AddressSpaceSQLite(backend=stdBackend) as stdAspace:
    <...>
```
where /path/to/standard_address_space.sql is writable and also in a directory with write access.

brubbel:
multithreading: sqlite does not allow interleaving write-commits between multiple threads.
Fixed by protecting write + commit (if required) under the same Lock().

brubbel:
Rebased PR to master.

brubbel:
Rebased PR to master.

PR #781, opened 2019-01-09
Add example of how to implement a custom user manager

helmut-jacob:


examples/server-user-manager.py
```diff
@@ -0,0 +1,44 @@
...
+    # setup our own namespace, not really necessary but should as spec
+    uri = "http://examples.freeopcua.github.io"
+    idx = server.register_namespace(uri)
+
+    server.user_manager.set_user_manager(simple_user_manager)
+
+    # starting!
+    server.start()
+
+    try:
```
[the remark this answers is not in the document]
oroulet: apart from that , everything looks good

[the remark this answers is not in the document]
helmut-jacob: Pushed without that comment.

examples/server-user-manager.py
```diff
@@ -0,0 +1,43 @@
...
+    def embed():
+        vars = globals()
+        vars.update(locals())
+        shell = code.InteractiveConsole(vars)
+        shell.interact()
+
+
+from opcua import ua, Server
+
+def simple_user_manager(isession, name, pwd):
```
oroulet: same issue as the other PR. It might not be your code, but when readong the word manager I am expecting a class with a few methods.

oroulet: This user_manager stuff loks very strange. Ideally it should have been an interface class

class UserManagerInterface:
     def check_user_token(self, isession, token):
            copy code
     def check_password(self, isession, user, passwd)
           raise NotImplementedError

then have a default one:
class SimpleUserManager(UserManagerInterface):
         def check_password(self, isession, user_name, passwd):
                 if self.allow_remote_admin and userName in ("admin", "Admin"):
                         isession.user = UserManager.User.Admin
                 return True

helmut-jacob: > same issue as the other PR. It might not be your code, but when readong the word manager I am expecting a class with a few methods.

Fully agreed, this makes more sense. So let's drop the example code for now and add it once the user manager code stabilizes.

oroulet: @helmut-jacob do you look at this user manager stuff or should I do it? I am really busy currently so it might take a while

helmut-jacob:
Closing for now until somebody reworks the usermanager stuff.

PR #783, opened 2019-01-12
Implement the security endpoints check for identification

### LINKED ISSUE #782: How to implement the security ID check
I have to setup a server that allows clients to login only with username and password.
It work with the example file below. 
But I can not disable the Anonymous connection.

```
import time

from opcua import ua, Server
from opcua.server.user_manager import UserManager

# users database
users_db = {
    'user1': 'passwd1',
    'user2': 'passwd2',
    'user3': 'passwd3',
}


# user manager
def user_manager(isession, username, password):
    print(isession, username, password)
    isession.user = UserManager.User
    return username in users_db and password == users_db[username]


if __name__ == "__main__":

    # setup our server
    server = Server()
    server.set_endpoint("opc.tcp://0.0.0.0:4840/freeopcua/server/")

    # load server certificate and private key. This enables endpoints
    # with signing and encryption.
    server.load_certificate("certificate-example.der")
    server.load_private_key("private-key-example.pem")

    # set all possible endpoint policies for clients to connect through
    server.set_security_policy([
        # ua.SecurityPolicyType.NoSecurity,
        ua.SecurityPolicyType.Basic256Sha256_SignAndEncrypt,
        # ua.SecurityPolicyType.Basic256Sha256_Sign,
    ])

    # set the security endpoints for identification of clients
    # self.server.set_security_IDs(["Anonymous", "Basic256Sha256", "Username"])
    server.set_security_IDs(["Username"])

    # set the user_manager function
    server.user_manager.set_user_manager(user_manager)

    # starting!
    server.start()

    print("Endpoints : ", str(server.get_endpoints()).replace(',', '\n'))

    try:
        while True:
            time.sleep(5)
    finally:
        # close connection, remove subscriptions, etc
        server.stop()
```

In the server.py file, it's written that :

```
    def set_security_IDs(self, policyIDs):
        """
            Method setting up the security endpoints for identification
            of clients. During server object initialization, all possible
            endpoints are enabled:

            self._policyIDs = ["Anonymous", "Basic256Sha256", "Username"]

            E.g. to limit the number of IDs and disable anonymous clients:

                set_security_policy(["Basic256Sha256"])

            (Implementation for ID check is currently not finalized...)

        """
        self._policyIDs = policyIDs
```
Someone can tell me where to look to implement this ? (I'm new in opc ua world)

Thanks

okyame:
close #782

### DIFF
--- a/opcua/server/internal_server.py
+++ b/opcua/server/internal_server.py
@@ -310,18 +310,44 @@ class InternalSession(object):
 
     def activate_session(self, params):
         self.logger.info("activate session")
+
         result = ua.ActivateSessionResult()
         if self.state != SessionState.Created:
             raise utils.ServiceError(ua.StatusCodes.BadSessionIdInvalid)
+
+        # FIXME: Get the server available UserIdentityToken policy's ids. I don't how to get it in other way.
+        # iserver_policy_ids = ["anonymous", "certificate_basic256sha256", "username"]
+        # Note that it is different from server._policyIDs = ["Anonymous", "Basic256Sha256", "Username"]
+        iserver_policy_ids = []
+        edp = self.iserver.get_endpoints()[0]  # FIXME: is it the same list whatever endpoint ?
+        for token in edp.UserIdentityTokens:
+            iserver_policy_ids.append(token.PolicyId)
+
+        # Check if params UserIdentityToken is authorized by the server.
+        params_id_token = params.UserIdentityToken
+        if params_id_token.PolicyId not in iserver_policy_ids:
+            raise utils.ServiceError(ua.StatusCodes.BadIdentityTokenInvalid)
+
+        # Route to the correct manager
+        if isinstance(params_id_token, ua.AnonymousIdentityToken):
+            pass  # TODO: Call AnonymousIdentityToken Manager
+
+        elif isinstance(params_id_token, ua.UserNameIdentityToken):
+            if self.user_manager.check_user_token(self, params_id_token) == False:
+                raise utils.ServiceError(ua.StatusCodes.BadUserAccessDenied)
+
+        elif isinstance(params_id_token, ua.X509IdentityToken):
+            pass  # TODO: Call X509IdentityToken Manager
+
+        else:  # Not implement IdentityToken
+            raise utils.ServiceError(ua.StatusCodes.BadIdentityTokenInvalid)
+
         self.nonce = utils.create_nonce(32)
         result.ServerNonce = self.nonce
         for _ in params.ClientSoftwareCertificates:
             result.Results.append(ua.StatusCode())
         self.state = SessionState.Activated
-        id_token = params.UserIdentityToken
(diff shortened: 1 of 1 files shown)


okyame:
No more need of PR #620

examples/server_certificate_manager.py
```diff
@@ -0,0 +1,53 @@
+from opcua import ua, Server
+from opcua.crypto import uacrypto
+
+import time
+
+import sys
+sys.path.insert(0, "..")
+
+
+def simple_certificate_manager(isession, certificate, signature):
```
oroulet: I have not looked in detail at this. But is it enough with a function? Should'nt we use a class with a few methods? just asking

oroulet: if it is just a function, then using the name manager is a bit wrong. A manager is an object.  a function would be called something like set_certificate_validation_function or whatever

opcua/server/internal_server.py
```diff
@@ -310,18 +318,45 @@ def close_session(self, delete_subs=True):
...
     def activate_session(self, params):
         self.logger.info("activate session")
+
         result = ua.ActivateSessionResult()
         if self.state != SessionState.Created:
             raise utils.ServiceError(ua.StatusCodes.BadSessionIdInvalid)
+
+        # FIXME: Get the server available UserIdentityToken policy's ids. I don't how to get it in other way.
+        # iserver_policy_ids = ["anonymous", "certificate_basic256sha256", "username"]
+        # Note that it is different from server._policyIDs = ["Anonymous", "Basic256Sha256", "Username"]
```
oroulet: yes that seems to be the correct way to get policy IDs . Just write a comment that we assume that the same polcies are used n all endpoints.. maybe add a FIXME. I am not sure what spec says but in theory it should be possible to have different policies per endpoints

opcua/server/internal_server.py
```diff
@@ -310,18 +318,45 @@ def close_session(self, delete_subs=True):
...
+            iserver_policy_ids.append(token.PolicyId)
+
+        # Check if params UserIdentityToken is authorized by the server.
+        params_id_token = params.UserIdentityToken
+        if params_id_token.PolicyId not in iserver_policy_ids:
+            raise utils.ServiceError(ua.StatusCodes.BadIdentityTokenInvalid)
+
+        # Route to the correct manager
+        if isinstance(params_id_token, ua.AnonymousIdentityToken):
+            pass  # TODO: Call AnonymousIdentityToken Manager
```
oroulet: we should at least check that "anonymous" is in the list of Policies and return an UaError if not

oroulet: Can't we use the user_manager for that? maybe just have a bool member called something like "allow_anonymous"? not sure I have not looked at that code for very long...

oroulet:
thanks for the patch anyway. this looks good

oroulet:
@okyame what is the status?

PR #814, opened 2019-03-26
Fix object creation

kozachynskyi:
Hi again, thank you developers for your work.

Issue: while creating object via `Node.add_object()`, the resulting object has either: Namespace index in DisplayName or Default Namespace 0 in BrowseName. I fixed it and checked functionality:

with the new version user can supply `bname` attribute to `Node.add_object()` as `QualifiedName`, as `QualifiedName.to_string()` and simple string and all versions worked properly.

Nevertheless, I'm not sure if my test are sufficient.


oroulet:
looks correct

PR #817, opened 2019-04-02
Register nodes

sdransfeld:
Add some functions to increase read/write speed. Tested with Siemens PLC.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -570,3 +570,25 @@ class Client(object):
         This enums will be available in ua module
         """
         return load_enums(self)
+
+    def register_nodes(self, nodes):
+        """
+        Register nodes for faster read and write access
+        """
+        nodes = [node.nodeid for node in nodes]
+        nodeids = self.uaclient.register_nodes(nodes)
+        for node, nodeid in zip(nodes, nodeids):
+            node.oldnodeid = node.nodeid
+            node.nodeid = nodeid
+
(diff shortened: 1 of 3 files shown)


oroulet:
The low level stuff is completely done, but I am not very keen in changing self.nodeid of a Node. That may surprise users and lead to all kinds of strange behaviours. What about simply returning a list of new nodes? Maybe even creating a Node superset that has a old_nodeid attribute of you want

sdransfeld:
But, when using node.register(), should this return a new node? Then the
user must track two nodes. In my opinion this is not very elegant and
intuitive either.

Another possibility is that the registered node is hidden for the user, and
only used in communication with the server. The problem with this solution
is that it probably requires a lot of small changes.

Den tir. 2. apr. 2019, 18:18 skrev oroulet <notifications@github.com>:

oroulet:
I agree that returning a new node is strange,. but this is what the opcua protocol specifies, so I am afraid to hide it away. 
you can always change the node id locally in you application if you want, in order to not have to handle two nodes
btw I am quite sure that in most applications, you will register your node then overwrite you local node variable with the new registered node. There is probably no reason to keep track of original node after registration

sdransfeld:
If you unregister the node, you would want the old nodeid back. So to have
this functionality we need to store the original nodeid.

The patch is currently wrong, need some fixups.

oroulet:
merged

PR #820, opened 2019-04-13
Fix Server-field access for server registration at discovery server

testunde:
If I want to register my server to a discovery server, I get the following error:
```
  File "testServer.py", line 112, in <module>
    register.register_to_discovery(serverToRegister=server, discoveryUrl="opc.tcp://localhost:52010/discovery/")
  File "/usr/lib/python3.7/site-packages/opcua/server/registration_service.py", line 56, in register_to_discovery
    self._renew_registration(serverToRegister, registrationClient, period=period)
  File "/usr/lib/python3.7/site-packages/opcua/server/registration_service.py", line 80, in _renew_registration
    self._register_server(serverToRegister, registrationClient)
  File "/usr/lib/python3.7/site-packages/opcua/server/registration_service.py", line 113, in _register_server
    uaRegSrv.ServerUri = serverToRegister.application_uri
AttributeError: 'Server' object has no attribute 'application_uri'
```

To fix this, you can simply change the accessing line to the actual field-name (`serer._application_uri`, see this merge request), or it is possible to create a specific `server.get_application_uri()` method for the server class.

Best regards,
testunde

### DIFF
--- a/tests/tests_common.py
+++ b/tests/tests_common.py
@@ -521,8 +521,8 @@ class CommonTests(object):
         self.assertEqual([1], val)
 
     def test_use_namespace(self):
-        idx = self.opc.get_namespace_index("urn:freeopcua:python:server")
-        self.assertEqual(idx, 1)
+        idx = self.opc.get_namespace_index("http://opcfoundation.org/UA/")
+        self.assertEqual(idx, 0)
         root = self.opc.get_root_node()
         myvar = root.add_variable(idx, 'var_in_custom_namespace', [5])
         myid = myvar.nodeid
--- a/tests/tests_common.py
+++ b/tests/tests_common.py
@@ -521,8 +521,10 @@ class CommonTests(object):
         self.assertEqual([1], val)
 
     def test_use_namespace(self):
-        idx = self.opc.get_namespace_index("http://opcfoundation.org/UA/")
-        self.assertEqual(idx, 0)
+        idx_orig = 1
+        uri = self.opc.get_namespace_array()[idx_orig]
+        idx = self.opc.get_namespace_index(uri)
+        self.assertEqual(idx_orig, idx)
         root = self.opc.get_root_node()
         myvar = root.add_variable(idx, 'var_in_custom_namespace', [5])
         myid = myvar.nodeid
--- a/opcua/server/registration_service.py
+++ b/opcua/server/registration_service.py
@@ -110,7 +110,7 @@ class RegistrationService(object):
         if uaDiscoveryConfiguration is provided, the newer register_server2 service call is used
         """
         uaRegSrv = ua.RegisteredServer()
-        uaRegSrv.ServerUri = serverToRegister.application_uri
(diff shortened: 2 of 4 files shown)


oroulet:
@testunde your changes seems to break some tests. Can you have a look since you just looked at that code?

testunde:
I really do not get it, in the server class (server.py) there is only a member `_application_uri`, but the tests use `application_uri`, am I getting something wrong here?

oroulet:
I have not looked closely at it. But there is no reason to have it hidden. So application_uri seems correct

testunde:
> I have not looked closely at it. But there is no reason to have it hidden. So application_uri seems correct

okay, so either way we use the set/get method I used in the last commit (0cedcd4) of this merge, or we expose application_uri. For the latter the tests seem to work on my local machine.
So what is preferred? Then I will make a final commit so you can merge it.

oroulet:
@testunde OK now I really had a look
* there is already a set_application_uri because setting application uri also requires to set som nodes inside server address space
* _application_uri should be hidden because of explanation above
* adding get_application_uri is fine, it was missing
* the tests were completely broken. They were outdated and for some strange/random reason still work: you can fix them by simply calling the existing set_application_uri method

testunde:
Okay, so just applying the `set_application_uri()` method breakes `test_use_namspace` in tests_common.py (see commit 0ef11f8). This is because using the set method overwrites the namespace with index 1, which the test_use_namespace thinks is still the default one. I just set the test to use the namespace index 0, which also works for the client-side tests.

tests/tests_common.py
```diff
@@ -521,8 +521,8 @@ def test_array_size_one_value(self):
         self.assertEqual([1], val)
 
     def test_use_namespace(self):
-        idx = self.opc.get_namespace_index("urn:freeopcua:python:server")
-        self.assertEqual(idx, 1)
+        idx = self.opc.get_namespace_index("http://opcfoundation.org/UA/")
+        self.assertEqual(idx, 0)
```
oroulet: OK that test was also outdated. usually users should only create nodes in custom namespace. so the best is probably to  set again the application URI here 
```
self.opc.set_application_uri("urn:freeopcua:python:server")
 idx = self.opc.get_namespace_index("urn:freeopcua:python:server")
 self.assertEqual(idx, 1)
```
that should work

oroulet: otherwise everything is fine

testunde: @oroulet The problem with this solution is with the Client-side test of this part, because then `opc` is a Client instance which has no `set_application_uri()` member. Tests say the same.

oroulet: ok then what about:
```
idx_orig = 1
uri = self.opc.get_namespace_array()[idx]
idx = self.opc.get_namespace_index(uri)
self.assertEqual(idx_orig, idx)
```
that way we test several methods

testunde: sounds good :+1:

PR #830, opened 2019-04-29
[subscriptions] Fix race condition in client when creating subs

djipko:
Due to how Futures are used, we could timeout on the client side, but
still sucessfully create a subscription and add callbacks in the
_receiver thread. Once this happens, receiver thread will deadlock on
receiving the publish response from the server due to a while loop that
waits for the subscription_id to be set. This will never happen as it
was supposed to be set in the main thread, which timed out and failed.

The fix is to make sure we set the subscription_id in the _receiver
thread as a separate callback. This is not the ideal fix, which would
probably include re-designing the Subscription class API in a way to
allow checking for the staus, and changing the way
Client.create_subscription works. Since we don't want to make large
changes to this - let's just go with a simple fix that removes the
deadlock, and keeps the API pretty much compatible.

This commit also contains unittests that excercise the offending code
path and clearly avoids a deadlock.

### DIFF
--- a/opcua/client/ua_client.py
+++ b/opcua/client/ua_client.py
@@ -409,23 +409,29 @@ class UaClient(object):
         response.ResponseHeader.ServiceResult.check()
         return response.Results
 
-    def create_subscription(self, params, callback):
+    def create_subscription(self, params, publish_callback, ready_callback=None):
         self.logger.info("create_subscription")
         request = ua.CreateSubscriptionRequest()
         request.Parameters = params
         resp_fut = Future()
-        mycallbak = partial(self._create_subscription_callback, callback, resp_fut)
+        mycallbak = partial(self._create_subscription_callback, publish_callback, ready_callback, resp_fut)
         self._uasocket.send_request(request, mycallbak)
         return resp_fut.result(self._timeout)
 
-    def _create_subscription_callback(self, pub_callback, resp_fut, data_fut):
+    def _create_subscription_callback(self, pub_callback, ready_callback, resp_fut, data_fut):
         self.logger.info("_create_subscription_callback")
         data = data_fut.result()
         response = struct_from_binary(ua.CreateSubscriptionResponse, data)
         self.logger.debug(response)
         response.ResponseHeader.ServiceResult.check()
+        if ready_callback:
+            ready_callback(response)
         self._publishcallbacks[response.Parameters.SubscriptionId] = pub_callback
         resp_fut.set_result(response.Parameters)
+
+    def registered_subscriptions(self):
+        """Get all subscriptions we know about"""
+        return [cb.__self__ for cb in self._publishcallbacks.values()]
 
     def delete_subscriptions(self, subscriptionids):
         self.logger.info("delete_subscription")
--- a/opcua/common/subscription.py
+++ b/opcua/common/subscription.py
@@ -85,9 +85,12 @@ class Subscription(object):
         self._monitoreditems_map = {}
         self._lock = Lock()
         self.subscription_id = None
-        response = self.server.create_subscription(params, self.publish_callback)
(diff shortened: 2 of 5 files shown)


oroulet:
Thanks!!
very interesting. The while loop was added to fix a race condition and I have always been afraid of it. It looks like you found the issue...
I am still afraid of you proposition though, that stuff is too complicated...

what about just adding a timeout to the while loop, since anyway we are fucked in that case? other suggestions?

@cbergmiller this is some code you were looking at too. we need think if it is possible to implement a better solution for opcua-asyncio. the asyncio stuff allows to remove callbacks in many situations..

djipko:
@oroulet thanks for the quick response!

> what about just adding a timeout to the while loop, since anyway we are fucked in that case? other suggestions?

I think this is fine as well, and this was the first thing that came to my mind to! the problem is that it leaves the session object in an unusable state, because it's subscription_id is still left as None, so then for example a .delete() call won't work etc. 

This way I exposed it on `UaClient.registered_subscriptions()` method so they could be cancelled/cleaned up in high-level code without resorting to private members. Probably should have added an example in the test of why it's still useful to add this bit of complexity.

But if you're really worried - let me know and I can change the patch.

opcua-asyncio does not have this problem, as it actually has a separate `init` method https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/common/subscription.py#L88 which is what would be the cleanest fix here too, but it's a breaking change and probably more invasive than this.

djipko:
Unfortunately we can't switch to opcua-asyncio right now, but definitely plan to in the near future

oroulet:
>> thanks for the quick response!

good PR/analysis need to good response ;-)

>> opcua-asyncio does not have this problem, as it actually has a separate init

Are you sure? since both operations are done in the same method, then we can also end up with a successfull creation and a timeout so self.subscription_id is not set... but maybe there is not timeout here since we simply await?

djipko:
> but maybe there is not timeout here since we simply await

Exactly - the problem in the sync world is that even if we time out on the data_future, we still *could* receive the CreateSubscriptionResponse, as we only time out in the "main" thread, but do not cancel callbacks on the `_receive` thread, which calls `_create_subscription_callback` and sets up callbacks. No such thing happens in asyncio world - if we fail [here](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/ua_client.py#L412) - callback won't accidentally be set up

oroulet:
In a way, it sounds safer to have a timeout. Waiting forever for server is not a very good idea either. But hopefully there is some timeout somewhere, in socket read for example

djipko:
>  But hopefully there is some timeout somewhere, in socket read for example

Yep it's [here](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/ua_client.py#L124) passed in from [here](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/ua_client.py#L229) but as mentioned above - the difference is that raising a timeout in `create_subscription` in the opcua-asyncio implementation will tear down the whole loop. In python-opcua it just raises in the main thread while the `_receiver` thread could still receive the response and register callbacks guaranteed to deadlock at that point.

If you still think it's less risky to do a timeout (count) on the while loop - I'm happy to do it, but still feel this is more consistent fix (albeit just slightly :) ) .

oroulet:
@djipko Thanks for the analysis. I am not saying that timeout is better, I am just trying to understand the issue and fix opcua-asyncio at the same time. You have good arguments so I think I will merge your Pr as-is.

djipko:
Many thanks! asyncio-opcua is safe from this one I think (not tested it though).

djipko:
Heya :) a gentle ping on this. Thanks again for taking a look!

PR #835, opened 2019-05-06
Convert StatusCode value to name and doc on-the-fly

helmut-jacob:
Otherwise StatusCode objects will carry a wrong name and doc after
deserialization.

Also add a test for proper resolution of name/doc from value.


helmut-jacob:
Python 2.7 build failure seems unrelated to me ...

oroulet:
make sense!

PR #838, opened 2019-05-10
Set OperationLimits values correctly

### LINKED ISSUE #815: OperationLimits (MaxNodesPerRead, etc) not set
Hi,

all Nodes under `OperationLimits` (`i=11704`), like `MaxNodesPerRead`, `MaxNodesPerBrowse`, etc don’t have any values set. This makes some OPC-UA-Clints, such as KepwareServerEX (as a client) break.

I’d like to prepare a PR for this, but would like to discuss how to set them beforehand: I don’t think we have a problem with very large arrays or structures on python-side, but that could lead to problems with timeouts, or clients not coping if they actually blindly fill up these limits. So the question is: Do we set them conservatively, or liberally?

Thanks!
-Dario

Nebukadneza:
Some client systems, such as KepWare-Server, query the OperationLimits
and adapt their behavior accordingly. Unfortunately, if the
OperationLimits are missing, connecting will fail.

To redeem these cases, let’s set liberal but managable
OperationLimit-defaults.

closes #815

### DIFF
--- a/opcua/server/internal_server.py
+++ b/opcua/server/internal_server.py
@@ -136,6 +136,29 @@ class InternalServer(object):
 
         results = self.isession.add_references([it, it2])
 
+        params = ua.WriteParameters()
+        for nodeid in (ua.ObjectIds.Server_ServerCapabilities_OperationLimits_MaxNodesPerRead,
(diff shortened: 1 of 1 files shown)


oroulet:
thanks

PR #844, opened 2019-05-22
opcua.Client: make session timeout and secure channel timeout configrable

Phlogistique:
The default timeouts are one hour, which is too long in some situations.

For example when developing a frequently crashing client program, it is easy to reach the limit of the number of sessions allowed by the server. Configuring a shorter timeout works around this problem.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -97,6 +97,17 @@ class Client(object):
         :param timeout:
             Each request sent to the server expects an answer within this
             time. The timeout is specified in seconds.
+
+        Some other client parameters can be changed by setting
+        attributes on the constructed object:
+
+        secure_channel_timeout
+            Timeout for the secure channel, specified in milliseconds.
+
+        session_timeout
+            Timeout for the session, specified in milliseconds.
+
+        See the source code for the exhaustive list.
         """
         self.logger = logging.getLogger(__name__)
         self.server_url = urlparse(url)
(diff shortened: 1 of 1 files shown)


opcua/client/client.py
```diff
@@ -86,7 +86,7 @@ class Client(object):
     the raw OPC-UA services interface.
     """
 
-    def __init__(self, url, timeout=4):
+    def __init__(self, url, timeout=4, secure_channel_timeout=3600000, session_timeout=3600000):
```
oroulet: it is not because the parameters are not passed to constructor that you cannot set them, they are not hidden. They were simply not taken as arg to constructor becasue they are not so common. but any more serious application should change a bunch of them like product_uri, certificates, etc....

oroulet: otherwise all changes are fine

Phlogistique:
Should I leave out the parameters and add the documentation elsewhere?

opcua/client/client.py
Phlogistique: Should I leave out the parameters and add the documentation elsewhere?

oroulet: I tend to think we should not have these parameters as arguments. They are not more important than the others

Phlogistique: Where should I document them then?

oroulet: Inside constructor string is fine. Just write that they need to be set up after constructing the object

oroulet:
What is the status here? @Phlogistique

Phlogistique:
Here's an updated branch with the constructor parameters left out.

oroulet:
thanks

PR #849, opened 2019-06-04
Fix localtime field of events

Nebukadneza:
Before, the localtime field was set to a wrong datatype (DateTime
instead of TimeZoneDataType), and didn’t contain the local timezone
offset, but the current localtime.


Nebukadneza:
Hm, dang — with the forcepush, i fixed the python2 test-breakage, but … how do i re-run travis as the PR author, not the repo maintainer?

Issue #854, opened 2019-06-30
inconsistent user and password decode UTF-8 with and without secured endpoints


biancode:
There is a tiny inconsistent user and password behaviour decode UTF-8 with and without secured endpoints. On using secured connection it decodes the password from a client on check_user_token in the file user manager. With None, it doesn't decode and the user has to handle to different use cases after in the check_user routine.

It would be easier for the user to get a consistent decoded password on every use case None, Sign, Sign&Encrypt, please.

biancode:
user has to do this in check_user

```
    try:
        passwordUTF8 = password.decode('utf-8')
    except:
        passwordUTF8 = password
```

PR #855, opened 2019-06-30
decode password on None

biancode:
fix on issue #854

### DIFF
--- a/opcua/server/user_manager.py
+++ b/opcua/server/user_manager.py
@@ -72,6 +72,11 @@ class UserManager(object):
             except Exception as exp:
                 self.logger.warning("Unable to decrypt password")
                 return False
+        else:
+            try:
+                passwd = passwd.decode('utf-8')
+            except:
+                self.logger.warning("Unable to decode password")
 
         # call user_manager
         return self.user_manager(isession, userName, passwd)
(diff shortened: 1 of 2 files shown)


opcua/server/user_manager.py
```diff
@@ -72,6 +72,11 @@ def check_user_token(self, isession, token):
             except Exception as exp:
                 self.logger.warning("Unable to decrypt password")
                 return False
+        else:
+            try:
+                passwd = passwd.decode('utf-8')
+            except:
+                self.logger.warning("Unable to decode password")
```
oroulet: Is a warning enough? Aren't we fucked and should return am error?

biancode: I tested this with the UAExpert and the password was a bytestring on None, too.

biancode: If the password is not to decode, then it is to handle by ceck_user fro me. So it should put the password thru and user has to handle errors. What do you think?

oroulet: Ok. Make sense

biancode: Perfect. The test in Python2.7 crashed, but Python3 is done. Maybe it is to fix Travis by run the 2.7 tests again.

PR #869, opened 2019-07-26
Fix import warning

libra146:
Fix warning when importing 'collections'.
Warning will only appear in Python 3.7.
Please see:[https://docs.python.org/3/library/collections.html](https://docs.python.org/3/library/collections.html)

> Moved Collections Abstract Base Classes to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.7. Subsequently, they will be removed entirely.

### DIFF
--- a/opcua/common/callback.py
+++ b/opcua/common/callback.py
@@ -3,7 +3,7 @@
 server side implementation of callback event 
 """
 
-from collections import OrderedDict
+from collections.abc import OrderedDict
 from enum import Enum
 
 class CallbackType(Enum):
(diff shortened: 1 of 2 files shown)


oroulet:
does not seem to work for any python version.. where did you test?

libra146:
> 似乎不适用于任何python版本..你在哪里测试？

Oh, sorry, it seems that the problem is not on opcua. I am using pymssql and it may be a problem.

libra146:
> does not seem to work for any python version.. where did you test?

Oh, sorry, it seems that the problem is not on opcua. I am using pymssql and it may be a problem.

PR #888, opened 2019-09-09
Bugfix: Node.set_value() does not work with enumerations

MarcoH82:
If a client tries to set a enum variable without explicitly force the data type it becomes implicit converted to a ua.Variant.Int64. The server expects an Int32 for an enum value. This is why set_value() crashes if you try to set an enum value using a integer literal like "set_value(2)".

The enumeration seems to need a special treadment at this point. See ua_utils.py def data_type_to_variant_type().

If no data type was given in method set_value(), this change checks the data type of the node object and automatically converts the value argument to the correct target data type of the server.

### DIFF
--- a/opcua/common/node.py
+++ b/opcua/common/node.py
@@ -209,10 +209,18 @@ class Node(object):
         elif isinstance(value, ua.Variant):
             datavalue = ua.DataValue(value)
             datavalue.SourceTimestamp = datetime.utcnow()
+        elif varianttype is None:
+            # Use the data type of the node (self) if no data type is given. 
+            varianttype = self.get_data_type_as_variant_type()
+            datavalue = ua.DataValue(ua.Variant(value, varianttype))
+            datavalue.SourceTimestamp = datetime.utcnow()
         else:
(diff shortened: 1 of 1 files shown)


MarcoH82:
Can anyone please watch this PL? My client is doing fine with this change. The unit tests have also gone through successfully. However, in my opinion, a test for setting an enumerarion unsing an integer literal lacks. Can someone help with implementing a test for the change?

MarcoH82:
The tests that fail here also fail with the main branch without any changes. Does anyone have an explanation for this?

oroulet:
Thanks for the PR but I am a bit afraid of that one. you are introusing a second network operation on a write call...

MarcoH82:
> Thanks for the PR but I am a bit afraid of that one. you are introusing a second network operation on a write call...

Oh dear, I did not know that. How could this be done without a second network operation?

oroulet:
Maybe it is impossible.... But maybe we should write numbers with unspecified data types as 4 bytes by default. 32 seems to be the default in opcua...

PR #890, opened 2019-09-17
Fix WhereClauseEvauluator

einarmo:
There were two major issues. First, _eval_el was called instead of _eval_op in recursive
calls, which would always fail.

Secondly, there was a misplaced line of code in "and" evaluation in _eval_el, which would
cause it to always fail.

### DIFF
--- a/opcua/server/internal_subscription.py
+++ b/opcua/server/internal_subscription.py
@@ -423,30 +423,29 @@ class WhereClauseEvaluator(object):
         # ops = [self._eval_op(op, event) for op in el.FilterOperands]
         ops = el.FilterOperands  # just to make code more readable
         if el.FilterOperator == ua.FilterOperator.Equals:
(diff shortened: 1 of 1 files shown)


oroulet:
great patch! how did you found that issue? ideally we should have tests for that but nobody seems to use these very often..

einarmo:
I'm writing a tool to extract the contents of opcua servers using the reference .NET sdk, the client side sdk is relatively easy to use, and works well when we need it to be fast, however the server side sdk takes much more work, so we're using python-opcua to create test servers. This was just one recent feature that broke.

That does mean that this is tested with the official client. The clientside code is more or less based on stuff found [here](https://github.com/OPCFoundation/UA-.NETStandard/blob/master/SampleApplications/Workshop/AlarmCondition/Client/FilterDefinition.cs)

PR #893, opened 2019-09-27
Handle lists of multiple types in ua.Variant._guess_type

alexanian:
Thank you for taking the time to review my PR!

### Summary
* When guessing the type of a list, look at the type of every element, not just the first one.
* Raise an error if there are multiple types, unless the types are `int` and `float`.
* Lists of `int` and `float` are treated as `float`.
* Tests for the above.

### Motivation
While working with some device drivers, I was seeing errors with long stack traces and the message:

    struct.error: required argument is not an integer

This was caused by writing lists of coordinates in which not every element has decimal precision. I found it perplexing that I wouldn't see the error with:

    [-110.1,0,64.25,180,0,180]
but would with:

    [-110,0,64.25,180,0,180]

### Additional Notes
This is converting `[1, 2.0, 3]` to `[1.0, 2.0, 3.0]` with just a `debug` to show for it. That kind of conversion already happens for lists like `[1.0, 2, 3]` (which would become `[1.0, 2.0, 3.0]`) but I can see the case for cutting that `elif` and just raising an exception.

### DIFF
--- a/opcua/ua/uatypes.py
+++ b/opcua/ua/uatypes.py
@@ -770,12 +770,28 @@ class Variant(FrozenClass):
         return not self.__eq__(other)
 
     def _guess_type(self, val):
+
         if isinstance(val, (list, tuple)):
             error_val = val
-        while isinstance(val, (list, tuple)):
-            if len(val) == 0:
-                raise UaError("could not guess UA type of variable {0}".format(error_val))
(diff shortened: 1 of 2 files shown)


zerox1212:
Debug is probably fine. Looks like a good fix for @oroulet to merge.

Issue #905, opened 2019-10-15
UaClient.close_session(): delay on closed socket.


brubbel:
When an OPC connection is closed at server side, calling [uaClient.close_session()](https://github.com/FreeOpcUa/python-opcua/blob/83fb9ea7e2881af8899fab117df0ea1bf8789000/opcua/client/ua_client.py#L284) still tries to send a ua.CloseSessionRequest to the server, which of course results in a [timeout](https://github.com/FreeOpcUa/python-opcua/blob/83fb9ea7e2881af8899fab117df0ea1bf8789000/opcua/client/ua_client.py#L70) while waiting for the server's response.
This prevents a fast reconnect by the client.
Of course one can use a smaller timeout, but this may result in intermittent timeouts on slow connections.

Better is to also close the [SecureConnection](https://github.com/FreeOpcUa/python-opcua/blob/83fb9ea7e2881af8899fab117df0ea1bf8789000/opcua/client/ua_client.py#L34) as soon as the server closes the socket, as (1) a closed socket will be never reopened and (2) the current session (BadSessionIdInvalid) becomes invalid on a new connection.

PR #906, opened 2019-10-15
Don't wait on close_session() when underlying socket is closed

brubbel:
cfr. https://github.com/FreeOpcUa/python-opcua/issues/905

### DIFF
--- a/opcua/client/ua_client.py
+++ b/opcua/client/ua_client.py
@@ -100,6 +100,7 @@ class UASocketClient(object):
                 break
             except UaError:
                 self.logger.exception("Protocol Error")
+        self._cancel_all_callbacks()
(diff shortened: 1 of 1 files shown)


brubbel:
Update https://github.com/FreeOpcUa/python-opcua/pull/906/commits/e269e429638802c82597a7e90f03367e245cc5e8:
Don't wait for future timeout when underlying socket is closed.

oroulet:
Finished? No idea if this is correct but your explanation makes sense

brubbel:
Not yet, there will be some additional changes later on.

oroulet:
Change name to start with wip: then. So everybody knows it is work in progress

brubbel:
Ready

PR #913, opened 2019-11-22
Get rid of FrozenClass and use __slots__ instead

helmut-jacob:
Instead of FrozenClass we can easily enforce a statis attribute
structure by using __slots__.

This has a second advantage: working with huge numbers of OPCUA
data structures is less memory intensive and faster improving the
overall speed of python-opcua. This should speed up history reads
where tons of DataValues are created and serialized.

Example measurement using python 3.6:

	import opcua
	import timeit
	def run():
		dvs=[]
		for i in range(1, 100000):
			dvs.append(opcua.ua.DataValue(i))
	timeit.timeit(run, number=10)

Result before: 6.1s
Result after:  5.2s

That's a ~15% increase.

Measurements on history read with the SQLite history backend
is ~10% faster with this change.

### DIFF
--- a/opcua/common/connection.py
+++ b/opcua/common/connection.py
@@ -9,11 +9,11 @@ from opcua import ua
 logger = logging.getLogger('opcua.uaprotocol')
 
 
-class MessageChunk(ua.FrozenClass):
+class MessageChunk(object):
     """
     Message Chunk, as described in OPC UA specs Part 6, 6.7.2.
     """
-
+    __slots__ = ['MessageHeader', 'SecurityHeader', 'SequenceHeader', 'Body', 'security_policy']
     def __init__(self, security_policy, body=b'', msg_type=ua.MessageType.SecureMessage, chunk_type=ua.ChunkType.Single):
         self.MessageHeader = ua.Header(msg_type, chunk_type)
         if msg_type in (ua.MessageType.SecureMessage, ua.MessageType.SecureClose):
--- a/opcua/common/copy_node.py
+++ b/opcua/common/copy_node.py
@@ -66,7 +66,8 @@ def _rdesc_from_node(parent, node):
 
 
 def _read_and_copy_attrs(node_type, struct, addnode):
-    names = [name for name in struct.__dict__.keys() if not name.startswith("_") and name not in ("BodyLength", "TypeId", "SpecifiedAttributes", "Encoding", "IsAbstract", "EventNotifier")]
+    keys = struct.__dict__.keys() if struct.__dict__  else struct.__slots__
+    names = [name for name in keys if not name.startswith("_") and name not in ("BodyLength", "TypeId", "SpecifiedAttributes", "Encoding", "IsAbstract", "EventNotifier")]
     attrs = [getattr(ua.AttributeIds, name) for name in names]            
     for name in names:
         results = node_type.get_attributes(attrs)
--- a/opcua/ua/uaprotocol_auto.py
+++ b/opcua/ua/uaprotocol_auto.py
@@ -1049,15 +1049,18 @@ class ExceptionDeviationFormat(IntEnum):
     Unknown = 4
 
 
-class DataTypeDefinition(FrozenClass):
+class DataTypeDefinition(object):
     '''
     '''
 
     ua_types = [
                ]
 
+    __slots__ = [
+               ]
+
     def __init__(self):
-        self._freeze = True
+        pass
 
     def __str__(self):
         return 'DataTypeDefinition(' +  + ')'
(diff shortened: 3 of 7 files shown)


helmut-jacob:
@oroulet I would assume the same change makes sense for asyncua as well ...

oroulet:
How can this improve perfrmance?. I thought FrozenObject was not used, unless we had an environement variable set. Also I though we tested slots and could not make it work... but if you tell me it works, I am fine
and yes if it works here, we should port that to opcua.asyncio

helmut-jacob:
> How can this improve perfrmance?. I thought FrozenObject was not used, unless we had an environement variable set.

Correct, by default we are not using FrozenObject and my tests have been without the magic environment variable.

The speedup happens due to the usage of slots where less memory per object needs to be allocated and attribute accesses are a bit faster. Since a history read can generate thousands of DataValues containing Variant and StatusCode objects the time is saved at object creation time.

> Also I though we tested slots and could not make it work... but if you tell me it works, I am fine

Ok, do you remeber by any chance what issues you had with slots? The default behavior of the code changed with slots to _never_ allow un-declared attributes (as if we are always running with FrozenClass) and as a bonus we're getting the improved object creation speed.

> and yes if it works here, we should port that to opcua.asyncio

oroulet:
A quick search for slots led to this: https://github.com/FreeOpcUa/python-opcua/pull/76

oroulet:
I was obvisouly against slots at that time. but maybe something has changed...

oroulet:
BtwWhat about dataclass don't they have something similar?

helmut-jacob:
> I was obvisouly against slots at that time. but maybe something has changed...

I think your original complaint that you'd have to add the __slots__ property to every class in uaprotocol_hand.py is still valid. This is a drawback of using slots. On the other hand we are getting rid of the self._freeze. For uaprotocol_auto.py it shouldn't matter much anyway.

> BtwWhat about dataclass don't they have something similar?

This wouldn't yield the same benefits as we get with slots:
1. "FrozenClass" behavior by default
2. Improved performance when dealing with lots of objects (aka historical reads)

I'd prefer if we take in slots if you don't have a strong objection.

oroulet:
yes for autogenerated class it does not matter.

But one more thing: I have though for a long time that we should use dataclass, typing and __annotation__ for all these structures instead of our custom uatypes member. Something like

```
class DecimalDataType(FrozenClass):
    """
    :ivar Scale:
    :vartype Scale: Int16
    :ivar Value:
    :vartype Value: ByteString
    """

    ua_types = [
        ('Scale', 'Int16'),
        ('Value', 'ByteString'),
               ]

    def __init__(self):
        self.Scale = 0
        self.Value = None
        self._freeze = True

    def __str__(self):
        return f'DecimalDataType(Scale:{self.Scale}, Value:{self.Value})'

    __repr__ = __str__

```

Would become:

```
@dataclass
class DecimalDataType:

    def __init__(self):
        self.Scale: Int16 = 0
        self.Value:ByteString = None
```

That is quite a lot of saving....

then in the serializing code we just loop through `DecimaDataType.__annotations__`

But is this compatible with __slots__?

helmut-jacob:
Good point, I don't think data classes and slots can be used at the same time but let me look into this a bit more first ...

helmut-jacob:
There's a package available called dataslots https://pypi.org/project/dataslots/ :-), I did a quick test with the following example code:
`
@with_slots 
@dataclass 
class XXX: 
  x: opcua.ua.ua_binary.Primitives.Byte 
  y: opcua.ua.ua_binary.Primitives.Int32
`

* This will give us a class with slots -> performance increase
* We could look over the __annotations__ and directly call .pack() on them

=> Looks promising, let me dig deeper into this

oroulet:
this looks really good in fact :+1: @helmut-jacob

oroulet:
but implementing it is probably opcua-asyncio only due to typing etc...

helmut-jacob:
You mean because python-opcua should still work with python 2.7, right? While opcua-asyncio requires python 3.6 or later.

oroulet:
Yes. I am also considering to say >3.7

oroulet:
did you have a look @helmut-jacob ?
I had a short look at it and I am not really sure how us these dataclass

`
@dataclass_slots
@dataclass
class MyUaStruct:
    a: UInt32 = UInt32()
    b: String = String("toto is back")
`
does not look good. It is going to be heavy to use. I would prefere something like:

`
@dataclass_slots
@dataclass
class MyUaStruct:
    a: UInt32 = 0
    b: String = "totois back"
`
but I am afraid the code over makes tools like mypy cry. It really looks like a misuse of typing (although a correct use of dataclass as far as I understand)

helmut-jacob:
I didn't have time to further look into this yet.

helmut-jacob:
I did not follow up on this one. So closing for now but I think we have reasonable options to go forward ...

PR #923, opened 2019-12-18
Allow customizing buildinfo

helmut-jacob:
This allows consumers of the library to set buildinformation.
Additionally this fixes buildinformation representation in individual
nodes (like ua.ObjectIds.Server_ServerStatus_BuildInfo_ProductUri).
Before the individual buildinfo nodes haven't been populated. Now,
they are kept in sync with the buildinfo struct.


helmut-jacob:
Thanks @oroulet are you going to foward port this to asyncua as well?

oroulet:
if I get time... great if you can do it..

PR #930, opened 2019-12-31
removed unused code

brubbel:


oroulet:
that one is strange.. it really looks like code for some service method. it maight be missing some connection... to some ua call

brubbel:
Currently, the uaprocessor will return a 'BadServiceUnsupported' for a HistoryUpdate request, while this unused code will just return 'BadNotWritable'.

The HistoryUpdate service is defined here (Update, insert and remove historical values)
https://reference.opcfoundation.org/v104/Core/docs/Part4/5.10.5/
But that is not implemented.

The FreeOpcua server also reports these capabilities as 'Null'. Maybe they should be set to a correct value when setting the server state nodes.
e.g. HistoryServerCapabilities->UpdateDataCapability
https://documentation.unified-automation.com/uasdkdotnet/2.6.0/html/Doc_OpcUa_HistoricalAccess_ObjectTypes.html

oroulet:
yes nobody has implemented these capabilities...

PR #932, opened 2020-01-02
readability: parameter is direction i.o. BrowseDescription

brubbel:
https://github.com/FreeOpcUa/python-opcua/blob/4af9cda09afee5fb636ab6e43df1f588e88becbc/opcua/server/address_space.py#L96

vs.

https://github.com/FreeOpcUa/python-opcua/blob/d9a9f42ddf17f4fc89e9aed8350c9d4046f3f986/opcua/server/address_space.py#L134-L141

desc -> direction


oroulet:
always better to not pass too much to functions anyway. so good

PR #945, opened 2020-01-22
[OpcuaClient][Backport async] Support of ClientSignature based on Sec…

ZuZuD:
Let's backport this and take advantage of what's in this [commit](https://github.com/FreeOpcUa/python-opcua/commit/9834d4966074c355329cba0e4fe0e2c3389fae9e)

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -429,7 +429,15 @@ class Client(object):
             challenge += self.security_policy.server_certificate
         if self._server_nonce is not None:
             challenge += self._server_nonce
-        params.ClientSignature.Algorithm = "http://www.w3.org/2000/09/xmldsig#rsa-sha1"
+
+        if self.security_policy.AsymmetricSignatureURI:
+            params.ClientSignature.Algorithm = (
+                self.security_policy.AsymmetricSignatureURI
+            )
+        else:
+            params.ClientSignature.Algorithm = (
+                "http://www.w3.org/2000/09/xmldsig#rsa-sha1"
+            )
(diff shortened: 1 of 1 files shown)


opcua/client/client.py
```diff
@@ -429,7 +429,15 @@ def activate_session(self, username=None, password=None, certificate=None):
...
             challenge += self._server_nonce
-        params.ClientSignature.Algorithm = "http://www.w3.org/2000/09/xmldsig#rsa-sha1"
+
+        if self.security_policy.AsymmetricSignatureURI:
+            params.ClientSignature.Algorithm = (
+                self.security_policy.AsymmetricSignatureURI
+            )
+        else:
+            params.ClientSignature.Algorithm = (
+                "http://www.w3.org/2000/09/xmldsig#rsa-sha1"
```
oroulet: Isn't that string defined somewhere so we do not need to hardcode it here?

ZuZuD: I copied on purpose what's in [opcua-async](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/client.py#L407-L410) but that's updated with the last commit.

Few client/server tests by the way:
[NoSecurity](https://gist.github.com/ZuZuD/a65ece60c1b54003e9b2a7d5520ce091)
[Basic256Sha256_SignAndEncrypt](https://gist.github.com/ZuZuD/dd64116a885fb0260573dc90b51c9cb9)

oroulet:
thanks

Issue #963, opened 2020-02-18
Calling a method with no arguments causes a TypeError


craigh92:
**Error**

```
press any key to exitERROR:opcua.server.address_space:Error executing method call CallMethodRequest(ObjectId:FourByteNodeId(ns=2;i=26), MethodId:FourByteNodeId(ns=2;i=27), InputArguments:None), an exception was raised: 
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/opcua/server/address_space.py", line 466, in _call
    result = node.call(method.ObjectId, *method.InputArguments)
TypeError: wrapper() argument after * must be an iterable, not NoneType
```

**Cause**

This is because line [466](https://github.com/FreeOpcUa/python-opcua/blob/1c73bb8ab9c03987a4900c7515c25acf1d6d8597/opcua/server/address_space.py#L466) expects an iterable:

```python
result = node.call(method.ObjectId, *method.InputArguments)
```

**Fix**

This can be fixed by passing in an empty array if there are no input arguments:

```python
if method.InputArguments is None:
    method.InputArguments = []             
result = node.call(method.ObjectId, *method.InputArguments)
```

```

craigh92:
Fixed in Open Merge Request #964

PR #964, opened 2020-02-18
TypeError on method call with no inputArguments. Added case to handle this.

craigh92:
Closes Issue [963](https://github.com/FreeOpcUa/python-opcua/issues/963)

### DIFF
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -463,6 +463,8 @@ class MethodService(object):
                 res.StatusCode = ua.StatusCode(ua.StatusCodes.BadNothingToDo)
             else:
                 try:
+                    if method.InputArguments is None:
+                        method.InputArguments = [] 
                     result = node.call(method.ObjectId, *method.InputArguments)
                     if isinstance(result, ua.CallMethodResult):
                         res = result


tests/tests_subscriptions.py
```diff
@@ -128,7 +128,7 @@ def test_subscription_count_list(self):
             val = copy(val)  # we do not want to modify object in our db, we need a copy in order to generate event
             val.append(i)
             var.set_value(copy(val))
-        time.sleep(0.2)  # let last event arrive
```
craigh92: This was too short and causing the test to occasionally fail. I have increased it to 0.3.

opcua/server/address_space.py
```diff
@@ -463,6 +463,8 @@ def _call(self, method):
                 res.StatusCode = ua.StatusCode(ua.StatusCodes.BadNothingToDo)
             else:
                 try:
+                    if method.InputArguments is None:
+                        method.InputArguments = []
```
oroulet: these lines will never fail. No need to have them under try/except

oroulet:
Probably fine, but canyou add  a test for that new behaviour? I am curious to know when it happend. I thought this was well supported

opcua/server/address_space.py
craigh92: These lines will never fail, but the actual method call could. I have put this check in the same scope as the method call for readability. Would you prefer it outside of the try statement?

oroulet: OK I just looked at that code. In fact we should only have try/except  around the node.call line. Can you move everything else out?

oroulet:
@craigh92 any news?

PR #968, opened 2020-02-24
[Reconciliate MI] Add tests and fix nits

ZuZuD:
Add unit test for the client / subscription reconciliate methods.

Also fix some nits raised here: https://github.com/FreeOpcUa/python-opcua/commit/de6a5185bcec893136d6e493b87753779657eea2#commitcomment-37372079


oroulet:
I do not uinderstand the mock part, but that should be OK

PR #971, opened 2020-02-25
Introduces 'is_connected' client call. Fixes #521

### LINKED ISSUE #521: get client connection state
I did not find a possibility to check that a client connection has been successfully established like `client.is_connected`.

OptrixAU:
In applications involving subscribing to data and waiting for data to stream in, there's no clear way to check if the client has been disconnected (for example, by shutting down the OPC-UA server).

This function isn't intended to check if a connection was initially successful (although theoretically it would do that too) - the intention is to check to see if the connection has been _dropped_.

Usage Pattern:
In my application, after 10 seconds with no new incoming data from the subscription, I check to see if the connection is alive. If it's still alive, it's just a lull in data, but if it's dead, we need to disconnect, reconnect and resubscribe to keep up-to-date with new data.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -652,3 +652,9 @@ class Client(object):
         for result in results:
             result.check()
 
+    def is_connected(self):
+        """
+        Get OPC-UA client connection status.
+        """
+        return self.uaclient._uasocket._thread.isAlive()
+


oroulet:
checking that the thread is alive does not tell us of communication is OK or not. The only real way we have to know if we are connected is to try to read a node. In fact reading the server state node is the best as the server mught be on but in a bad state. This is done in KeepAlive class.
So maybe you could write something like*
        server_state = self.client.get_node(ua.FourByteNodeId(ua.ObjectIds.Server_ServerStatus_State))
and then check the value is correct inside that a method called is_connection_ok() or similar

luisfmelo:
I'm picking this up

oroulet:
Did open other PR? Icannot see them now

OptrixAU:
I'd say it will be sent tomorrow.



Sent from my Samsung Galaxy smartphone.



-------- Original message --------

Issue #972, opened 2020-02-27
Bug if NodeId IdentifierType is GUID


Michael-List:
I would like to add a variable with GUID as NodeIdType:
```
my_folder.add_variable(ua.NodeId(identifier=UUID('777b9d07-9e48-4a2f-9084-7ceae7043c8a').bytes, namespaceidx=ns, nodeidtype=NodeIdType.Guid), 'guid Test', 33)
```
But I'm getting the error _AttributeError: 'bytes' object has no attribute 'time_low'_

The UUID object has the attribute time_low. I tried to create the NodeId with an UUID identifier:
```
int_identifier_folder.add_variable(ua.NodeId(identifier=UUID('777b9d07-9e48-4a2f-9084-7ceae7043c8a'), namespaceidx=ns, nodeidtype=NodeIdType.Guid), 'guid Test', 33)
```
But here im getting the error _TypeError: expected string or bytes-like object_
Which is from this file/line:
```
opcua\ua\uatypes.py", line 340, in has_null_identifier
if self.NodeIdType == NodeIdType.Guid and re.match(b'0.', self.Identifier):
```

Can someone tell me what I'm doing wrong here? Or is there a bug in the re.match?

If I change the above line from uatypes.py to this it works:
```
if self.NodeIdType == NodeIdType.Guid and re.match(b'0.', self.Identifier.bytes):
```

zerox1212:
I don't know if anyone has used UUID for NodeID. There is probably a bug here. I looked at the tests and there is a test for GUID node id, but it's being checked a different way. The test creates a `ua.GuidNodeId()` object and encode/decodes it with the binary protocol.

Michael-List:
I think I fixed it. Pull request is open

Michael-List:
Resolved in merge request #975

PR #975, opened 2020-03-13
Fix error with guid as NodeIdType

Michael-List:
Fix error with guid as NodeIdType. For more information see issue #972

### DIFF
--- a/examples/server-example.py
+++ b/examples/server-example.py
@@ -1,3 +1,4 @@
+import uuid
 from threading import Thread
 import copy
 import logging
@@ -5,6 +6,9 @@ from datetime import datetime
 import time
 from math import sin
 import sys
+
+from opcua.ua import NodeId, NodeIdType
+
 sys.path.insert(0, "..")
 
 try:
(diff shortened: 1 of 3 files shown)


oroulet:
That seems to make sense. Is there any case where someone will want to provide bytes and not uuid object?

Michael-List:
If someone would like to provide bytes we have to convert it to an uuid afterwards. If it's an byte we run into an error because we cannot pack it like below.

ua_binary.py:
```
class _Guid(object):
    @staticmethod
    def pack(guid):
        # convert python UUID 6 field format to OPC UA 4 field format
        f1 = Primitives.UInt32.pack(guid.time_low)
        f2 = Primitives.UInt16.pack(guid.time_mid)
        f3 = Primitives.UInt16.pack(guid.time_hi_version)
        f4a = Primitives.Byte.pack(guid.clock_seq_hi_variant)
        f4b = Primitives.Byte.pack(guid.clock_seq_low)
        f4c = struct.pack('>Q', guid.node)[2:8]  # no primitive .pack available for 6 byte int
        f4 = f4a + f4b + f4c
        # concat byte fields
        b = f1 + f2 + f3 + f4

        return b
```

zerox1212:
I think I wrote the binary protocol for GUID. We should not use a bytes object. If someone supplies bytes we need to convert it to UUID as mentioned above. However I would recommend people use the python UUID whenever possible, so I do not think it's worth supporting bytes at this time.

oroulet:
Ok

PR #982, opened 2020-03-27
FEATURE: Allow to save to file that doesn't exist

kozachynskyi:
If file in path doesn't exist, it won't be created. "t" is not needed as a default option.


oroulet:
I did not even know about "t"... not sure why it was there...

PR #984, opened 2020-03-27
FIX: Missing "cryptography" warning moved to class

kozachynskyi:
When someone imports opcua, he/she sees 3 same warnings, even in
opcua is not used. I see the reasoning behind this feature, so I
propose to move it in classes. 

If you have another proposal, let me know, I can change the commit

### DIFF
--- a/opcua/server/user_manager.py
+++ b/opcua/server/user_manager.py
@@ -7,11 +7,13 @@ use_crypto = True
 try:
     from opcua.crypto import uacrypto
 except ImportError:
-    logging.getLogger(__name__).warning("cryptography is not installed, use of crypto disabled")
     use_crypto = False
 
 
 class UserManager(object):
+
+    if use_crypto is False:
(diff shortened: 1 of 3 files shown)


opcua/server/user_manager.py
```diff
@@ -7,12 +7,14 @@
...
 except ImportError:
-    logging.getLogger(__name__).warning("cryptography is not installed, use of crypto disabled")
     use_crypto = False
 
 
 class UserManager(object):
 
+    if use_crypto is False:
+        logging.getLogger(__name__).warning("cryptography is not installed, use of crypto disabled")
+
```
oroulet: if someone uses user_manager, they will also use Server so no need to print an extra warning in that file

oroulet:
fine. just remove all warnings in user_manager.py. they are not needed

kozachynskyi:
Done

Issue #811, opened 2019-03-20
[Client] activate session with security policy Basic256Sha256 fails


mihainstr:
https://github.com/FreeOpcUa/python-opcua/blob/cdcaed36b772abf0ea1dffe42b2d8cdaada8f5e1/opcua/client/client.py#L414

**ISSUE:**

Attempting a CLIENT connect with the security policy set to Basic256Sha256 will fail when trying to activate the session. 

The issue is related to the incorrect ClientSignature.Algorithm used as a parameter for the activate session request, the current implementation uses "http://www.w3.org/2000/09/xmldsig#rsa-sha1" as the default algorithm which corresponds to  "http://opcfoundation.org/UA/SecurityPolicy/#Basic256". 

**FIX:**
Changing the ClientSignature.Algorithm to "http://www.w3.org/2000/09/xmldsig#rsa-sha256" activates the session correctly when security policy is set to Basic256Sha256.

The permanent fix should ensure that the correct encryption algorithm is deduced based on the used security policy.

sixdiamants:
I faced the same issue running the python client against a [UA-.NETStandard ](https://github.com/OPCFoundation/UA-.NETStandard)server.  The server logged this:

> 26-7-2019 09:51:09.930 Server - SESSION ACTIVATE failed. Unexpected signature algorithm for Basic256Sha256/Aes128_Sha256_RsaOaep: http://www.w3.org/2000/09/xmldsig#rsa-sha1
Expected signature algorithm: http://www.w3.org/2001/04/xmldsig-more#rsa-sha256  

as reported [here](https://opcfoundation.org/forum/opc-ua-implementation-stacks-tools-and-samples/sha2-signed-certificate-is-interpreted-as-sha1-signed-and-wrongly-rejected/).

Applying the fix mentioned above by @mihainstr still returned an unexpected signature algorithm. Only when I replaced 
 >  params.ClientSignature.Algorithm = "http://www.w3.org/2000/09/xmldsig#rsa-sha1"   

by the URI suggested by the server log
 
>params.ClientSignature.Algorithm =  "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"  

Did the server accept the connection. In other words, the URI to the algorithm for the signature method `http://www.w3.org/2000/09/xmldsig#rsa-sha256` didn't work for me.

alexanderpetrenz:
Is it desired to make this algorithm string configurable?

oroulet:
[...]

PR #994, opened 2020-04-14
Adjusting Client Signature Algorithm (FreeOpcUa/python-opcua#811)

alexanderpetrenz:
I adjusted the string for die Client Signature Algorithm. As SHA1 is considered to be insecure, servers may reject encrypted connections when SHA1 is offered. Changing the algorithm to SHA256 seems to fix this issue.


alexanderpetrenz:
This is also discussed in FreeOpcUa/python-opcua#811

PR #996, opened 2020-04-16
Fix _find_element_in_node to cover all input arguments

helmut-jacob:
Otherwise _translate_browsepath_to_nodeid behaves incorrect in various
situations. For example if requesting a browse path translation for
['2:X', '2:X', '2:X'] it would have returned the first object instead of
the last one as inverse references haven't been treated properly.

Now we treat the inverse flag, the reference type and its subtypes when
translating a browse path.


helmut-jacob:
I did not yet check if this applies to async-ua as well but since I just stumbled over this one with the "old" lib ...

PR #1013, opened 2020-05-01
Update generate_certificate.sh

mstfldmr:
If subjectAltName is not defined while creating the certificate, you get error:

asyncua.ua.uaerrors._auto.BadCertificateUriInvalid: The URI specified in the ApplicationDescription does not match the URI in the certificate.(BadCertificateUriInvalid)

### DIFF
--- a/examples/generate_certificate.sh
+++ b/examples/generate_certificate.sh
@@ -1,2 +1,41 @@
+: '
+Generate your own x509v3 Certificate
+
+Step 1: Change ssl.conf (subjectAltname, country, organizationName, ...)
+
+ssl.conf:
+
+[ req ]
+default_bits = 2048
+default_md = sha256
+distinguished_name = subject
+req_extensions = req_ext
+x509_extensions = req_ext
+string_mask = utf8only
+prompt = no
+
+[ req_ext ]
+basicConstraints = CA:FALSE
+nsCertType = client, server
+keyUsage = nonRepudiation, digitalSignature, keyEncipherment, dataEncipherment, keyCertSign
+extendedKeyUsage= serverAuth, clientAuth
+nsComment = "OpenSSL Generated Certificat"
+subjectKeyIdentifier=hash
+authorityKeyIdentifier=keyid,issuer
+subjectAltName = URI:urn:opcua:python:server,IP: 127.0.0.1
+
+[ subject ]
+countryName = DE
+stateOrProvinceName = HE
+localityName = HE
+organizationName = AndreasHeine
+commonName = PythonOpcUaServer
+
+Step 2: openssl genrsa -out key.pem 2048
+Step 3: openssl req -x509 -days 365 -new -out certificate.pem -key key.pem -config ssl.conf
+
+this way is proved with Siemens OPC UA Client/Server!
+'
+
 openssl req -x509 -newkey rsa:2048 -keyout my_private_key.pem -out my_cert.pem -days 355 -nodes -addext "subjectAltName = URI:urn:example.org:FreeOpcUa:python-opcua"
 openssl x509 -outform der -in my_cert.pem -out my_cert.der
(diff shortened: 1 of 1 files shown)


AndreasHeine:
Generate your own x509v3 Certificate

Step 1: Change ssl.conf (subjectAltname, country, organizationName, ...)

ssl.conf
```
[ req ]
default_bits = 2048
default_md = sha256
distinguished_name = subject
req_extensions = req_ext
x509_extensions = req_ext
string_mask = utf8only
prompt = no

[ req_ext ]
basicConstraints = CA:FALSE
nsCertType = client, server
keyUsage = nonRepudiation, digitalSignature, keyEncipherment, dataEncipherment, keyCertSign
extendedKeyUsage= serverAuth, clientAuth
nsComment = "OpenSSL Generated Certificat"
subjectKeyIdentifier=hash
authorityKeyIdentifier=keyid,issuer
subjectAltName = URI:urn:opcua:python:server,IP: 127.0.0.1

[ subject ]
countryName = DE
stateOrProvinceName = HE
localityName = HE
organizationName = AndreasHeine
commonName = PythonOpcUaServer
```

Step 2: openssl genrsa -out key.pem 2048
Step 3: openssl req -x509 -days 365 -new -out certificate.pem -key key.pem -config ssl.conf

oroulet:
is my added commit OK? does it work?

mstfldmr:
to be compliant with 'examples/client-with-encryption.py', can you modify it as:

`subjectAltName = URI:urn:example.org:FreeOpcUa:python-opcua,IP: 127.0.0.1`

change file names and add the 4th step:

```
Step 2: openssl genrsa -out my_private_key.pem 2048
Step 3: openssl req -x509 -days 365 -new -out my_cert.pem -key my_private_key.pem -config ssl.conf
Step 4: openssl x509 -outform der -in my_cert.pem -out my_cert.der
```

AndreasHeine:
.der format is optional (my opinion: its not necessary because openssl only changes the fileextention not the file data)

why do you adding step 4? do i miss something?

oroulet:
I let you guys decide. I know nothing about certificates

AndreasHeine:
after some research:
[Source](https://support.ssl.com/Knowledgebase/Article/View/19/0/der-vs-crt-vs-cer-vs-pem-certificates-and-how-to-convert-them)
> 
> Certificates and Encodings
> 
> At its core an X.509 certificate is a digital document that has been encoded and/or digitally signed according to RFC 5280.
> 
> In fact, the term X.509 certificate usually refers to the IETF’s PKIX Certificate and CRL Profile of the X.509 v3 certificate standard, as specified in RFC 5280, commonly referred to as PKIX for Public Key Infrastructure (X.509).
> X509 File Extensions
> 
> The first thing we have to understand is what each type of file extension is.   There is a lot of confusion about what DER, PEM, CRT, and CER are and many have incorrectly said that they are all interchangeable.  While in certain cases some can be interchanged the best practice is to identify how your certificate is encoded and then label it correctly.  Correctly labeled certificates will be much easier to manipulat
> Encodings (also used as extensions)
> 
>     .DER = The DER extension is used for binary DER encoded certificates. These files may also bear the CER or the CRT extension.   Proper English usage would be “I have a DER encoded certificate” not “I have a DER certificate”.
>     .PEM = The PEM extension is used for different types of X.509v3 files which contain ASCII (Base64) armored data prefixed with a “—– BEGIN …” line.
> 
> Common Extensions
> 
>     .CRT = The CRT extension is used for certificates. The certificates may be encoded as binary DER or as ASCII PEM. The CER and CRT extensions are nearly synonymous.  Most common among *nix systems
>     CER = alternate form of .crt (Microsoft Convention) You can use MS to convert .crt to .cer (.both DER encoded .cer, or base64[PEM] encoded .cer)  The .cer file extension is also recognized by IE as a command to run a MS cryptoAPI command (specifically rundll32.exe cryptext.dll,CryptExtOpenCER) which displays a dialogue for importing and/or viewing certificate contents.
>     .KEY = The KEY extension is used both for public and private PKCS#8 keys. The keys may be encoded as binary DER or as ASCII PEM.
> 
> The only time CRT and CER can safely be interchanged is when the encoding type can be identical.  (ie  PEM encoded CRT = PEM encoded CER)

mstfldmr:
only wanted to be in accord with other examples

AndreasHeine:
well then we need to stay with .pem because we use the v3 extention! in the other examples they used regular x509 certs.

mstfldmr:
OK, agreed

AndreasHeine:
we should make a seperate example with a version witch fitts x509v3!
maybe i'll find time tomorow...

the other thing is when do we need to reject a client connection if the cert is not valid!? so a miss match of subjectAltname. i guess its not implementet right now...

PR #1027, opened 2020-05-13
Properly remove keepalive thread if session activation fails

helmut-jacob:
Previously we would have just raised an exception in case of
a session activiation failure (for example wrong password) but
the keepalive thread was kept running.

This change makes sure that all setup is torn down again in case
of a failure within "connect".

Hence, the library user can call connect() again if the previous
connect failed without having to explicitly clean up the half-baked
connect by calling disconnect.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -276,11 +276,21 @@ class Client(object):
         try:
             self.send_hello()
             self.open_secure_channel()
-            self.create_session()
+            try:
+                self.create_session()
+                try:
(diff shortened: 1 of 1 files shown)


oroulet:
looks a bit compllicated, isn't there a way to avoid so many tr/except levels?

helmut-jacob:
This seems to be the most correct way of implementing it. We could also just stuff the code into a single try block and call close_session(), close_secure_channel() and disconnect_socket() in sequence.

PR #1042, opened 2020-05-27
fix race connection at secure_open for client

oroulet:


oroulet:
seems to break encryption initialization....

Issue #1057, opened 2020-06-02
.../opcua/common/node.py => Node class => __str__ and __repr__ method


AndreasHeine:
Hey guys i plan to make a PR on Nodeclass as seen from the topic its about the "__str__" and the "__repr__" methods

https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/common/node.py
```
    def __str__(self):
        return "Node({0})".format(self.nodeid)
    __repr__ = __str__
```

what i like to change is that if you cast a node to string 
Now: "str(node)" & "repr(node)" => Node(NumericNodeId(ns=2;i=2))
After: "str(node)" => "ns=2;i=2" and if you call repr(node) => Node(NumericNodeId(ns=2;i=2))

what is your opinion?

@oroulet

oroulet:
Sounds correct. Having str and repr the same is a bit of an abuse

AndreasHeine:
@oroulet => #1058 tell me if its not ok

AndreasHeine:
`__repr__ = __str__`

is used in many places in the same way i could change them in the same manner if you like!?

swamper123:
> 
> 
> `__repr__ = __str__`
> 
> is used in many places in the same way i could change them in the same manner if you like!?

This seems to be usefull, especially when you rewrite both.

oroulet:
This makes sense for nodeid since they have a specified string representation. But do other objects have such a representation?

AndreasHeine:
maybe Browsname would make sence

of cource only in these cases where it makes sence

oroulet:
Ok.

PR #1058, opened 2020-06-02
changed Nodeclass  __str__ and __repr__ method

AndreasHeine:
#1057:

    def __str__(self):
        return self.nodeid[self.nodeid.find("(")+1:self.nodeid.find(")")]
    
    def __repr__(self):
        return "Node({0})".format(self.nodeid)

"str(node)" => "ns=2;i=2" and if you call repr(node) => Node(NumericNodeId(ns=2;i=2))

### DIFF
--- a/opcua/common/node.py
+++ b/opcua/common/node.py
@@ -62,7 +62,8 @@ class Node(object):
         return not self.__eq__(other)
 
     def __str__(self):
-        return self.nodeid[self.nodeid.find("(")+1:self.nodeid.find(")")]
+        node_str = str(self.nodeid)
+        return node_str[node_str.find("(")+1:node_str.find(")")]
     
     def __repr__(self):
         return "Node({0})".format(self.nodeid)
--- a/opcua/common/node.py
+++ b/opcua/common/node.py
@@ -62,8 +62,10 @@ class Node(object):
         return not self.__eq__(other)
 
     def __str__(self):
+        return self.nodeid[self.nodeid.find("(")+1:self.nodeid.find(")")]
+    
+    def __repr__(self):
         return "Node({0})".format(self.nodeid)
-    __repr__ = __str__
 
     def __hash__(self):
         return self.nodeid.__hash__()


AndreasHeine:
... made a mistak (-_-) one sec

AndreasHeine:
```
    def __str__(self):
        node_str = str(self.nodeid)
        return node_str[node_str.find("(")+1:node_str.find(")")]

    def __repr__(self):
        return "Node({0})".format(self.nodeid)
```

swamper123:
Just a question, because I'm not so into oder NodeID types: Is it possible that `\"` or `"` could be part of the nodeid? May this cause trouble?

AndreasHeine:
could you provide an example, i am not sure what you mean?

swamper123:
Something like this:
`"ns=2;s=iams\"anode"` or `"ns=2;b=k2309"lq='"`

AndreasHeine:
i guess this will break something!

swamper123:
Maybe it would be better to construct a clean nodeid string from it's base:

`return f"ns={NamespaceIndex},{NodeIdType}={Identifier}"`
For the NodeIdType the used letter shall be added.

AndreasHeine:
plan was to get something like: 
"ns=2;i=2" or "ns=2;s=MyTemperature" or ns=1;g=09087e75-8e5e-499b-954f-f2a9603db28a 
from -> str(node)

instead on this "Node(NumericNodeId(ns=2;i=2))"

guess you mean the same!

AndreasHeine:
> Something like this:
> "ns=2;s=iams\"anode" or "ns=2;b=k2309"lq='"

maybe its possible to escape some charakters but its a more general problem/question witch chars are allowed in a OPC UA

swamper123:
Yes I mean the same. I also have no clue how other Server/Client provider deal with such things. 
I would guess many other languages have their "special chars" which they have to deal with as well (or it hasn't happened yet).

AndreasHeine:
@swamper123 : ive tested some different node ids

> Something like this:
> "ns=2;s=iams\"anode" or "ns=2;b=k2309"lq='"

you brought me to think more about it...

```
    var1 = await obj.add_variable(ua.NodeId.from_string(f'ns={idx};s=s=iams\"//a/node'), 'var1', 0)
result => ns=2;s=s=iams"//a/node
    var2 = await obj.add_variable(ua.NodeId.from_string(f'ns={idx};s=b=k2309;"lq='), 'var2', 0)
result => ns=2;s=b=k2309
```

you can use most speacial charakters! but not ";" "{" "}" => 'ns=2;s=b=k2309;"lq=' => "ns=2;s=b=k2309"
and you have to make sure that that you dont break the string like: "asdf"1234" instead 'asdf"1234'
or in my opinion the safest way to do it """ns=2;s=b=k"23'0\9;"lq=""" so you use " and ' and / and \

PR #1061, opened 2020-06-03
fix: Update node.py

AndreasHeine:
now it returns propperly if "(" and ")" is in Identifier Type : String

### DIFF
--- a/opcua/common/node.py
+++ b/opcua/common/node.py
@@ -63,7 +63,7 @@ class Node(object):
 
     def __str__(self):
         node_str = str(self.nodeid)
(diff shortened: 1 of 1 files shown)


opcua/common/node.py
```diff
@@ -63,7 +63,7 @@ def __ne__(self, other):
 
     def __str__(self):
         node_str = str(self.nodeid)
-        return node_str[node_str.find("(")+1:node_str.find(")")]
+        return node_str[node_str.find("(")+1:len(node_str)-1]
```
oroulet, line 66: return self.nodeid.to_string() should be enough

PR #1088, opened 2020-07-02
Honor option to instantiate optional children

### LINKED ISSUE #1087: In instantiate_node, the flag for optional instantiation is not honored recursively
Since the flag `instantiate_optional` of `_instantiate_node` is not passed down in recursion, most optional components are in fact always instantiated. The flag currently only affects direct children of the toplevel Object created.

Nebukadneza:
Previously, the flag in which the user specified whether or not to
instantiate optional children was only honored on the top-level object
(i.e., first recursion round)

Fixes #1087


oroulet:
do we have the same issue in opcua-asyncio?

oroulet:
we shoudl add tests for it too...

PR #1093, opened 2020-07-08
Also treat OptionalPlaceholder ModellingRule as optional in `instantiate`

### LINKED ISSUE #1092: Instantiate mis-treats ModellingRule: OptionalPlaceholder
Hi,

in `instantiate`, when passing `instantiate_optional=False`, only children with ModellingRule `Optional` are ignored. However, `OptionalPlaceholder` children should be ignored as well.

Nebukadneza:
`instantiate` with `instantiate_optional=False` would also instantiate
children that have the ModellingRule OptionalPlaceholder. So far, only
Optional children were ignored.

This commit fixes this behavior, and treats both of them the same w.r.t.
`instantiate_optional`.

Fixes #1092


oroulet:
Thanks. Ideally you should also try to add a test for that behavior

Issue #1096, opened 2020-07-09
Subscription with list of nodes fails to unsubscribe


lkaupp:
Hi devs,

~~i got an Error thrown on disconnect after my succesfull unsubscribe. Any ideas why?~~
i got to the root of evil think i have a problem with unsubscribe multiple nodes
 ```
 try:
            client.connect()
            client.load_type_definitions()
            nodesToSubscribe = load_node_list_from_file(path_to_nodes)
            nodesToSubscribe = [Node(client.uaclient, item) for item in nodesToSubscribe]

            handler = SubHandler()
            opcua_subscription = client.create_subscription(100, handler)
            print("subscribe")
            start = time.time()
            print(time.asctime(time.localtime(time.time())))

            opcua_handle = opcua_subscription.subscribe_data_change(nodesToSubscribe)
            end = time.time()
            print(end - start)
            print("handle")
            time.sleep(30)

            print("unsubscribe")
            opcua_subscription.unsubscribe(opcua_handle)
        finally:
             client.disconnect()
```
```
opcua_subscription.unsubscribe(opcua_handle)
  File "C:\Users\Lukas Kaupp\AppData\Local\Programs\Python\Python38-32\lib\site-packages\opcua\common\subscription.py", line 304, in unsubscribe
    results = self.server.delete_monitored_items(params)
  File "C:\Users\Lukas Kaupp\AppData\Local\Programs\Python\Python38-32\lib\site-packages\opcua\client\ua_client.py", line 553, in delete_monitored_items
    data = self._uasocket.send_request(request)
  File "C:\Users\Lukas Kaupp\AppData\Local\Programs\Python\Python38-32\lib\site-packages\opcua\client\ua_client.py", line 81, in send_request
    future = self._send_request(request, callback, timeout, message_type)
  File "C:\Users\Lukas Kaupp\AppData\Local\Programs\Python\Python38-32\lib\site-packages\opcua\client\ua_client.py", line 55, in _send_request
    binreq = struct_to_binary(request)
  File "C:\Users\Lukas Kaupp\AppData\Local\Programs\Python\Python38-32\lib\site-packages\opcua\ua\ua_binary.py", line 261, in struct_to_binary
    packet.append(to_binary(uatype, val))
  File "C:\Users\Lukas Kaupp\AppData\Local\Programs\Python\Python38-32\lib\site-packages\opcua\ua\ua_binary.py", line 284, in to_binary
    return struct_to_binary(val)
[...]

PR #1097, opened 2020-07-09
Unsub list handle

swamper123:
Fix double listing like mentioned in #1096

### DIFF
--- a/opcua/common/subscription.py
+++ b/opcua/common/subscription.py
@@ -300,14 +300,14 @@ class Subscription(object):
             return
         params = ua.DeleteMonitoredItemsParameters()
         params.SubscriptionId = self.subscription_id
-        params.MonitoredItemIds = handles
+        params.MonitoredItemIds = [handle]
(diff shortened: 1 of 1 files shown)


oroulet:
seems to break something

swamper123:
Jep ... I will investigate it today. I guess the handling with single lists was not expected.

lkaupp:
Found your error:
```
        handles = [handle] if type(handle) is int else handle # you spare this operation
        if not handles:
            return
        params = ua.DeleteMonitoredItemsParameters()
        params.SubscriptionId = self.subscription_id
        params.MonitoredItemIds = handle #should be "handles" not handle
```

swamper123:
@lkaupp A wise man told me once: "If you hear hooves stamping, you should think about horses not zebras." 
Well I was looking for zebras everywhere, but not the horse I placed myself there. Thanks for the hint.
It seems that there are some other issues as well. Travis may accept the tests, but Errors still appear. But this is something for other issues.

PR #1112, opened 2020-07-27
Make sure all inverse references are removed on node deletion

helmut-jacob:
Iterating over the references list and deleting inside the same list
resulted in dangling inverse references if nodes where deleted.

Fix this by iterating over a copy of the references list.


helmut-jacob:
Pushed again to bring code in sync with asyncua code (use [:] instead of list())

PR #1116, opened 2020-08-24
Fix locking around internal_subscription queues

helmut-jacob:
Under various situations data change events can be lost due to
insufficient locking.

Data change events will be notified by an aspace callback to
MonitoredItemService which enqueues the accotrding events for processing
by the subscription publishing loop.

From internal_subscription.py:

  339     def _pop_triggered_datachanges(self, result):
  340         if self._triggered_datachanges:
  341             notif = ua.DataChangeNotification()
  342             notif.MonitoredItems = [item for sublist in self._triggered_datachanges.values() for item in sublist]

-> if anything is added to _triggered_datachanges in a second thread it'll be lost, see enqueue_datachange_event below

  343             self._triggered_datachanges = {}
  344             self.logger.debug("sending datachanges notification with %s events", len(notif.MonitoredItems))
  345             result.NotificationMessage.NotificationData.append(notif)

  380     def enqueue_datachange_event(self, mid, eventdata, maxsize):
  381         self._enqueue_event(mid, eventdata, maxsize, self._triggered_datachanges)

-> self._triggered_datachanges if referenced without lock being held leading to datachanges being added to a dangling
   reference of self._triggered_datachanges

Fix this by moving the locking into enqueue_datachange_event and enqueue_event.

### DIFF
--- a/opcua/server/internal_subscription.py
+++ b/opcua/server/internal_subscription.py
@@ -378,25 +378,26 @@ class InternalSubscription(object):
                 return ua.NotificationMessage()
 
     def enqueue_datachange_event(self, mid, eventdata, maxsize):
-        self._enqueue_event(mid, eventdata, maxsize, self._triggered_datachanges)
+        with self._lock:
+            self._enqueue_event(mid, eventdata, maxsize, self._triggered_datachanges)
 
     def enqueue_event(self, mid, eventdata, maxsize):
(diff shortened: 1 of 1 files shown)


helmut-jacob:
The same fix is not necessary for asyncua as the subscription handling was reworked anyway.

Issue #896, opened 2019-09-30
Historizing of events incorrectly uses SourceNode as Emitting node


einarmo:
This means that historizing events will fail if SourceNode is set, generally speaking.

einarmo:
Looks like the fix is fairly simple actually, I'll make a PR

PR #1123, opened 2020-09-03
Fix history_sql.py emitting node is not the source - sql error

AndreasHeine:
changes:

```python
    def save_event(self, event):
        with self._lock:
            _c_sub = self._conn.cursor()

            table = self._get_table_name(event.SourceNode) #Emitting node is not always the Source (event.SourceNode)
            columns, placeholders, evtup = self._format_event(event) #from event_prop_dict
            event_type = event.EventType  # useful for troubleshooting database

            # insert the event into the database
            try:
                _c_sub.execute(
                    'INSERT INTO "{tn}" ("_Id", "_Timestamp", "_EventTypeName", {co}) VALUES (NULL, "{ts}", "{et}", {pl})'
                    .format(tn=table, co=columns, ts=event.Time, et=event_type, pl=placeholders), evtup)

            except sqlite3.Error as e:
                self.logger.error('Historizing SQL Insert Error for events from %s: %s', event.SourceNode, e)

            self._conn.commit()

            # get this node's period from the period dict and calculate the limit
            period = self._datachanges_period[event.SourceNode]

            if period:
                # after the insert, if a period was specified delete all records older than period
                date_limit = datetime.utcnow() - period

                try:
                    _c_sub.execute('DELETE FROM "{tn}" WHERE Time < ?'.format(tn=table),
                                   (date_limit.isoformat(' '),))
                except sqlite3.Error as e:
                    self.logger.error('Historizing SQL Delete Old Data Error for events from %s: %s',
                                      event.SourceNode, e)

                self._conn.commit()
```

### DIFF
--- a/opcua/server/history_sql.py
+++ b/opcua/server/history_sql.py
@@ -163,8 +163,8 @@ class HistorySQLite(HistoryStorageInterface):
         with self._lock:
             _c_sub = self._conn.cursor()
 
-            table = self._get_table_name(event.emitting_node)
-            columns, placeholders, evtup = self._format_event(event)
+            table = self._get_table_name(event.SourceNode) #Emitting node is not always the Source (event.SourceNode)
+            columns, placeholders, evtup = self._format_event(event) #from event_prop_dict
             event_type = event.EventType  # useful for troubleshooting database
 
             # insert the event into the database
@@ -174,12 +174,12 @@ class HistorySQLite(HistoryStorageInterface):
                     .format(tn=table, co=columns, ts=event.Time, et=event_type, pl=placeholders), evtup)
 
             except sqlite3.Error as e:
-                self.logger.error('Historizing SQL Insert Error for events from %s: %s', event.emitting_node, e)
+                self.logger.error('Historizing SQL Insert Error for events from %s: %s', event.SourceNode, e)
 
             self._conn.commit()
 
             # get this node's period from the period dict and calculate the limit
-            period = self._datachanges_period[event.emitting_node]
+            period = self._datachanges_period[event.SourceNode]
 
             if period:
                 # after the insert, if a period was specified delete all records older than period
@@ -190,10 +190,10 @@ class HistorySQLite(HistoryStorageInterface):
                                    (date_limit.isoformat(' '),))
                 except sqlite3.Error as e:
                     self.logger.error('Historizing SQL Delete Old Data Error for events from %s: %s',
-                                      event.emitting_node, e)
+                                      event.SourceNode, e)
 
                 self._conn.commit()
-
+                
     def read_event_history(self, source_id, start, end, nb_values, evfilter):
         with self._lock:
             _c_read = self._conn.cursor()


zerox1212:
Aren't you reintroducing this issue?
https://github.com/FreeOpcUa/python-opcua/issues/896

AndreasHeine:
dont understand #896 what was the problem ? without you can only historize one event....

zerox1212:
> dont understand #896 what was the problem ? without you can only historize one event....

I suppose if `SourceNode` changes (such as if multiple nodes are emitting via server event generator) history will crash because the table won't exist.

AndreasHeine:
then let us test different scenarios!?

AndreasHeine:
@zerox1212 right now it is like he cant find a the event property because he is looking for the wrong table! it only works for the default opcua-event-types (equal porps = columns) and not for the custom ones (different columns)... atm we build tables but did not write in them just in one (server table) results in no event history for custom event types...

@einarmo could you discribe the problem you encounterd?

einarmo:
The emitting node and source node of an event in OPC-UA are two very different things. The emitting node is the node that has a EventNotifier set, and which may be subscribed to by a client in order to receive events. The SourceNode is simply an attribute of the BaseEventType (you can find it in the type-hierarchy yourself), that indicates which node is responsible for the event.

For example, an event may be emitted from some higher level node, because that is the main node of the OPC-UA server that generated that event, but it may actually concern a completely different node much deeper in the hierarchy.

In fact, there is no reason why there would be a SourceNode at all. An event can concern the server as a whole, and does not have to have a logical source in the hierarchy.

The emitting node, on the other hand, will always have a value. Any HistoryRead operation will use the emitting node, so it is logical to use the emitting node as table name.

AndreasHeine:
> In fact, there is no reason why there would be a SourceNode at all. An event can concern the server as a whole, and does not have to have a logical source in the hierarchy.

thats not according to the specification! every event which we create is a subtype of BaseEventType.
https://reference.opcfoundation.org/v104/Core/docs/Part5/6.4.2/ 
and the SourceNode is the only way to devide!

https://github.com/FreeOpcUa/opcua-asyncio/issues/275

@einarmo please provide me an example in which the source-node breaks something or misbehave in your oppinion and i will have a look on it (atm our lib is not consistent... stuff is broken or just half rewritten...)!

AndreasHeine:
if it is null/none => special case => has to be treated differently 
if it is not null/none => regular case (originated from a node) => store in a table named with its source

we need both cases not only one...

einarmo:
I really don't see the issue with using the emitting node. That way there is no null case, it is consistent, and it is efficient. In your case we would need to

 - Keep track of which events are emitted from which node, so that when a client reads history from a node, we need to find out which tables correspond to that emitter, or read from _every_ table, and filter on an emittingNode field.
 - Handle the ugly null case

If we use the emitting node, we avoid both of these issues. It is more efficient, cleaner, and closer to the standard.

I don't fully understand what issue this PR is supposed to solve, do you have an example?

AndreasHeine:
still waiting for your example ;) try to get your point of view... there are not much ways to store it in a db
1. all in one event table (complex lookups, giant tables, slower over time in long periods) <= would be the result of emitting-node
2. by EventType (a little better but still difficult to devide same eventtypes to a source maybe a solution if source is null)
3. by SourceNode (in my oppinion the best way to get events only from a specific source)

but first things first!

i was working on missing async/await syntax in asyncua "server-events-history.py" https://github.com/FreeOpcUa/opcua-asyncio/issues/275 found it in python-opcua aswell:

server-events-history.py

```
  File "C:\Users\HeineAn\AppData\Local\Programs\Python\Python37-32\lib\site-packages\aiosqlite\core.py", line 153, in run
    result = function()
sqlite3.OperationalError: table 0_2253 has no column named MyNumericProperty
Historizing SQL Insert Error for events from NumericNodeId(i=2253): table 0_2253 has no column named MyNumericProperty
```

that was the initial piont to look deeper

https://github.com/FreeOpcUa/python-opcua/issues/896 => broke stuff which already worked => the result of changing it to emitting-node is that all database stuff has to be rewritten, because the columns doesnt match the table

> It is more efficient, cleaner, and closer to the standard.

-keeping tables small will improve performence on lookup
-the standard does not define the database storage / structure

dont get me wrong i am looking for a solution but in my oppinion emitting-node was a small change with side effects, atm we dont historize the events and arent able to read the event history (reason table is empty)

hope now its clear what my point of view is!

einarmo:
I'm trying to recall exactly what kind of issue caused me to make this change. I believe it was related to null SourceNode, or reading from non-server emitters. I absolutely did it because of a bug.

What I find is that SourceNode is very arbitrary. It is just some random attribute on the event. 

Essentially, your argument is that it is easier to change this here, breaking whatever #896 fixed, than actually making the fix in the asyncio library?

 > keeping tables small will improve performence on lookup

In this case this isn't true. Remember that all HistoryRead calls _must_ operate on the emitting node. The code before I made the change essentially _enforced_ SourceNode = emitting node, which is clearly a bug as I explained. To make this change you would need to rewrite everything so that HistoryRead for, say, node i=1 goes like this:

 - Receive HistoryRead request for events on node i=1
 - Verify that i=1 is a historizing event-emitter
 - Read events from _all_ event tables, since SourceNode is arbitrary
 - Filter events that were emitted from node i=1

If we use emitting node instead:

 - Receive HistoryRead request for events on node i=1
 - Verify that i=1 is a historizing event-emitter
 - Read events from the tables corresponding to i=1

Say that we have 100 emitters, each with the same number of events. Your solution means reading 100 times more data.

AndreasHeine:
> Essentially, your argument is that it is easier to change this here, breaking whatever #896 fixed, than actually making the fix in the asyncio library?

no! the point is atm we have a not working part in this lib and the last working stand was with SourceNode (consistent to examples and the rest of event history) sadly it got merged without testing (missing unittest)... but never mind

if we take https://github.com/FreeOpcUa/python-opcua/blob/master/examples/server-events-history.py as a base what has to be changed/added to have a working example to reproduce the bug you mention (later we can extract a unittest of that case)?

einarmo:
Again, it's been a while, I used this library to create test servers, which I tested with the official .NET SDK, there may be bugs in the python _client_, but the .NET SDK is going to be correct.

This is the relevant part of the test server I used:

`eventPropType = server.create_custom_event_type(2, 'EventExtraProperties', ua.ObjectIds.BaseEventType, [('PropertyNum', ua.VariantType.Float), ('PropertyString', ua.VariantType.String), ("SubType", ua.VariantType.String)])
    eventBasicType1 = server.create_custom_event_type(2, 'EventBasic1', ua.ObjectIds.BaseEventType)
    eventBasicType2 = server.create_custom_event_type(2, 'EventBasic2', ua.ObjectIds.BaseEventType)
    eventCustomType = server.create_custom_event_type(2, 'EventCustomType', ua.ObjectIds.BaseEventType, [('TypeProp', ua.VariantType.String)])


    # Test basic getting of events with properties
    propEmitter = server.get_event_generator(eventPropType)
    propEmitter.event.SourceNode = myobj.nodeid
    propEmitter.event.PropertyNum = 1
    propEmitter.event.SubType = "TestSubType"
    propEmitter.event.PropertyString = "TestString 0"
    # Test getting of events from other emitter
    propEmitterOther = server.get_event_generator(eventPropType, myobj)
    propEmitterOther.event.SourceNode = myobj.nodeid
    propEmitterOther.event.PropertyNum = 1
    propEmitterOther.event.SubType = "TestSubType"
    propEmitterOther.event.PropertyString = "TestStringOther 0"
    propEmitterOther.event.SourceName = myobj.get_display_name().Text
    # Test filtering out events from third emitter
    propEmitterOther2 = server.get_event_generator(eventPropType, myobj2)
    propEmitterOther2.event.SourceNode = myobj.nodeid
    propEmitterOther2.event.SourceName = myobj.get_display_name().Text
    # Test getting of second event type from same emitter
    basicPassEmitter = server.get_event_generator(eventBasicType1)
    basicPassEmitter.event.SourceNode = myobj.nodeid
    basicPassEmitter.event.SourceName = myobj.get_display_name().Text
    # Test filtering out event types
    basicBlockEmitter = server.get_event_generator(eventBasicType2)
    basicBlockEmitter.event.SourceNode = myobj.nodeid
    basicBlockEmitter.event.SourceName = myobj.get_display_name().Text
    # Test with different source nodes
    basicPassSourceEmitter = server.get_event_generator(eventBasicType1)
    basicPassSourceEmitter.event.SourceNode = myobj2.nodeid
    basicPassSourceEmitter.event.SourceName = myobj2.get_display_name().Text
    
    basicPassSourceEmitter2 = server.get_event_generator(eventBasicType1, myobj)
    basicPassSourceEmitter2.event.SourceNode = myobj2.nodeid
    basicPassSourceEmitter2.event.SourceName = myobj2.get_display_name().Text
    
    basicVarSourceEmitter = server.get_event_generator(eventBasicType1)
    basicVarSourceEmitter.event.SourceNode = myvar.nodeid
    basicVarSourceEmitter.event.SourceName = myvar.get_display_name().Text
    
    basicNoSourceEmitter = server.get_event_generator(eventBasicType1)
    
    basicExcludeSourceEmitter = server.get_event_generator(eventBasicType1)
    basicExcludeSourceEmitter.event.SourceNode = myobjexclude.nodeid 
    basicExcludeSourceEmitter.event.SourceName = myobjexclude.get_display_name().Text

    mappedTypeEmitter = server.get_event_generator(eventCustomType)
    mappedTypeEmitter.event.TypeProp = "MySpecialType"
    mappedTypeEmitter.event.SourceNode = myobj.nodeid
    mappedTypeEmitter.event.SourceName = myobj.get_display_name().Text

    # Make server node event-historizing
    server_node = server.get_node(ua.ObjectIds.Server)
    server.historize_node_event(server_node, period=None)
    
    try:
        count = 0
        while True:
            time.sleep(1)            
            propEmitter.event.PropertyString = "TestString " + str(count)
            propEmitter.trigger(message="prop " + str(count))
            
            propEmitterOther.event.PropertyString = "TestStringOther " + str(count)
            propEmitterOther.trigger(message="propOther " + str(count))

            propEmitterOther2.trigger(message="propOther2 " + str(count))
            basicPassEmitter.trigger(message="basicPass " + str(count))
            basicBlockEmitter.trigger(message="basicBlock " + str(count))
            basicPassSourceEmitter.trigger(message="basicPassSource " + str(count))
            basicPassSourceEmitter2.trigger(message="basicPassSource2 " + str(count))
            basicVarSourceEmitter.trigger(message="basicVarSource " + str(count))
            basicNoSourceEmitter.trigger(message="basicNoSource " + str(count))
            basicExcludeSourceEmitter.trigger(message="basicExcludeSource " + str(count))
            mappedTypeEmitter.trigger(message="mappedType " + str(count))
            count += 1`

AndreasHeine:
ok i will have a look on that, what i can tell you is that your change only work if there is just one event and/or if they all have the same type (BaseEventType)! if you have multiple events from different types (BaseEventtype + Custom with additional props) under a object it breaks. 

could take a while!

einarmo:
> ok i will have a look on that, what i can tell you is that your change only work if there is just one event and/or if they all have the same type (BaseEventType)! if you have multiple events from different types (BaseEventtype + Custom with additional props) under a object it breaks.
> 
> could take a while!

Ah, I see. I did consider that at one point, and I can see how that could happen. Sorting by SourceNode will _not_ fix that, however, as it is fully possible to emit different events with the same SourceNode.

I don't think this PR fixes it either, or if it does, it is purely accidental.

AndreasHeine:
> As for base types, it is not legal to create events that are not derived from BaseEventType in OPC-UA.

if i talk from custom events i mean derived from BaseEventType + additional Properties!

einarmo:
I think the actual solution to this, and I've done something similar in a different case, is to group by both event-type and emitter, then when querying from an emitter, search all tables for that node.

zerox1212:
I like the discussion here. I just want to make a point, that when I implemented SQL history I only tested it with Python OPC UA. In our library the SourceNode will match emitter by default. So it always worked by chance.

I do not know how often people are changing source node property. I always had 1 emitter per node that generated events. Routing all events through a single emitter doesn't seem very robust. I do not like it's even allowed to emit events with `SourceNode` null. That just makes our job difficult for no reason since an event must always come from some where and in any other case it should come from the server itself. Bad spec in my opinion.

At a minimum we should revert to SourceNode until a better solution is found. At least this way examples work and Python OPC UA is at least fully compatible with itself.

zerox1212:
One other thing, design a flexible history is not going to be easy because tracking specific nodes in a server is difficult, and if the address space changes then what do we do? If a user changes event properties after history has been started you will again have missing columns.

Plus it's still based on subscription to record data which adds overhead and if user stops the server history won't record. I was new at Python when I wrote this, and I think in the end it's a bad implementation that probably needs to be scrapped.

einarmo:
I still think using emitterId is the best solution:

 - It is always known and non-null when reading, since any HistoryRead operation will use it
 - It is always known and non-null when writing, since events must be emitted from a node in the standard
 - It applies to any event

I've implemented something very similar to this once, storing OPC-UA events in influxdb. It was necessary there to have a "table" for each emitter and event-type combination. Fortunately, here we always know about all such combinations, because that is how we define event generators.

As for changes to the node hierarchy, SourceNode does not have to be consistent. We can require it, and it really should be, but in the end that specific detail should be handled by the user. If an emitter is deleted, then much like if a normal historical variable is deleted, we should delete the history. It is perfectly reasonable to tie event history to the node that generated it.

AndreasHeine:
just thinking loud:
i guess the solution to both is to use a combination of tables mixed with the right choice of primary and foraign keys but this will take some time to get it right! probably its better at this point to make a step back to a working stand (even if its not perfect but consistent inside the lib) and in the next step we can make a plan and sql structure (@zerox1212 @einarmo together if you like) which makes all of us happy?

oroulet:
Can someone summarize the issue?
I did not really underdtand the issue and did not read closely the code but it kind of looks like there is a table per emitting or source node. Is that correct? Why not saving everything in one big table with both attributes then?

oroulet:
Also should I merge that one?

einarmo:
@oroulet The issue at the bottom of this is that each event-type may have different attributes, so we can't really store different events in the same table. For @AndreasHeine switching to grouping by SourceNode happened to fix the issue, for me with #896 it was fixed by grouping by emitting node...

This PR just changes the issue into another form (it breaks using null SourceNode, without really fixing the underlying issue). It is pretty subjective whether or not it is better.

Either way, the real solution is to group by both event type and emitting node, so create a table-name on the form `[emitter]-[type]`. I don't really think this is all that difficult to do.

oroulet:
Thanks for the summary @einarmo and what about saving everything in one big table with both attributes? I do not really see the point of having several table. you can just do a query.

einarmo:
@oroulet The difficulty is that it is impossible to know all attributes at runtime, we would have to modify the table when a new emitter is added, which seems unnecessarily complicated. Easier to add a new table each time we create a new event generator.

oroulet:
That does not make sense.Just make one big table something like

emitter TEXT , source TEXT,  srvts TIMESTAMP, source_teimstamp TIMESTAM, etc...

And when historizing a new new node, just a add a subscription adding lines to the same table

when reading history for a node, just make a query on that table

oroulet:
since we are using a not types db you also add one more column to save all resting attributes to json or whatever (not binary, they will give us comptability issues)

AndreasHeine:
sorry for my late reponse had not much time this week.. this PR just makes it back consitent to the point things work together (like @zerox1212 worked it out)!
@einarmo i agree that this is not the final thing! SQL-History has to be reworked (my plan is next week) but to leaving things broken is not a solution.

from @zerox1212 
> At a minimum we should revert to SourceNode until a better solution is found. At least this way examples work and Python OPC UA is at least fully compatible with itself.

that also my oppinion!

einarmo:
> That does not make sense.Just make one big table something like
> 
> emitter TEXT , source TEXT, srvts TIMESTAMP, source_teimstamp TIMESTAM, etc...
> 
> And when historizing a new new node, just a add a subscription adding lines to the same table
> 
> when reading history for a node, just make a query on that table

In OPC-UA, events are dynamically modeled in the type hierarchy, and it is possible to add an arbitrary number of new attributes, even at runtime.

einarmo:
> sorry for my late reponse had not much time this week.. this PR just makes it back consitent to the point things work together (like @zerox1212 worked it out)!
> @einarmo i agree that this is not the final thing! SQL-History has to be reworked (my plan is next week) but to leaving things broken is not a solution.
> 
> from @zerox1212
> 
> > At a minimum we should revert to SourceNode until a better solution is found. At least this way examples work and Python OPC UA is at least fully compatible with itself.
> 
> that also my oppinion!

In the end it doesn't really matter to me, as I don't use this particular feature anymore. My point is that reverting to SourceNode is not productive, it just breaks one thing, doesn't really fix the main issue, and just accidentally fixes some specific circumstances.

einarmo:
> since we are using a not types db you also add one more column to save all resting attributes to json or whatever (not binary, they will give us comptability issues)

Adding a json field works. I still think it is easier to just add a table per type, since we already explicitly create event generators for each type/emitter combination. It fits better with the current architecture.

oroulet:
>> In OPC-UA, events are dynamically modeled in the type hierarchy, and it is possible to add an arbitrary number of new attributes, even at runtime.

yes they are and the number of properties saved is already hard coded. Adding the source and emiter as columns instead of table name is just better DB design and doesnot change anything to the dynamic attributes issue

einarmo:
> > > In OPC-UA, events are dynamically modeled in the type hierarchy, and it is possible to add an arbitrary number of new attributes, even at runtime.
> yes they are and the number of properties saved is already hard coded. 

Not true. It is fully possible to add custom nodes to the type hierarchy, and this sdk supports that. We would have to use JSON.

Even without that, take a look at the event type hierarchy, there are a _lot_ of different possible properties.

einarmo:
Also, if anything is bad database design it is sticking every type of object in the same table. If any object you insert into the table only uses 5% of the columns, you are doing something wrong.

oroulet:
>> Not true. It is fully possible to add custom nodes to the type hierarchy, and this sdk supports that. 

I wrote most of that code so I know...

oroulet:
the fact is that Events have something like 5 or 10 default attributes so we should have a column for each of them. Then users can add attributes to events, so most probably we should save them in one column one way or another (json for example) or not save them at all like it is done now (If I read the current code correctly because I did not write the sql part)
Another possibility is to use another Db something like nosql which allow by design to save whatever you want, but this should another history plugin

zerox1212:
@oroulet Events with custom properties are historized currently, that is why we make a table per (custom) event generator. Each event generator gets a table with the necessary columns.

If we are going to serialize the custom properties then we should at least use XML so we don't have yet another complex serializer/parser. However I'm against having all events in a single table, if nothing else for performance reasons.

oroulet:
>> Events and their custom properties are historized currently, that is why we make a table per (custom) event generator.
 
good point

>>  then we should at least use XML so we don't have yet another complex serializer/parser. 

yes good point too. I had my head in another project using json everywhere...

AndreasHeine:
why we dont use the beneficial features of sql? 
one table for events (mapping), eventtypes, emitting nodes, source-nodes... the querys would be more complex but i think it would be possible to get it all together (above conversation)

!?

zerox1212:
@AndreasHeine If you want to make a nicer, more proper SQL design, go for it. I wrote this SQL implementation to produce a basic history with permanence because the current one was only an in memory dict. Doing a real SQL design with some primary keys might be better and more flexible for the future.

AndreasHeine:
first things first, merge! so we are back in a consistend state and in the next step i would make a database-design-draft put it into an issue and we can discuss it!?

PR #1130, opened 2020-09-15
Move user_manager from Server into InternalServer

JensPfeifle:
Removes the need to patch in the Server object via the _parent attribute
in InternalServer if the InternalServer is instantiated outside of
Server.__init__.

Before this commit, the Server object passed itself into the
InternalServer constructor. If we instantiate an InternalServer (that we
want to pass in) before the Server, we obviously don't have a Server
instance to pass in as the parent. So we end up having to do this:

>>> my_iserver = CustomInternalServer()
>>> server = Server(iserver = my_iserver)
>>> server.iserver._parent = server # <-- no longer necessary

To preserve compatibility and keep changes to a minimum, a user_manager
property and setter are added to the Server class which redirect to the
user_manager attribute of the internal server.

### DIFF
--- a/opcua/server/internal_server.py
+++ b/opcua/server/internal_server.py
@@ -39,10 +39,9 @@ class SessionState(Enum):
 
 class InternalServer(object):
 
-    def __init__(self, shelffile=None, parent=None, session_cls=None):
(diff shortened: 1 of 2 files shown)


oroulet:
thanks.

Issue #926, opened 2019-12-27
Secure channel renewal is broken for security other than NoSecurity


brubbel:
**Background:**
When opening a secure channel, the server sends a `security_token.RevisedLifetime` to the client with value `params.RequestedLifetime`. For UaExpert this is by default 3600 sec.
https://github.com/FreeOpcUa/python-opcua/blob/5bcb6820877db7f73d153430df2e1eedbca58b78/opcua/common/connection.py#L183
Then after 3/4 of the RevisedLifetime (45 minutes by default), UaExpert will renew the secure channel  (`SecurityTokenRequestType.Renew`).
Which always fails and closes the connection if the client tries to communicate or when the subscription service tries to send an update to the client, whichever comes first.

**Problem:**
1. Renewal of the secure channel is first handled here, with a call to select_policy:
https://github.com/FreeOpcUa/python-opcua/blob/5bcb6820877db7f73d153430df2e1eedbca58b78/opcua/common/connection.py#L323
2. Select policy for an encrypted channel uses for example the policy factory SecurityPolicyBasic256Sha256 here:
https://github.com/FreeOpcUa/python-opcua/blob/5bcb6820877db7f73d153430df2e1eedbca58b78/opcua/crypto/security_policies.py#L518
3. which, in turn creates a new Cryptography instance here:
https://github.com/FreeOpcUa/python-opcua/blob/5bcb6820877db7f73d153430df2e1eedbca58b78/opcua/crypto/security_policies.py#L559
The Cryptography instance is initialized with Encryptor/Decryptor references set to None.
However these Encryptor/Decryptor references will never be initialized here:
https://github.com/FreeOpcUa/python-opcua/blob/5bcb6820877db7f73d153430df2e1eedbca58b78/opcua/common/connection.py#L179-L194
because (a) the secure channel is in the open state and (b) the SecurityTokenRequestType is Renew.
The following lines (required to initialize the Encryptor/Decryptor) are never called:
https://github.com/FreeOpcUa/python-opcua/blob/5bcb6820877db7f73d153430df2e1eedbca58b78/opcua/common/connection.py#L188-L189

After that, any communication over the secure channel will fail when trying to access the required methods required for message_to_binary(). E.g.
https://github.com/FreeOpcUa/python-opcua/blob/5bcb6820877db7f73d153430df2e1eedbca58b78/opcua/crypto/security_policies.py#L125

**Suggested solution:**
[...]

PR #1135, opened 2020-09-20
Simple fix suggested by brubbel.

TurtleDogBird:
Pull request for the fix proposed by Brubbel in issue https://github.com/FreeOpcUa/python-opcua/issues/926.

As far as I can tell it fixes the problem.

### DIFF
--- a/opcua/common/connection.py
+++ b/opcua/common/connection.py
@@ -299,7 +299,14 @@ class SecureConnection(object):
         if header.MessageType == ua.MessageType.SecureOpen:
             data = body.copy(header.body_size)
             security_header = struct_from_binary(ua.AsymmetricAlgorithmHeader, data)
-            self.select_policy(security_header.SecurityPolicyURI, security_header.SenderCertificate)
+
+            if self.is_open():
+                # In case the channel is already open we don't want to call
+                # select_policy. Since it will break the Secure channel renewal
+                pass
+            else:
+                self.select_policy(security_header.SecurityPolicyURI, security_header.SenderCertificate)
+
         elif header.MessageType in (ua.MessageType.SecureMessage, ua.MessageType.SecureClose):
             data = body.copy(header.body_size)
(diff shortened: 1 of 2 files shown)


opcua/common/connection.py
```diff
@@ -299,7 +299,14 @@ def receive_from_header_and_body(self, header, body):
         if header.MessageType == ua.MessageType.SecureOpen:
             data = body.copy(header.body_size)
             security_header = struct_from_binary(ua.AsymmetricAlgorithmHeader, data)
-            self.select_policy(security_header.SecurityPolicyURI, security_header.SenderCertificate)
+
+            if self.is_open():
+                # In case the channel is already open we don't want to call
+                # select_policy. Since it will break the Secure channel renewal
+                pass
+            else:
```
oroulet, line 307: better write:

if not self.is_open():
   do something

instead of suing pass. but keep the comment

opcua/server/uaprocessor.py
```diff
@@ -51,8 +51,13 @@ def send_response(self, requesthandle, seqhdr, response, msgtype=ua.MessageType.
...
 
-        self._connection.select_policy(
-            algohdr.SecurityPolicyURI, algohdr.SenderCertificate, request.Parameters.SecurityMode)
+        if self._connection.is_open():
+            # In case the channel is already open we don't want to call
+            # select_policy. Since it will break the Secure channel renewal
+            pass
+        else:
+            self._connection.select_policy(
+                algohdr.SecurityPolicyURI, algohdr.SenderCertificate, request.Parameters.SecurityMode)
```
oroulet, line 60: same here

oroulet:
also good work @TurtleDogBird @brubbel . I suppose may were annoyd by that issue

opcua/common/connection.py
TurtleDogBird: done

PR #1138, opened 2020-10-02
Demo GUI for real-time visualization

tushar-kamra:


iudeen (changes requested):
Can import statements be formatted according to the standards given in [PEP8](https://www.python.org/dev/peps/pep-0008/#imports)?

> Imports should be grouped in the following order:


> Standard library imports.
> Related third party imports.
> Local application/library specific imports.
> You should put a blank line between each group of imports.

zerox1212:
I do not think this work really fits with the purpose of this library. The library is for implementing the OPC UA standard protocol. The only place GUI stuff might apply is with the client gui repo.

swamper123:
I agree with @zerox1212. 
I don't won't to make your work bad in any way, it is quite impressive to build something like this just in tkinter (I wish one day there will be an easy GUI maker like MFC-classes in C++ or similar in pure Python....). 
FreeOpcUa offers an [opcua-client-gui](https://github.com/FreeOpcUa/opcua-client-gui) repo already and this repo, is just about implementing the OPC UA standard protocol and not a Visualisation via a GUI.

oroulet:
It is a gui for what btw?

swamper123:
Getting Nodes from a server as client with different kind of signals, plot the values in a Tkinter enviroment and log them (that's what I interpret out of the code).

Issue #1127, opened 2020-09-09
Crash if ApplicationDescription.ApplicationUri is None


CuriousCrook:
I tried to establish a connection to a proprietary OPC UA server using the opcua-client-gui. The application crashes if you press the "Querry server capability" button in the ConnectionDialog. 
It seems that the application_to_strings() method cannot handle ApplicationDescription.DiscoveryUrls == None.
In my case, the method application_to_strings was called implicitly by endpoint_to_strings ().

_In python-opcua/opcua/tools.py:_
```
def application_to_strings(app):
 ...
    **for url in app.DiscoveryUrls:** # Crash here -> TypeError: 'NoneType' object is not iterable
        result.append(('Discovery URL', url))
    return result  # ['{}: {}'.format(n, v) for (n, v) in result]
```

**I assume that a server without at least one discovery url is generally allowed. (Unfortunately, I don't know my way around the point) Can someone look at this please?**

_Console output:_
uaclient.uaclient - INFO - Endpoint 1:')
uawidgets.utils - ERROR - 'NoneType' object is not iterable')
Traceback (most recent call last):
  File "C:\_repo\opcua-client-gui-master\venv\lib\site-packages\uawidgets\utils.py", line 21, in wrapper
    result = func(self, *args)
  File "C:\_repo\opcua-client-gui-master\uaclient\connection_dialog.py", line 34, in query
    endpoints = self.parent.uaclient.get_endpoints(self.uri)
  File "C:\_repo\opcua-client-gui-master\uaclient\uaclient.py", line 48, in get_endpoints
    for (n, v) in endpoint_to_strings(ep):
  File "C:\_repo\opcua-client-gui-master\venv\lib\site-packages\opcua\tools.py", line 448, in endpoint_to_strings
    result += application_to_strings(ep.Server)
  File "C:\_repo\opcua-client-gui-master\venv\lib\site-packages\opcua\tools.py", line 423, in application_to_strings
    for url in app.DiscoveryUrls:
TypeError: 'NoneType' object is not iterable
uawidgets.utils - WARNING - Error class <uaclient.connection_dialog.ConnectionDialog object at 0x000001E0ED21E3A8>s no member show_error or error')

AndreasHeine:
guess it should be an empty list not none and if its none it should be overwritten with an empty list!?

CuriousCrook:
> guess it should be an empty list not none and if its none it should be overwritten with an empty list!?
> 
[...]

PR #1144, opened 2020-10-12
Update ua_binary.py

CuriousCrook:
Fix for issue #1127 

Ran 410 tests in 75.357s
OK

All tests ran file and the connection to the Softing Server now also works for me.

### DIFF
--- a/opcua/ua/ua_binary.py
+++ b/opcua/ua/ua_binary.py
@@ -505,6 +505,7 @@ def struct_from_binary(objtype, data):
         val = from_binary(uatype, data)
         if val is None and uatype[:6] == 'ListOf':
             val = []
+            
         setattr(obj, name, val)
     return obj
 
--- a/opcua/ua/ua_binary.py
+++ b/opcua/ua/ua_binary.py
@@ -503,6 +503,8 @@ def struct_from_binary(objtype, data):
             if not test_bit(val, idx):
                 continue
         val = from_binary(uatype, data)
+        if val is None and uatype[:6] == 'ListOf':
+            val = []
         setattr(obj, name, val)
     return obj
 


AndreasHeine:
those port errors... maybe we should modify the test to try/except for next free port.
just commit a blank line and the test will be run again!

CuriousCrook:
I didn't understand the problem, but it's working now ;-)

opcua/ua/ua_binary.py
```diff
@@ -503,6 +503,9 @@ def struct_from_binary(objtype, data):
             if not test_bit(val, idx):
                 continue
         val = from_binary(uatype, data)
+        if val is None and uatype[:6] == 'ListOf':
+            val = []
+
```
oroulet, line 508: What is the issue exactly?the UA protocols allows to return Null values which are best translated to None in Python. So the original code looked correct

oroulet, line 508: What is the issue exactly?the UA protocols allows to return Null values which are best translated to None in Python. So the original code looked correct

AndreasHeine:
@CuriousCrook can you track down where the doscovery urls got overwritten? its got initialized as an empty list!?

opcua/ua/ua_binary.py
CuriousCrook: Because the empty lists are iterated in the more high level modules. A non iterable NoneType
leads to an error there. See issue #1127. This error occurred when trying to communicate with a Softing server.

For example:
File "C:_repo\opcua-client-gui-master\venv\lib\site-packages\opcua\tools.py", line 423, in application_to_strings
for url in app.DiscoveryUrls:
TypeError: 'NoneType' object is not iterable

Instead of checking the lists against None everywhere I chose to fix this issue on its roots.

CuriousCrook:
> @CuriousCrook can you track down where the doscovery urls got overwritten? its got initialized as an empty list!?

See above. 
It is overwritten in opcua / ua / ua_binary.py.

In my opinion it is initialized 

In my opinion the list DiscoveryUrls is initialized in a meaningful way. Then it is overwritten by the server response with None.

You can just call application_to_strings(app): in python-opcua/opcua/tools.py while connected to a server without DiscoveryUrls. to reproduce this error. The Softing server seems to return "null", which seems to be valid by the opc ua specification.

I would also like to write a Test, but I haven't found out how to initialize the DiscoveryUrls in the python-opcua server with None.

CuriousCrook:
Let's get the facts together, then we can make a fact-based decision:

Solution according to the Pull-Request
pro:
- Uniform representation of empty lists (initialization status and after a response from server)
- Universal solution for all lists
- List do not have to be checked against None before using (seems more pythonic to me)
- Change only necessary in one place in the source code

contra:
- May slow down communication. (just a guess. Can we measure whether this change is actually causing a relevant slowdown?)

Do you know any other advantages or disadvantages?

oroulet:
but if you change the protocol to return [] everytime we receiver NULL then you may get screwed if one days at one place it is necessary to receiver NULL/None because it has a special meaning

oroulet:
unless someone demonstrate that empty lists cannot be null in UA spec we cannot merge this

CuriousCrook:
This solution was not accepted.

PR #1149, opened 2020-10-19
 Add batching functionality to delete_nodes to prevent timeouts

### LINKED ISSUE #1148: client.delete_nodes on a large set of nodes can lead to timeout-disconnect
If using `client.delete_nodes` on a large set of nodes on a `python-opcua` server, the client will disconnect with a timeout. This is either true if the list of nodes is long, or `recursive=True` is used on a node with lots of childs.

(PR incoming)

Nebukadneza:
When `delete_nodes` is given a large amount of nodes, or is instructed
to recursively delete a big structure, the connection between client and
server can time out.

To remedy this, allow the user to supply a `batch` parameter, which
controls how many nodes are deleted in one request.

closes #1148


oroulet:
why does that help? the client must anyway wait for all chunks to be processed isnt'it?

Nebukadneza:
Indeed, in the end the client must wait until everything is processed on serverside. However, when deleting large structures, using python-opcua as client and as server, the delete operation will time out, because the server takes to long to process the single request. Allowing to batch these into managable chunks will at least keep the network-side working, and the requests themselves won’t time-out.

PR #1152, opened 2020-10-21
Update split of NodeID from string

Musiame:
Split at the first ";" only, to handle the case where ";" is in the string identifier like in `"ns=1;s=V40;0;1"`


oroulet:
that mere request breaks tests. you need to fix tests if you want it merged

Issue #1192, opened 2020-12-10
Servertimestamp not set automatically of client does not set the Servertimestamp


mbay-SAG:
This is a bug report. to get help go to https://github.com/FreeOpcUa/python-opcua/discussions

**Describe the bug**
If you use an OPCUA Client such as UA Expert from Prosys and change values within the gui than the Servertimestamp is not set while writing the node, which is from our perspective not as specified by the OPCUA guidelines.
The Timestamp will be set to EPOCH Time which is on Windows in the year 1601. 
The bug is contained in address_space.py in the write function. There the datavalue for Servertimestamp should be set before writing the node.

**To Reproduce**
Start a simple server with any structure. Use UAEpert, write variables from the GUI. See that Timestamp is set to EPOCH time.

**Expected behavior**
The Servertimestamp should set the timestamp while wirting the node if a new value is set that does not contain the set Servertimestamp.
If the Sourcetimestamp is not set by the client, it is to be discussed whether the SourceTimestamp should be set as well

**Screenshots**

mbay-SAG:
We fixed that in the write function and referenced that already in: 
[https://github.com/FreeOpcUa/python-opcua/issues/850]
If you open a branch we can commit the changed for a pull-request.

AndreasHeine:
you can fork -> create a branch -> commit your changes -> PR

mbay-SAG:
Thanks @AndreasHeine for referencing to the specs. Just created the PR.

PR #1197, opened 2020-12-11
The OPCUA Server sets the Servertimestamp in the moment of writing th…

mbay-SAG:
If you use an OPCUA Client such as UA Expert from Prosys and change values within the gui than the Servertimestamp is not set while writing the node, which is from our perspective not as specified by the OPCUA guidelines.
The Timestamp will be set to EPOCH Time which is on Windows in the year 1601.
The bug is contained in address_space.py in the write function. There the datavalue for Servertimestamp should be set before writing the node.

See: [(https://github.com/FreeOpcUa/python-opcua/issues/1192)]

If the Sourcetimestamp is not set by the client, you can use the commented line with SourceTimestamp as well.

### DIFF
--- a/opcua/server/address_space.py
+++ b/opcua/server/address_space.py
@@ -10,7 +10,7 @@ except:
 
 from opcua import ua
 from opcua.server.user_manager import UserManager
-
+import pytz
 
 class AttributeValue(object):
 
@@ -63,6 +63,9 @@ class AttributeService(object):
                 if not ua.ua_binary.test_bit(al.Value.Value, ua.AccessLevel.CurrentWrite) or not ua.ua_binary.test_bit(ual.Value.Value, ua.AccessLevel.CurrentWrite):
                     res.append(ua.StatusCode(ua.StatusCodes.BadUserAccessDenied))
                     continue
+            tz = pytz.timezone('Europe/Berlin')
+            writevalue.Value.ServerTimestamp = datetime.now(tz)
+            #writevalue.Value.SourceTimestamp = datetime.now(tz)
             res.append(self._aspace.set_attribute_value(writevalue.NodeId, writevalue.AttributeId, writevalue.Value))
         return res
 


opcua/server/address_space.py
```diff
@@ -63,6 +63,9 @@ def write(self, params, user=UserManager.User.Admin):
                 if not ua.ua_binary.test_bit(al.Value.Value, ua.AccessLevel.CurrentWrite) or not ua.ua_binary.test_bit(ual.Value.Value, ua.AccessLevel.CurrentWrite):
                     res.append(ua.StatusCode(ua.StatusCodes.BadUserAccessDenied))
                     continue
+            tz = pytz.timezone('Europe/Berlin')
+            writevalue.Value.ServerTimestamp = datetime.now(tz)
```
swamper123, line 67: `datetime.now(tz.utc)` would be better, since every time thing on server side is based on UTC. Clients shall provide any conversions between UTC and local time.

swamper123 (changes requested):

opcua/server/address_space.py
```diff
@@ -63,6 +63,9 @@ def write(self, params, user=UserManager.User.Admin):
                 if not ua.ua_binary.test_bit(al.Value.Value, ua.AccessLevel.CurrentWrite) or not ua.ua_binary.test_bit(ual.Value.Value, ua.AccessLevel.CurrentWrite):
                     res.append(ua.StatusCode(ua.StatusCodes.BadUserAccessDenied))
                     continue
+            tz = pytz.timezone('Europe/Berlin')
```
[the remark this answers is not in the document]
mbay-SAG: Totally agree!

opcua/server/address_space.py
mbay-SAG: Agreed.

mbay-SAG (commented):
Agreed with the comments!

oroulet:
That feature always comes back. I think it was removed due to performance. We should check if we have performance issues with that branch

PR #1208, opened 2021-01-07
Update client.py

davesliu:
expose the session_timeout and secure_channel_timeout parameters with default value, so that the user can set the session or secure channel timeout.
Otherwise, it may print warning log info if session timeout value is inconsistent between OPCUA client and server.

### DIFF
--- a/opcua/client/client.py
+++ b/opcua/client/client.py
@@ -92,7 +92,7 @@ class Client(object):
     if use_crypto is False:
         logging.getLogger(__name__).warning("cryptography is not installed, use of crypto disabled")
 
-    def __init__(self, url, timeout=4):
+    def __init__(self, url, timeout=4, session_timeout=3600000, secure_channel_timeout=3600000):
         """
 
         :param url: url of the server.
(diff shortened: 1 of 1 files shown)


AndreasHeine:
you know that the used sessiontimeout is negotiated in create_session req/res
so the warning simply say that the client asks for a sessiontimeout which the server dont want to have!

https://reference.opcfoundation.org/v104/Core/docs/Part4/5.6.2/

revisedSession Timeout 	
Duration 
	Actual maximum number of milliseconds that a Session shall remain open without activity. The Server should attempt to honour the Client request for this parameter, but may negotiate this value up or down to meet its own constraints.

opcua/client/client.py
```diff
@@ -125,8 +125,8 @@ def __init__(self, url, timeout=4):
         self.product_uri = "urn:freeopcua.github.io:client"
         self.security_policy = ua.SecurityPolicy()
         self.secure_channel_id = None
-        self.secure_channel_timeout = 3600000  # 1 hour
-        self.session_timeout = 3600000  # 1 hour
+        self.secure_channel_timeout = session_timeout
+        self.session_timeout = secure_channel_timeout
```
oroulet, line 129: there is no reason to make thes properties available as argument. Just set the properties

PR #1451, opened 2022-04-01
Support custom struct optional fields

deeTEEcee:
Add support for [OPC UA Optional FIelds](https://reference.opcfoundation.org/v104/Core/docs/Part6/5.2.7/)

Looking at the code, the library already supports encoding/decoding for custom objects (see struct_to_binary and struct_from_binary) but it was not fully implemented. The missing piece was loading and structuring the generated python code correctly to match the later steps for encoding/decoding. What this means for implementation is:

Handle bits and byte encoding through the use of ua_switches and creating UInt32 bit-encoding objects with necessary.
I also added some unit tests to make things easier and let us know what we're really looking at. 

**Extra**:
* The existing unit test doesn't register custom object types so nested objects can't get loaded as they haven't been added to the ua namespace. I've added in register=True in my unit test.
* I modified the directories a bit in the tests because it loads it assuming your root directory is the test folder itself. In my case, I'm using Pycharm and have to do adjustments to get that working. I thought this might make things easier for IDE's with the test kit already ready.


deeTEEcee:
@oroulet thanks for merging this in! can we expect a version upgrade too?

PR #1542, opened 2024-05-18
Update address_space.py

BruceGitHub2015:
fixing this issue https://github.com/FreeOpcUa/opcua-asyncio/issues/1529


oroulet:
you are refering to a bug in opcua-asyncio but now we have a merge request for pyhton-opcua?
The fix might be correct but it is applied to wrong repository

BruceGitHub2015:
> you are refering to a bug in opcua-asyncio but now we have a merge request for pyhton-opcua? The fix might be correct but it is applied to wrong repository

python-opcua has the same issue

